PR #258, opened 2021-06-08
feat: CI/CD Release to PyPi via Cloud Build

dhercher:
- Adding the files to release to PyPi via CloudBuild
- Aligning release docs with new process

We still need to test this fully on the next tag, then we can automate it


renzokuken (approved):
LGTM

PR #267, opened 2021-06-24
feat: add table matching score as a param incase adjusted is needed

dhercher:
The distance score is calculated between 0 and 1.  Generally the default, 80%, works for most cases.  If you find the default is not working well, this feature allows the value to be overridden

### DIFF
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -170,6 +170,10 @@ def _configure_find_tables(subparsers):
     find_tables_parser.add_argument(
         "--allowed-schemas", "-as", help="List of source schemas to match."
     )
+    find_tables_parser.add_argument(
+        "--score-cutoff", "-score",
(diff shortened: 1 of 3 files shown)


data_validation/cli_tools.py
```diff
@@ -170,6 +170,10 @@ def _configure_find_tables(subparsers):
     find_tables_parser.add_argument(
         "--allowed-schemas", "-as", help="List of source schemas to match."
     )
+    find_tables_parser.add_argument(
```
nehanene15, line 173: Can we document this argument in the README?

nehanene15 (changes requested):

data_validation/cli_tools.py
dhercher: done

nehanene15 (approved):

PR #272, opened 2021-06-28
docs: clean setup

dhercher:
cleaning requirements file away and using setup to avoid pypi build issues and for better clarity


renzokuken (approved):
LGTM

PR #275, opened 2021-06-29
feat: Allow user to specify a format for stdout (#242)

### LINKED ISSUE #242: Allow user to specify a format for stdout
Currently, the default stdout output is not readable due to the large amount of fields we print. The user should be able to specify a `--format` flag similar to the [gcloud format flag](https://cloud.google.com/sdk/gcloud/reference#--format).

We can also trim down the amount of fields we output to stdout if they are user specified parameters.

yogeshtewari:
Updated code to take in `--format` flag and pretty print formatted results on the console. Default format is tabular grid.

### DIFF
--- a/data_validation/__main__.py
+++ b/data_validation/__main__.py
@@ -272,11 +272,12 @@ def convert_config_to_yaml(args, config_managers):
     return yaml_config
 
 
-def run_validation(config_manager, verbose=False):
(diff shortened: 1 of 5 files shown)


yogeshtewari:
Closing this PR to work on a fresh branch which leverages config_managers instead of the raw args.

PR #280, opened 2021-07-12
feat: logic to deploy dvt on Cloud Run

dhercher:
Adding logic in samples which walk through deploying DVT on Cloud Run.

The README for the sample walks through
- Dockerfile used in the build
- How to build DVT on gcr with gcloud
- Deploying a service to Cloud Run
- An example test against the new Cloud Run service

### DIFF
--- a/samples/run/README.md
+++ b/samples/run/README.md
@@ -1,5 +1,16 @@
 # Data Validation on Cloud Run
 
+### Quick Steps
+
+The dpeloyment logic is discussed in more detail below; however, to quickly
+deploy you follow follow this simple script:
+
+```
+export PROJECT_ID=<PROJECT-ID>
+./deploy.sh
+python3 test.py
+```
+
 ### Build Docker Image
 
 You will need to build a Docker image to be used by Cloud Run. In order to add
@@ -99,3 +110,17 @@ res = requests.post(
     url, headers={"Authorization": "Bearer " + get_token()}, json=data)
 print(res.content.decode())
 ```
+
+### Oracle Setup
+
+If you would like to use Data Validation against an Oracle DB you will need to
+supply your own license files.  To do so:
+
+1) Create an `oracle` directory and add your .rpm files into it.
+
+- oracle/oracle-instantclient12.2-basiclite-12.2.0.1.0-1.x86_64.rpm
+- oracle/oracle-instantclient12.2-devel-12.2.0.1.0-1.x86_64.rpm
+- oracle/oracle-instantclient12.2-odbc-12.2.0.1.0-2.x86_64.rpm
+
+2) Uncomment all logic in the Dockerfile under the Oracle Dependencies comments
+
--- a/samples/run/README.md
+++ b/samples/run/README.md
@@ -1,0 +1,96 @@
+# Data Validation on Cloud Run
+
+### Build Docker Image
+
+You will need to build a Docker image to be used by Cloud Run. In order to add
+Teradata or Oracle, you will need to customize the Dockerfile and add your
+licensed utilities.
+
+```
+export PROJECT_ID=<PROJECT-ID>
+gcloud builds submit --tag gcr.io/${PROJECT_ID}/data-validation \
+    --project=${PROJECT_ID}
+```
+
+### Deploy to Cloud Run
+
+```
+gcloud run deploy --image gcr.io/${PROJECT_ID}/data-validation \
+    --project=${PROJECT_ID}
+```
+
+### Test Cloud Run Endpoint
+
+You can easily run a request via Python. For example:
+
+```
+# Copyright 2020 Google LLC
+#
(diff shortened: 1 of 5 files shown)


samples/run/README.md
```diff
@@ -0,0 +1,96 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import re
+import requests
+
+PROJECT_ID = os.environ.get("PROJECT_ID", "pso-kokoro-resources")
```
nehanene15: We should get rid of the 'pso-kokoro-resources' default so it is more universal

samples/run/README.md
```diff
@@ -0,0 +1,96 @@
...
+### Deploy to Cloud Run
+
+```
+gcloud run deploy --image gcr.io/${PROJECT_ID}/data-validation \
+    --project=${PROJECT_ID}
+```
+
+### Test Cloud Run Endpoint
+
+You can easily run a request via Python. For example:
```
nehanene15: We can mention here that this is the `test.py` code. We should also add that the service takes in a JSON configuration object which is equivalent to the ConfigManager._config dictionary.

samples/run/README.md
```diff
@@ -0,0 +1,96 @@
...
+    },
+    "target_conn": {
+        "source_type": "BigQuery",
+        "project_id": PROJECT_ID,
+    },
+    "type": "Column",
+    "schema_name": "bigquery-public-data.new_york_citibike",
+    "table_name": "citibike_stations",
+    "target_schema_name": "bigquery-public-data.new_york_citibike",
+    "target_table_name": "citibike_stations",
```
nehanene15, line 92:
```suggestion
    "target_table_name": "citibike_stations",
    "result_handler": {
      "type": "BigQuery",
      "project_id": PROJECT_ID,
      "table_id": "pso_data_validator.results"
   },
```

samples/run/README.md
```diff
@@ -0,0 +1,96 @@
...
+            "target_column": None,
+            "field_alias": "count",
+            "type": "count"
+        }],
+}
+
+url = get_cloud_run_url("data-validation", PROJECT_ID)
+res = requests.post(
+    url, headers={"Authorization": "Bearer " + get_token()}, json=data)
+print(res.content)
```
nehanene15:
```suggestion
print(res.content.decode())
```

samples/run/deploy.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+export PROJECT_ID=pso-kokoro-resources
```
nehanene15: We can delete this line so the user can set their own project_id

samples/run/test.py
```diff
@@ -0,0 +1,63 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import re
+import requests
+
+PROJECT_ID = os.environ.get("PROJECT_ID", "pso-kokoro-resources")
```
nehanene15: Same as above; remove default as 'pso-kokoro-resources'

samples/run/test.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+data = {
+    "source_conn": {"source_type": "BigQuery", "project_id": PROJECT_ID,},
+    "target_conn": {"source_type": "BigQuery", "project_id": PROJECT_ID,},
+    "type": "Column",
+    "schema_name": "bigquery-public-data.new_york_citibike",
+    "table_name": "citibike_stations",
+    "target_schema_name": "bigquery-public-data.new_york_citibike",
+    "target_table_name": "citibike_stations",
```
nehanene15, line 50: Same as above; add result handler in example since that will be a common use case

samples/run/test.py
```diff
@@ -0,0 +1,63 @@
...
+            "target_column": None,
+            "field_alias": "count",
+            "type": "count",
+        }
+    ],
+}
+
+url = get_cloud_run_url("data-validation", PROJECT_ID)
+res = requests.post(url, headers={"Authorization": "Bearer " + get_token()}, json=data)
+print(res.content)
```
nehanene15:
```suggestion
print(res.content.decode())
```

nehanene15 (changes requested):

samples/run/test.py
dhercher: I have the BigQuery output example in the Readme, but since this is intended to validate the Cloud Run job is working it makes more sense just to return here IMO

nehanene15 (approved):
LGTM

PR #296, opened 2021-08-05
feat: Allow user to specify a format for stdout T2 (#242)

yogeshtewari:


data_validation/__main__.py
```diff
@@ -143,6 +143,7 @@ def build_config_managers_from_args(args):
             table_obj,
             labels,
             threshold,
+            format=args.format,
```
nehanene15: This doesn't need to be an optional parameter. Before we initialize ConfigManager.build_config_manager(), we can set format to args.format similar to the 'threshold' parameter. For example,
```
format = args.format if args.format else 'table'
```

tests/unit/result_handlers/test_text.py
```diff
@@ -39,7 +39,7 @@ def test_import(module_under_test):
 def test_basic_result_handler(module_under_test):
     """Test basic handler executes """
     result_df = DataFrame(SAMPLE_RESULT_DATA)
-    result_handler = module_under_test.TextResultHandler()
+    result_handler = module_under_test.TextResultHandler(format="table")
```
nehanene15: Can you add a unit test (maybe in unit/test_cli_tools.py) that checks if a user passes in an invalid format an error will be thrown?

nehanene15 (changes requested):

tests/unit/result_handlers/test_text.py
yogeshtewari: Done.

data_validation/__main__.py
yogeshtewari: Done.

nehanene15 (approved):

PR #297, opened 2021-08-11
feat: Updated Cloud Functions sample

### LINKED ISSUE #168: Data Validation on Cloud Functions
Using the functions framework, build a utility that allows DV to run on Cloud Functions.  This will avoid complex dependancy management on composer which is likely too have bad results.
https://github.com/GoogleCloudPlatform/functions-framework-python

- Worries around connection management make this a trickier problem

nehanene15:
Closes Issue #168 

Updates Cloud Functions sample to make more generic and added documentation. 
This will need to be refactored once we support GCS based configs


renzokuken (approved):
LGTM. Thanks for cleaning up the project refs.

PR #303, opened 2021-08-31
feat: Refactor CLI to fit Command Pattern

### LINKED ISSUE #279: Refactor CLI to Command Pattern
As the CLI has grown, there's many options that don't make sense to be specified together however they all exist for the same command (`run`). For example, `--grouped-columns` is an optional arg to the `run` command but it's only valid for a specific type of validation `GroupedColumn`. Same with `--primary-keys` and `Row` (#277).

To support future growth of the CLI and to simplify usage, I propose each validation becomes it's own top-level command. This would change the command-line from:
```python
data-validation run [ARGS]
```
to
```python
# Technically Column & Grouped Column is the same command with different opts 
# specified at runtime (e.g. the presence of a groupby key).
data-validation validate column [ARGS] 
data-validation validate schema [ARGS]
data-validation validate row [ARGS]
```

This change could be made backward compatible because the `run` command can continue to exist but now each command can be separated out into their own module with only the options relevant to that command present and inheriting from a validation command class ([Command Design Pattern](https://refactoring.guru/design-patterns/command)). This is similar to how we designed the gcloud SDK.

nehanene15:
Closes #279 

Updates DVT to use validation type specific commands like so: 
```
data-validation validate column --flags
data-validation validate schema --flags
```

Updated documentation and examples to reflect the change. It is backwards compatible to continue support for the `data-validation run` command.


data_validation/cli_tools.py
```diff
@@ -25,20 +25,20 @@
...
 --grouped-columns starttime \
 --sum tripduration --count tripduration
 
-data-validation run -t Column \
+data-validation validate column \
 -sc my_bq_conn -tc my_bq_conn \
 -tbls bigquery-public-data.new_york_citibike.citibike_trips,bigquery-public-data.new_york_citibike.citibike_stations \
 --sum tripduration,start_station_name --count tripduration,start_station_name \
--rc pso-kokoro-resources.pso_data_validator.results
+-bqrh pso-kokoro-resources.pso_data_validator.results
```
renzokuken, line 41: I've seen inconsistencies with the -bqrh flag on the develop branch. Do we have a test that this is currently functional?

renzokuken (changes requested):
Overall LGTM. I have one question regarding the switch to the -bqrh flag but aside from that Approve.

data_validation/cli_tools.py
nehanene15: Yes, the --bqrh flag runs exactly the same as the -rc flag using the `cli_tools.get_result_handler()` function. We have a test for it in `test_cli_tools.py`. The code will first check to see if the --bqrh flag was supplied; if not, it will check if the --rc flag exists; If neither are provided it will set to None.

renzokuken (approved):
LGTM

PR #326, opened 2021-10-14
feat: Track DVT GCS connections

### LINKED ISSUE #229: Tracking DVT Usage with GCS Connection Configs
We want to be able to track DVT usage across all users once its made open source.

Our current method tracks when results are sent to BigQuery, but this is limited and will often not be the method a customer is using when they run the tool (currently all Oracle to Postgres usage is being missed, for example)

We want to add logic to track
- Each validation run that occurs
- what type of validation is happening
- Success/Failure

Other information might be interesting, but most other fields would be potentially PII.

Methods to track:
- Google Analytics

dmedora:
Resolves #229. Provides client_info to Storage client so that requests are attributed to the tool. Includes docs update marking GCS as recommended for conn config storage.
Also includes docs update on running unit tests.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -110,6 +110,7 @@ If you want to test local changes to the tool, run the following command from th
 ```
 pip install .
 ```
+The unit test suite can be executed using either `pytest tests/unit` or `nox -s unit` from the root directory.
 
 
 ## Build a Docker container 
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -110,7 +110,7 @@ If you want to test local changes to the tool, run the following command from th
 ```
 pip install .
 ```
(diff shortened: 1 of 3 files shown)


docs/installation.md
```diff
@@ -110,6 +110,7 @@ If you want to test local changes to the tool, run the following command from th
 ```
 pip install .
 ```
+The unit test suite can be executed using either `pytest tests/unit` or `nox -s unit` from the root directory.
```
nehanene15:
```suggestion
The unit test suite can be executed using either `pytest tests/unit` or `python -m nox -s unit` from the root directory. If you are using nox, you will need to run `pip install nox` first.
```

nehanene15 (changes requested):

nehanene15 (approved):

Issue #323, opened 2021-10-13
Row level hash validation gives int64 overflow for table with differences


nehanene15:
We need to check if the source and target aggregation values are not equal for row hash validation, instead of checking if the difference is 0. This will avoid the int64 overflow when calculating the difference between source and target agg values.

PR #328, opened 2021-10-15
fix: Issue323 row hash

nehanene15:
Closes Issue #323 

Fixes int64 overflow for row hash validations by removing the difference calculation and instead checking if `source_agg_value == target_agg_value`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,4 +75,4 @@ override.tf.json
 terraform.rc
 
 # Custom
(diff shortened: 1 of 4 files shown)


renzokuken (approved):
minor nit. I'd like to change the is_row_hash boolean to something more generic like is_value_comparison since I see this usecase applying to any cell comparison

PR #330, opened 2021-10-16
bug: improve env vars management in tests to avoid errors

dhercher:
- Fixed path join bug causing connections not to list (and added testing to state manager)
- Force the PSO home env var to be empty at the start of tests (expected state, change the environ on a test by test basis if required)
- Supply default Project for BigQuery (since its the only integration thats easy to run locally)


nehanene15 (approved):
LGTM

Issue #335, opened 2021-10-26
Breaking update on version 1.5.0 - Issue with ibis_teradata/compiler.py StringIO usage


aymanfarhat:
After updating from version 1.4.0 to 1.5.0, the tool is breaking on pretty much any command related to teradata querying.

Example command

```
data-validation -v validate column \
-sc TD_DVT \
-tc BQ_DVT \
-tbls TEST_DB.USA_NAMES_CORRECT=TEST_DB.USA_NAMES_CORRECT \
-bqrh rich-torus-323508.DVT_RESULTS.sample \
--avg=repeated
```

Output

```
/Users/aymanf/projects/teradata-experiments/tdvt-1.5/lib/python3.8/site-packages/pandas/core/frame.py:1549: FutureWarning: Using short name for 'orient' is deprecated. Only the options: ('dict', list, 'series', 'split', 'records', 'index') will be used in a future version. Use one of the above to silence this warning.
  warnings.warn(
-- ** Source Query ** --
Traceback (most recent call last):
  File "/Users/aymanf/projects/teradata-experiments/tdvt-1.5/bin/data-validation", line 8, in <module>
    sys.exit(main())
  File "/Users/aymanf/projects/teradata-experiments/tdvt-1.5/lib/python3.8/site-packages/data_validation/__main__.py", line 379, in main
    validate(args)
  File "/Users/aymanf/projects/teradata-experiments/tdvt-1.5/lib/python3.8/site-packages/data_validation/__main__.py", line 358, in validate
    run(args)
  File "/Users/aymanf/projects/teradata-experiments/tdvt-1.5/lib/python3.8/site-packages/data_validation/__main__.py", line 339, in run
    run_validations(args, config_managers)
  File "/Users/aymanf/projects/teradata-experiments/tdvt-1.5/lib/python3.8/site-packages/data_validation/__main__.py", line 315, in run_validations
    run_validation(config_manager, verbose=args.verbose)
  File "/Users/aymanf/projects/teradata-experiments/tdvt-1.5/lib/python3.8/site-packages/data_validation/__main__.py", line 304, in run_validation
    validator.execute()
  File "/Users/aymanf/projects/teradata-experiments/tdvt-1.5/lib/python3.8/site-packages/data_validation/data_validation.py", line 94, in execute
    result_df = self._execute_validation(
[...]

PR #336, opened 2021-10-26
Fix: Include StringIO into teradata ibis compiler.py

aymanfarhat:
This small updated makes sure StringIO is included into compiler.py. Referencing issue #335 in which Teradata queries are breaking on V1.5.0


renzokuken (approved):
thanks for catching this! This particular isn't covered by our current integration tests - I am working on a separate thread to improve our teradata coverage.

PR #340, opened 2021-11-02
feat: GCS support for validation configs

### LINKED ISSUE #288: Add GCS support for configurations/connections

dmedora:
Resolves #288. Allows validation configs to be saved to GCS when the `PSO_DV_CONFIG_HOME` env var is set. Also adds 'list' functionality and corresponding docs updates.

### DIFF
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -11,8 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-
 """The Data Validation CLI tool is intended to help to build and execute
 data validation runs with ease.
 
@@ -52,7 +50,6 @@ import uuid
 
 from data_validation import consts
 from data_validation import state_manager
-
 
 CONNECTION_SOURCE_FIELDS = {
     "BigQuery": [
@@ -122,7 +119,10 @@ CONNECTION_SOURCE_FIELDS = {
         ["host", "Desired Impala host"],
         ["port", "Desired Imapala port (10000 if not provided)"],
         ["database", "Desired Impala database (default if not provided)"],
-        ["auth_mechanism", "Desired Impala auth mechanism (PLAIN if not provided)"],
+        [
+            "auth_mechanism",
+            "Desired Impala auth mechanism (PLAIN if not provided)"
+        ],
         [
             "kerberos_service_name",
             "Desired Kerberos service name ('impala' if not provided)",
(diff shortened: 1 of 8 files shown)


data_validation/cli_tools.py
```diff
@@ -204,6 +201,9 @@ def _configure_run_config_parser(subparsers):
     run_config_parser = subparsers.add_parser(
         "run-config", help="Run validations stored in a YAML config file"
     )
+    run_config_subparsers = run_config_parser.add_subparsers(dest="run_config_cmd")
+    _ = run_config_subparsers.add_parser("list", help="List your validation configs")
```
dhercher: Without having tested this, using an optional command parser is a slightly strange feature (and I'm not sure its doable?).

The naming structure itself here is slightly strange as well (eg `run-config list` is not easy to find and understand)

It seems to me that the goal here is to create a `configs` CLI section like we have with `connections`.  If thats the case perhaps do so more explicitly.
`configs list` & `configs run` & also maybe add `configs get` though only a nice to have.

For backwards compat you'll need to keep `run-config` though perhaps it can be a hidden config going forward?

dmedora: I agree, `run-config list` is a little strange. I was trying to stick with `run-config`, but adding a `configs` makes sense. Will amend.

dmedora: Added the `configs` command (run, list, get), and left `run-config` in place. PTAL when you have a chance!

data_validation/state_manager.py
```diff
@@ -113,6 +160,8 @@ def _write_file(self, file_path: str, data: str):
             with open(file_path, "w") as file:
                 file.write(data)
 
+        print("Success! Config output written to {}".format(file_path))
```
nehanene15, line 163: Creating a connection also uses the `_write_file()` method on [line 62](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/data_validation/state_manager.py#L62) so we should only limit this print statement for config output writes or remove it

data_validation/state_manager.py
```diff
@@ -154,7 +203,7 @@ def _read_gcs_file(self, file_path: str) -> str:
         gcs_file_path = self._get_gcs_file_path(file_path)
         blob = self.gcs_bucket.get_blob(gcs_file_path)
 
-        return blob.download_as_string()
+        return blob.download_as_bytes()
```
nehanene15, line 206: The `_read_file()` method on [line 103](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/6b2614dcfae3ca67d85221a9ceab3881c5c6b30d/data_validation/state_manager.py#L103) expects a string from this method

data_validation/state_manager.py
dmedora: I was thinking that 'config' could refer to either connection config or validation config, I could adjust to 'Output file written..'?

data_validation/state_manager.py
dmedora: Changed function from string to bytes because `download_as_string` already used to return bytes [despite its name](https://github.com/googleapis/google-cloud-python/issues/9290), and is [deprecated](https://github.com/googleapis/python-storage/issues/504).

nehanene15 (changes requested):
There are some edits to the state manager that may affect GCS based connections. Besides that, LGTM

data_validation/state_manager.py
nehanene15: Ah ok that works

nehanene15 (approved):

PR #344, opened 2021-11-13
feat: deploy flask app via CLI

dhercher:
Adding a CLI option to deploy as a Flask App.

This brings the Cloud Run flask logic into the main code path to be shared for Cloud Run / GKE / and other containerized apps

### DIFF
--- a/data_validation/__main__.py
+++ b/data_validation/__main__.py
@@ -379,6 +379,7 @@ def main():
         validate(args)
     elif args.command == "deploy":
         from data_validation import app
+
         app.app.run(debug=True, host="0.0.0.0", port=int(os.environ.get("PORT", 8080)))
     else:
         raise ValueError(f"Positional Argument '{args.command}' is not supported")
--- a/data_validation/__main__.py
+++ b/data_validation/__main__.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
+import os
 import json
 from yaml import dump, load, Dumper, Loader
 
@@ -377,6 +377,9 @@ def main():
         print(run_raw_query_against_connection(args))
     elif args.command == "validate":
         validate(args)
+    elif args.command == "deploy":
+        from data_validation import app
+        app.app.run(debug=True, host="0.0.0.0", port=int(os.environ.get("PORT", 8080)))
     else:
         raise ValueError(f"Positional Argument '{args.command}' is not supported")
 
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -172,7 +172,7 @@ def configure_arg_parser():
 def _configure_deploy(subparsers):
     """Configure arguments for deploying as a service."""
     subparsers.add_parser(
-        "deploy", help="Deploy Data Validation as a Service (w/ Flask)"
+        "deploy", help="(Beta) Deploy Data Validation as a Service (w/ Flask)"
     )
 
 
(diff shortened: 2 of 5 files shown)


data_validation/cli_tools.py
```diff
@@ -164,10 +164,18 @@ def configure_arg_parser():
...
         _configure_run_parser(subparsers)
+        _configure_deploy(subparsers)
 
     return parser
 
 
+def _configure_deploy(subparsers):
+    """Configure arguments for deploying as a service."""
+    subparsers.add_parser(
+        "deploy", help="(Beta) Deploy Data Validation as a Service (w/ Flask)"
```
nehanene15: We added some functionality a while back to support beta CLI commands in PR #249. It would allow a user to run `data-validation beta deploy`

Looks like that functionality is broken. For some reason the original implementation of the beta CLI on [line 124](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/249/files#diff-ce217c386e702709267508ebe1bc725106c6bcba326c1ccb29951d506855ae6aR124) of the PR has moved [inside the if loop](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/data_validation/cli_tools.py#L150) which is a breaking change.

Could we fix the change and support the deploy argument in the beta CLI?

Something like this:
```
def configure_arg_parser():
    """Extract Args for Run."""
    parser = argparse.ArgumentParser(
        usage=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter
    )

    parser.add_argument("--verbose", "-v", action="store_true", help="Verbose logging")
    parser.add_argument("beta", nargs='?', help="Beta flag to enable beta features for the tool.")

    if "beta" in sys.argv:
        subparsers = parser.add_subparsers(dest="command")
        _configure_deploy(subparsers)
    else:
        subparsers = parser.add_subparsers(dest="command")
        _configure_validate_parser(subparsers)
        _configure_run_config_parser(subparsers)
        _configure_connection_parser(subparsers)
        _configure_find_tables(subparsers)
        _configure_raw_query(subparsers)
        _configure_run_parser(subparsers)

    return parser
```

data_validation/__main__.py
```diff
@@ -377,6 +377,10 @@ def main():
         print(run_raw_query_against_connection(args))
     elif args.command == "validate":
         validate(args)
+    elif args.command == "deploy":
```
nehanene15, line 395: Let's document that this is a beta CLI option and that the env var 'PORT' must be set. I also added an [example JSON config](https://github.com/GoogleCloudPlatform/professional-services-data-validator/tree/develop/samples/functions#json-configuration) input to the Flask app if you want to copy that

nehanene15 (changes requested):

data_validation/cli_tools.py
dhercher: Changing the beta design a little to make it prominent.  Should be resolved now

data_validation/__main__.py
dhercher: done

data_validation/cli_tools.py
```diff
@@ -145,29 +145,37 @@ def configure_arg_parser():
...
+    _configure_raw_query(subparsers)
+    _configure_run_parser(subparsers)
+    _configure_beta_parser(subparsers)
 
     return parser
 
 
+def _configure_beta_parser(subparsers):
+    """Configure beta commands for the parser."""
+    connection_parser = subparsers.add_parser(
```
nehanene15, line 163: The updated beta CLI code looks great. Last thing I would add is documentation in the README for the beta deploy command. Besides that, LGTM

nehanene15 (approved):

PR #345, opened 2021-11-17
feat: first class support for row level hashing

renzokuken:


data_validation/data_validation.py
```diff
@@ -216,6 +215,8 @@ def execute_recursive_validation(self, validation_builder, grouped_fields):
                     )
         elif self.config_manager.primary_keys:
             validation_builder.add_config_query_groups(self.config_manager.primary_keys)
+
+        elif self.config_manager.primary_keys and len(grouped_fields) == 0:
```
tswast, line 216: Possible the order of these if statements is wrong. I don't think this branch will ever be reached since `elif self.config_manager.primary_keys:` appears above it.

data_validation/__main__.py
```diff
@@ -28,6 +27,8 @@
 from data_validation.config_manager import ConfigManager
 from data_validation.data_validation import DataValidation
 
+Dumper.ignore_aliases = lambda *args: True
```
dhercher, line 32: Add a comment why this is here

data_validation/__main__.py
```diff
@@ -78,30 +79,68 @@ def get_aggregate_config(args, config_manager):
...
     Args:
         config_manager (ConfigManager): Validation config manager instance.
     """
-    config_manager.append_aggregates(get_aggregate_config(args, config_manager))
-    if args.primary_keys and not args.grouped_columns:
-        if not args.grouped_columns and not config_manager.use_random_rows():
-            logging.warning(
-                "No Grouped columns or Random Rows specified, ignoring primary keys."
+    config_manager.append_calculated_fields(get_calculated_config(args, config_manager))
+    if config_manager.validation_type == consts.COLUMN_VALIDATION:
```
dhercher, line 130: Could this be grouped column?  if so should it be included here?

data_validation/__main__.py
```diff
@@ -140,7 +177,7 @@ def build_config_managers_from_args(args):
 
     # Schema validation will not accept filters, labels, or threshold as flags
     filter_config, labels, threshold = [], [], 0.0
-    if config_type != consts.SCHEMA_VALIDATION:
+    if config_type == consts.COLUMN_VALIDATION:
```
dhercher: row validations (and ditto grouped throughout) don't support filters?

data_validation/cli_tools.py
```diff
@@ -266,6 +265,16 @@ def _configure_run_parser(subparsers):
         "-max",
         help="Comma separated list of columns for max 'col_a,col_b' or * for all columns",
     )
+    run_parser.add_argument(
```
dhercher: lets break out the overlapping arguments here and below into a function (they should always be the same, this one is historical I think

data_validation/combiner.py
```diff
@@ -161,30 +158,27 @@ def _calculate_differences(
         differences_joined = source.cross_join(target)
 
     differences_pivots = []
-
     for field, field_type in schema.items():
-        if field not in validation_fields:
+        if field not in validations.keys():
```
dhercher: is `.keys()` needed?  `if field not in validations` is more efficient

data_validation/config_manager.py
```diff
@@ -232,12 +242,12 @@ def get_source_ibis_table(self):
             )
         return self._source_ibis_table
 
-    def get_source_ibis_calculated_table(self):
+    def get_source_ibis_calculated_table(self, n=None):
```
dhercher: what is `n`.  should be documented or obvious

data_validation/data_validation.py
dhercher: I don't believe primary keys check below can ever be reached since the grouped_fields value here will always be ==0 at this point

data_validation/query_builder/query_builder.py
```diff
@@ -279,7 +301,7 @@ def hash(config, fields):
     def ifnull(config, fields):
```
dhercher, line 308: confusing to read.

1. is default_string used somewhere downstream?
2. why is it valid to take an arbitrary first value in fields and then overwrite itself?  Is this really the same object?

data_validation/data_validation.py
```diff
@@ -283,7 +286,10 @@ def _execute_validation(self, validation_builder, process_in_memory=True):
         source_query = validation_builder.get_source_query()
         target_query = validation_builder.get_target_query()
 
-        join_on_fields = validation_builder.get_group_aliases()
+        if self.config_manager.validation_type == consts.ROW_VALIDATION:
```
dhercher: more readable with single line
`a if value else b`

data_validation/config_manager.py
```diff
@@ -104,12 +102,7 @@ def random_row_batch_size(self):
         )
 
     def process_in_memory(self):
-        if (
-            self.validation_type == "Row"
-            and self.get_source_connection() == self.get_target_connection()
-        ):
-            return False
-
+        """TODO: (emceehilton) we need to eventually make remote execution a parameter"""
```
dhercher: Not a great place for a TODO

dhercher (dismissed):
Round 1 of changes

data_validation/__main__.py
renzokuken: I guess my question is GroupedColumn all that different from Column at this point now that row level validation has been split out? Happy to change it back just not sure what the value add is for a tertiary validation type?

nehanene15: I agree that GroupedColumn is basically the same as Column validation at this point. 

That being said, we do update the validation_type to `consts.GROUPED_COLUMN_VALIDATION` if args.grouped_column exists [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/data_validation/__main__.py#L123). So having the nested `if args.grouped_columns is not None:` block on line 128 will never be true if the validation type is a column validation

data_validation/query_builder/query_builder.py
renzokuken: agreed it's confusing. The default string is the replacement value when calling the ibis IFNULL equivalent. Looking at the statement it never actually reads from the config so I'll get that fixed.

https://ibis-project.org/docs/generated/ibis.expr.api.ValueExpr.fillna.html

Disclaimer: the only reason this was updated in this PR is because the order of inputs was backward.

renzokuken:
@nehanene15 @dhercher PR updated for review. Please let me know if you have any questions or concerns

data_validation/config_manager.py
```diff
@@ -104,12 +102,7 @@ def random_row_batch_size(self):
...
 
     def process_in_memory(self):
-        if (
-            self.validation_type == "Row"
-            and self.get_source_connection() == self.get_target_connection()
-        ):
-            return False
-
+        """Return whether to process in memory or on a remote platform."""
         return True
```
nehanene15, line 106: Why will this always be True?

data_validation/config_manager.py
```diff
@@ -157,6 +150,23 @@ def append_primary_keys(self, primary_key_configs):
...
+        return self._config.get(consts.CONFIG_COMPARISON_FIELDS, [])
+
+    def append_comparison_fields(self, field_configs):
+        """Append field configs to existing config."""
+        self._config[consts.CONFIG_COMPARISON_FIELDS] = (
+            self.comparison_fields + field_configs
+        )
+
+    @property
+    def is_grouped_row_validation(self):
```
nehanene15: Are we using this property anywhere?

data_validation/config_manager.py
```diff
@@ -347,6 +359,32 @@ def build_config_manager(
             verbose=verbose,
         )
 
+    def build_config_comparison_fields(self, fields, depth=None):
+        """Return list of field config objects."""
+        field_configs = []
+        # source_table = self.get_source_ibis_calculated_table(depth=depth)
```
nehanene15: nit: Any reason to leave these comments in here? Since the logic is now in `build_config_calculated_fields`

data_validation/config_manager.py
```diff
@@ -423,3 +461,86 @@ def build_config_column_aggregates(self, agg_type, arg_value, supported_types):
...
+        col_names = []
+        for i, calc in enumerate(order_of_operations):
+            if i == 0:
+                previous_level = [x for x in casefold_source_columns.values()]
+            else:
+                previous_level = [k for k, v in column_aliases.items() if v == i - 1]
+            if calc in ["concat", "hash"]:
+                col = {}
+                col["reference"] = previous_level
+                col["name"] = f"{calc}__all"
```
nehanene15, line 506: This means that if a user provides certain columns for the row validation via `data-validation validate row --comparison-fields col_a ...`, the validation name will still remain 'hash_all' correct?

data_validation/result_handlers/text.py
```diff
@@ -35,6 +36,17 @@ def print_formatted_(self, result_df):
         Utility for printing formatted results
         :param result_df
         """
+        # the text transformer chokes on bytestring results (e.g. SHA256) this
+        # dataframe slice is to remove the source and target values
+        if "validation_type" in result_df.columns:
```
nehanene15: Won't every validation fall under this if loop? Assuming you only want to change row validation result handlers.

data_validation/validation_builder.py
```diff
@@ -90,12 +95,21 @@ def get_metadata(self):
 
     def get_group_aliases(self):
         """ Return List of String Aliases """
-        return self.group_aliases.keys()
+        return list(self.group_aliases.keys())
+
+    def get_primary_keys(self):
+        """ Return List of String Aliases """
+        # do we need this?
```
nehanene15, line 102: Don't think we need this since you can always use config_manager.primary_keys()

data_validation/validation_builder.py
```diff
@@ -265,6 +329,7 @@ def get_source_query(self):
         }
         query = self.source_builder.compile(**source_config)
         if self.verbose:
+            print(source_config)
```
nehanene15, line 332: Let's make sure we print the target_config in the `get_target_query(self)` method too for standardization

tests/unit/test_schema_validation.py
```diff
@@ -188,7 +188,9 @@ def test_execute(module_under_test, fs):
 
     dv_client = data_validation.DataValidation(SAMPLE_SCHEMA_CONFIG, verbose=True)
     result_df = dv_client.schema_validator.execute()
+    print(result_df)
     failures = result_df[result_df["status"].str.contains("Fail")]
+    print(failures)
```
nehanene15: nit: any reason to print these in the tests?

data_validation/config_manager.py
renzokuken: It's an open debate for when/where we apply remote processing. It's not an option for heterogeneous migration *and* for row level validation is failing for BQ-to-BQ migrations currently. Eventually we want to refine the logic here for larger scale processign.

data_validation/result_handlers/text.py
renzokuken: this was actually to bypass the unit tests for the text handler which uses pandas series...I'm open to a better solution.

tests/unit/test_schema_validation.py
renzokuken: fixed

data_validation/validation_builder.py
renzokuken: In principal I agree but a lot of inconsistencies have sprung up between the validation builder and config manager where the two objects emit different datatypes of the same business objects. Most references in that method use the validation builder so I did the same. I'd like to propose a cleanup sprint sometime in the future to make the separation of functionality more explicit

data_validation/config_manager.py
renzokuken: technically yes...though I question that use case. The method you're referencing is to build all of the dependencies for a multi-column hash - if this is triggered then that means that in __main__.py the hash__all alias is automatically referenced as a comparison field. Basically if you use hash in the CLI there should be no reason to use comparison-fields. The name mismatch is only pertinent if a custom YAML is submitted.

data_validation/result_handlers/text.py
nehanene15: Ah I see.. I would instead add a 'validation_type' column within the unit test sample pandas series

data_validation/config_manager.py
nehanene15: So when would we use comparison fields via the CLI?

nehanene15:
We will also need to add documentation about this functionality - we can do that in this PR or open a new one. This would include:
- CLI flags/descriptions
- Updating the examples.md with row validations
- When to use calculated vs comparison fields, How to exclude columns from hashing
- Limitations of row validation

data_validation/cli_tools.py
```diff
@@ -380,12 +349,70 @@ def _configure_validate_parser(subparsers):
...
+    row_parser.add_argument(
+        "--labels", "-l", help="Key value pair labels for validation run"
+    )
+    row_parser.add_argument(
+        "--threshold",
+        "-th",
+        type=threshold_float,
+        help="Float max threshold for percent difference",
+    )
+    row_parser.add_argument(
```
nehanene15, line 421: Is this supposed to be here? (Grouped columns within the row parser)

nehanene15 (commented):
Could we also document a couple examples of row hashing in docs/examples?

nehanene15 (approved):

PR #349, opened 2021-12-01
fix: Issue348 casting

nehanene15:
- Fixes casting error when applied via YAML
- Fixes bug in code so that `end_time` is properly set before the validation result handler executes
- Updates YAML file example in docs


renzokuken (approved):
LGTM

PR #357, opened 2021-12-31
feat: Random Row Validation Logic

dhercher:
This CL covers the most basic random row logic.

- Adds flags to allow a user to `use-random-row` and specify `random-row-batch-size`
- Uses the primary-keys flags which already exist (though there need to be discussion around this)
- Implements a base query + limit(batch_size)
- If specified, tested that it works for BigQuery

MISSING:
- Need to add random ordering before the limit (unsure if/how this is supported, though the first version is still valuable)

NOTES:
- Batch size is likely limited by the DB query filter limits (most DBs have query text size and filter size limits)


renzokuken (approved):

bnb01:
Has this feature been placed in the documentation?

nehanene15:
> Has this feature been placed in the documentation?

Not yet. Currently we have merged support for limiting query size to a certain number of rows (using LIMIT), but are working on randomizing that sample. Once we add support for that, we will add it to documentation.

cc @dhercher

PR #359, opened 2022-01-18
Fix: Multiple updates

bluPhy:
Fixing spelling in several files
Adding double quote to prevent globbing and word splitting.


nehanene15 (approved):
LGTM. Thanks for catching these.

Issue #341, opened 2021-11-04
Status shows 'success' when pct_difference in nan


nehanene15:
![image](https://user-images.githubusercontent.com/25696756/140348888-f531aa36-4fe1-409f-8144-968377f3aac8.png)

In [line 129 in combiner.py](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/data_validation/combiner.py#L129) there should be a check to ensure that the pct_difference isn't null before the status is calculated.

PR #366, opened 2022-02-15
fix: test for nan when calculating fail/success in combiner (#341)

ajwelch4:
- Fixes erroneous `success` status when agg values are `nan` (#341).


nehanene15 (approved):
LGTM

Issue #329, opened 2021-10-16
Validation Status is NaN if Value DNE


dhercher:
If the entire set of values for a given slice of data is missing, the value becomes `nan` and it causes the status to be `nan` as well.

One possible solution is:
```
--- a/data_validation/combiner.py
+++ b/data_validation/combiner.py
@@ -80,6 +80,7 @@ def generate_report(
         print(documented.compile())
 
     result_df = client.execute(documented)
+    result_df.status.fillna("fail", inplace=True)
 
     return result_df
```

To reproduce, run:
```
data-validation validate column 
-sc my_bq_conn 
-tc my_bq_conn 
-tbls bigquery-public-data.new_york_citibike.citibike_trips 
--grouped-columns starttime 
--filters "starttime>'2018-05-26':starttime>'2018-05-27'"
```

PR #370, opened 2022-02-18
fix: ensure all statuses are success or fail, particularly after _join_pivots (#329)

ajwelch4:
I'm not sure if this is intended behavior or not but I would imagine we want a proper `success` or `fail` validation status for each row. In particular, if one result set contains a row that the other result set does not contain, we probably want to know about it. Currently in this case, `_join_pivots` will return a status of `nan` as described in #329. This PR sets all `nan` statuses to `fail`.

- fix: ensure all statuses are success or fail, particularly after `_join_pivots` (#329)

### DIFF
--- a/data_validation/combiner.py
+++ b/data_validation/combiner.py
@@ -90,6 +90,7 @@ def generate_report(
         print(documented.compile())
 
     result_df = client.execute(documented)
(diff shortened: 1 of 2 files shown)


nehanene15 (approved):
LGTM

PR #373, opened 2022-02-25
adding addons for impala hive hashing functions

renzokuken:
placeholder for review


nehanene15:
Hive hashing function is added in PR #392

Issue #377, opened 2022-02-26
Make status values consistent across validation types


ajwelch4:
Split out from #260.

> Schema validation outputs 'Pass'/'Fail' for the 'status' column in the result set. The other validation types output 'success'/'fail'.

PR #378, opened 2022-02-26
fix: make status values consistent across validation types (#377)

ajwelch4:
Schema validation outputs `Pass`/`Fail` for the `status` column in the result set. The other validation types output `success`/`fail`. This PR unifies `status` values across validation types.

- fix: make status values consistent across validation types (#377)

### DIFF
--- a/data_validation/combiner.py
+++ b/data_validation/combiner.py
@@ -89,7 +89,7 @@ def generate_report(
         print(documented.compile())
 
     result_df = client.execute(documented)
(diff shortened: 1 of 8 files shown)


ajwelch4:
These values should perhaps be in `consts`?

nehanene15:
> These values should perhaps be in `consts`?

Yeah, I would agree with that

ajwelch4:
Change made. I think I got all of them.

nehanene15 (approved):

Issue #376, opened 2022-02-26
Column validation now ignores filters, threshold and labels


ajwelch4:
After [this change](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/345/files#diff-ec3232fc84a73e24bf5cf68f7beb86e56e968c565509c7e07b420d026fd4f09fL143-R185), column validation now ignores filters, threshold and labels.

For example, the following command taken from the README:
```
data-validation validate column   -sc ajw0100-dvt    -tc ajw0100-dvt   -tbls bigquery-public-data.new_york_citibike.citibike_trips    --grouped-columns starttime    --filters "starttime>'2018-05-26':starttime>'2018-05-27'"
```
Now returns:
```
╒══════╤═══════════════════╤═══════════════════╤═══════════════════════════════════════════════════════╤══════════════════════╤════════════════════╤═══════════════════════════════════════════════════════╤══════════════════════╤════════════════════╤═════════════════════════════╤══════════════╤══════════════════╤══════════╕
│      │ validation_name   │ validation_type   │ source_table_name                                     │ source_column_name   │   source_agg_value │ target_table_name                                     │ target_column_name   │   target_agg_value │ group_by_columns            │   difference │   pct_difference │ status   │
╞══════╪═══════════════════╪═══════════════════╪═══════════════════════════════════════════════════════╪══════════════════════╪════════════════════╪═══════════════════════════════════════════════════════╪══════════════════════╪════════════════════╪═════════════════════════════╪══════════════╪══════════════════╪══════════╡
│    0 │ count             │ Column            │ bigquery-public-data.new_york_citibike.citibike_trips │                      │              33402 │ bigquery-public-data.new_york_citibike.citibike_trips │                      │              33402 │ {"starttime": "2015-05-13"} │            0 │                0 │ success  │
├──────┼───────────────────┼───────────────────┼───────────────────────────────────────────────────────┼──────────────────────┼────────────────────┼───────────────────────────────────────────────────────┼──────────────────────┼────────────────────┼─────────────────────────────┼──────────────┼──────────────────┼──────────┤
│    1 │ count             │ Column            │ bigquery-public-data.new_york_citibike.citibike_trips │                      │              55486 │ bigquery-public-data.new_york_citibike.citibike_trips │                      │              55486 │ {"starttime": "2016-07-13"} │            0 │                0 │ success  │
[...]

PR #379, opened 2022-02-26
fix: revert change from #345 that causes filters, threshold and labels to be ignored for column validations (#376)

ajwelch4:
- fix: revert change from #345 that causes filters, threshold and labels to be ignored for column validations (#376)


nehanene15:
Thanks for catching this. @renzokuken Was there a reason for [this change](https://github.com/GoogleCloudPlatform/professional-services-data-validator/commit/3d78ee578b4222a9bdb19c7091445f48f413b9a0#diff-ec3232fc84a73e24bf5cf68f7beb86e56e968c565509c7e07b420d026fd4f09fL143)?

nehanene15 (approved):

PR #383, opened 2022-03-03
feat: Issue356 db2 test

### LINKED ISSUE #356: Add external support for DB2 connections
We have a method to connect to DB2 via the [third_party/ibis/ibis_DB2](https://github.com/GoogleCloudPlatform/professional-services-data-validator/tree/develop/third_party/ibis/ibis_DB2) directory, but we don't support connecting through the CLI. 

This needs to be built out and tested

wanting4:
Add DB2 connection to DVT. Connection works


third_party/ibis/ibis_DB2/compiler.py
```diff
@@ -363,6 +363,29 @@ def reduction_compiler(t, expr):
...
+    return formatter
+
+def _reduction_format(t, sa_func, where, arg, *args):
+    if where is not None:
+        arg = t.translate(where.ifelse(arg, None))
+    else:
+        arg = t.translate(arg)
+
+    #Db2 doesn't allow '*' to be parameterized, probably better way to fix this... 
+    if arg == '*':
```
dhercher, line 385: Is is likely to break anything it touches?

I believe you could pull the column list from arg, but failing that perhaps it makes more sense to raise an exception

dhercher (dismissed):

third_party/ibis/ibis_DB2/compiler.py
ngdav: It should have a fairly limited impact. It only runs for count functions and only if '*' is returned as a parameter. It is tested to work with multiple column counts.

ngdav: The issue is that in other DBMS, you can bind * to a parameter marker, which isn't allowed in aggregation functions in Db2. The ibis code currently doesn't take this into account. This should only run in the case of count(), count() should have only one parameter, so this should have a limited impact. Replacing * with None instead causes the sqlalchemy layer to correctly substitute count(\*) explicitly instead of trying to bind * to a parameter which doesn't work in Db2. 
Pulling a column list and counting each column separately is not exactly the same as count(\*) and would not be an expected outcome.

renzokuken:
@ngdav  it looks like the integration tests are currently failing due to a stale version of our linting library which has been updated in the main branch. Please merge from develop and commit. Thanks!

nehanene15 (approved):
LGTM besides one doc update - we should add DB2 support to the top level README as well:
https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/README.md?plain=1#L22

PR #390, opened 2022-03-14
feat: Support for custom query

### LINKED ISSUE #315: Enable direct comparison of raw SQL outputs
This feature would allow a user to run a raw query on both source and target and compare the results. The first iteration of this issue will be column level validations (count, sum, min, max). BigQuery and Hive are the highest priority connection types that we should support.

Robby29:
Custom query support for column level aggregations. Adds custom query CLI options to the existing validation options.
Closes #315


data_validation/cli_tools.py
```diff
@@ -530,6 +533,85 @@ def _configure_schema_parser(schema_parser):
...
+    _add_common_arguments(custom_query_parser)
+    custom_query_parser.add_argument(
+        "--source-query-file",
+        "-sqf",
+        help="File containing the source sql commands",
+    )
+    custom_query_parser.add_argument(
+        "--target-query-file",
+        "-tqf",
+        help="File containing the source sql commands",
```
nehanene15: Typo - The help description should say 'target sql commands'

data_validation/data_validation.py
```diff
@@ -357,3 +359,11 @@ def combine_data(self, source_df, target_df, join_on_fields):
                 rsuffix=consts.OUTPUT_SUFFIX,
             )
         return df
+
+    def _adjust_schema(self,data_frame):
```
nehanene15: You may not need to do this since PR #392 implements a fix within the Ibis client so that the execute query will not append a 't0' to the column names.

data_validation/validation_builder.py
```diff
@@ -358,3 +389,38 @@ def add_query_limit(self):
...
+
+        if not query or query.isspace():
+            raise ValueError(
+            "Expected file with sql query, got empty file or file with white spaces. "
+            f"input file: {filename}"
+            )
+        file.close()
+        return query
+
+    def get_aggregation_query(self, agg_type, column_name):
```
nehanene15, line 414: @renzokuken Curious on your thoughts on this and if there's a better way to generate this SQL as a string. This is so that the user can run a sum/count/min/max on the source & target query results.

nehanene15 (changes requested):
Couple changes that would be useful:
- Add unit tests, specifically for generating the source/target aggregate query
- Add documentation in the README for this new CLI option

Also, have we tested with user provided CTEs/nested SQL yet?

data_validation/cli_tools.py
Robby29: done

data_validation/data_validation.py
Robby29: removed the above function.

nehanene15 (approved):

PR #392, opened 2022-03-14
feat: Hive hash function support

nehanene15:
Adds Hive hash function and overrides the default `if_null()` function required for the hash sanitization with a `coalesce()` method. Allows Hive row hash validation support.


renzokuken (approved):
LGTM

PR #393, opened 2022-03-18
fix: Status when source and target agg values are 0 

nehanene15:
- Fixes bug so that when the source and target agg values are 0, the validation is marked as success
- Updates documentation for row validation


renzokuken (approved):
LGTM

PR #396, opened 2022-03-21
fix: Comparison fields Key Error fix

### LINKED ISSUE #395: KeyError when using comparison fields
Currently when running a row level validation with comparison fields like so:
`data-validation -v validate row -sc hive_conn -tc hive_conn -tbls default.mascots --comparison-fields market -pk id -rr -rbs 10`
DVT throws a KeyError:
```
"/home/nehanene/professional-services-data-validator/env/lib/python3.7/site-packages/pandas/core/frame.py", line 3024, in __getitem__
    indexer = self.columns.get_loc(key)
  File "/home/nehanene/professional-services-data-validator/env/lib/python3.7/site-packages/pandas/core/indexes/base.py", line 3082, in get_loc
    raise KeyError(key) from err
KeyError: 'id'
```

This is due to the _fetch() method in ImpalaQuery appending a prefix of the table name to the result dataframe.

nehanene15:
Closes #395

### DIFF
--- a/third_party/ibis/ibis_impala/api.py
+++ b/third_party/ibis/ibis_impala/api.py
@@ -100,9 +100,7 @@ def _fetch(self, cursor):
         batches = cursor.fetchall(columnar=True)
         names = []
         for x in cursor.description:
-            name = x[0]
-            if name.startswith('t0.'):
-                name = name[3:]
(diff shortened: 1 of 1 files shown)


Robby29 (approved):
LGTM

PR #402, opened 2022-03-23
feat: adding entrypoint for partial row hashing

renzokuken:


nehanene15:
Added on to this PR with some additional fixes (for tables that have struct columns) in PR #407

PR #404, opened 2022-03-24
Feat: Random Sort for Pandas Queries

dhercher:
The branch adds a new operation for RandomSort logic.  It also implements this logic by injecting a wrapper around the pandas implementation (which should hold up relatively cleanly to Ibis upgrades).

The logic implements BigQuery by overriding the `format_order_by` logic in the SELECT object itself, and is likely to need scrutiny on Ibis upgrade (find me at that time)

Adding new alchemy sources likely just takes adding the correct SQL (eg `RAND()`) which is required for random sorting on the given DB.

### DIFF
--- a/data_validation/query_builder/random_row_builder.py
+++ b/data_validation/query_builder/random_row_builder.py
@@ -101,13 +101,16 @@ class RandomRowBuilder(object):
     def maybe_add_random_sort(
         self, data_client: ibis.client, table: ibis.Expr
     ) -> ibis.Expr:
-        """ Return a randomly sorted query if it is supported for the client."""
+        """Return a randomly sorted query if it is supported for the client."""
         if type(data_client) in RANDOM_SORT_SUPPORTS:
             return table.sort_by(
                 RandomSortKey(RANDOM_SORT_SUPPORTS[type(data_client)]).to_expr()
             )
 
-        logging.warning("Data Client %s Does Not Enforce Random Sort on Sample", str(type(data_client)))
+        logging.warning(
+            "Data Client %s Does Not Enforce Random Sort on Sample",
+            str(type(data_client)),
+        )
         return table
 
 
--- a/data_validation/query_builder/random_row_builder.py
+++ b/data_validation/query_builder/random_row_builder.py
@@ -148,12 +148,13 @@ pandas_util.compute_sorted_frame = compute_sorted_frame
 ##### Override Order By for SQL #####
 #####################################
 
+
 def format_order_by(self):
     if not self.order_by:
         return None
 
     buf = StringIO()
-    buf.write('ORDER BY ')
+    buf.write("ORDER BY ")
 
     formatted = []
     for expr in self.order_by:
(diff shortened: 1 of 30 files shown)


data_validation/query_builder/random_row_builder.py
```diff
@@ -43,6 +92,95 @@ def compile(
...
+        randomly_sorted_table = self.maybe_add_random_sort(data_client, table)
+        query = randomly_sorted_table.limit(self.batch_size)[self.primary_keys]
 
         return query
+
+    def maybe_add_random_sort(
+        self, data_client: ibis.client, table: ibis.Expr
+    ) -> ibis.Expr:
+        """ Return a randomly sorted query if it is supported for the client."""
+        if type(data_client) in RANDOM_SORT_SUPPORTS:
```
nehanene15, line 105: Can we log a warning in case the data client does not support random sort?

tests/system/data_sources/test_bigquery.py
```diff
@@ -340,15 +351,36 @@ def test_cli_find_tables():
...
-#     os.remove(file_path)
+def test_random_row_query_builder():
+    bq_client = clients.get_data_client(BQ_CONN)
+    row_query_builder = random_row_builder.RandomRowBuilder(["station_id"], 10)
+    query = row_query_builder.compile(
+        bq_client, "bigquery-public-data.new_york_citibike", "citibike_stations"
+    )
+
+    random_rows = bq_client.execute(query)
+    print(random_rows)
```
nehanene15: Is there value in keeping the print() statements in this test?

tests/system/data_sources/test_bigquery.py
```diff
@@ -340,15 +351,36 @@ def test_cli_find_tables():
...
+    print(random_rows["station_id"].values)
+    print(query.compile())
+
+    assert query.compile() == EXPECTED_RANDOM_ROW_QUERY
+    assert len(random_rows["station_id"]) == 10
+    assert list(random_rows["station_id"]) != [
```
nehanene15, line 412: How do we know the random rows will not select the following station_ids?

data_validation/query_builder/random_row_builder.py
```diff
@@ -12,15 +12,64 @@
...
+### Adding new datasources should be
+### done by adding the Client and
+### syntax for random sorts in the
+### dict below. If upstream injection
+### is required, feel free to reach
+### out to dhercher
+######################################
+RANDOM_SORT_SUPPORTS = {
+    PandasClient: "NA",
+    BigQueryClient: "RAND()",
```
nehanene15, line 45:
```suggestion
    BigQueryClient: "RAND()",
    ImpalaClient: "RAND()", 
```

Hive uses the RAND() function the same way BQ does. We will also need to import ImpalaClient for this.

nehanene15 (changes requested):
A couple nits, otherwise looks great.

data_validation/query_builder/random_row_builder.py
dhercher: This should be added in a separate PR and tested, so removing from here

data_validation/query_builder/random_row_builder.py
dhercher: done

tests/system/data_sources/test_bigquery.py
dhercher: nope, done

tests/system/data_sources/test_bigquery.py
dhercher: there are 5000 rows and we randomly select 10 in a random order

so odds this fails randomly are roughly 1/10E36

nehanene15 (approved):

PR #405, opened 2022-03-24
fix: replace nan with None in query output dataframes

dmedora:
Initial fix for Ibis turning NULL values in String columns from Hive to 'nan', whereas those from BigQuery are 'None'.


nehanene15:
Will be fixed with PR #406

PR #406, opened 2022-03-24
fix: Impala strings/objects now return None instead of NaN

ngdav:
* Previously Impala strings, chars and other objects filled a dataframe with NaN
instead of NULL which caused mismatch issues when comparing with
BigQuery dataframes.
* Moved _column_batches_to_dataframe and _chunks_to_pandas_array functions from the ibis.backends.impala.client into api.py to override default functionality.

### DIFF
--- a/third_party/ibis/ibis_impala/api.py
+++ b/third_party/ibis/ibis_impala/api.py
@@ -15,10 +15,12 @@
 from ibis.backends.base_sql import fixed_arity
 from ibis.backends.impala import connect, udf
 from ibis.backends.impala.compiler import rewrites
-from ibis.backends.impala.client import ImpalaClient, ImpalaQuery, _column_batches_to_dataframe
+from ibis.backends.impala.client import ImpalaClient, ImpalaQuery
 import ibis.expr.datatypes as dt
 import ibis.expr.operations as ops
 import ibis.expr.schema as sch
+import numpy as np
+import pandas as pd
 
 _impala_to_ibis_type = udf._impala_to_ibis_type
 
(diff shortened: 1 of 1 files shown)


third_party/ibis/ibis_impala/api.py
```diff
@@ -105,6 +107,81 @@ def _fetch(self, cursor):
...
+            pos += k
+
+    if have_nulls:
+        if numpy_type in ('bool', 'datetime64[ns]'):
+            target = np.empty(total_length, dtype='O')
+            na_rep = np.nan
+        elif numpy_type.startswith('int'):
+            target = np.empty(total_length, dtype='f8')
+            na_rep = np.nan
+        elif numpy_type in ('object'):
```
dhercher, line 156: this filter seems likely to be saying "anything" and negating the else below.

Does it work equally well to remove the `object` block and use None in the `else:`?

third_party/ibis/ibis_impala/api.py
```diff
@@ -105,6 +107,81 @@ def _fetch(self, cursor):
...
+    total_length = 0
+    have_nulls = False
+    for c in chunks:
+        total_length += len(c)
+        have_nulls = have_nulls or c.nulls.any()
+
+    type_ = chunks[0].data_type
+    numpy_type = _HS2_TTypeId_to_dtype[type_]
+
+    def fill_nonnull(target, chunks):
```
dhercher, line 127: Everything that is unchanged should be imported rather than re-written IMO, since we plan to push this upstream asap

nehanene15 (approved):

third_party/ibis/ibis_impala/api.py
nehanene15: +1, but in this case it's a nested function so I'm not aware of a way to import that

dmedora: Would be good if `_HS2_TTypeId_to_dtype` and `_column_batches_to_dataframe()` are imported then, but yes I'm also not sure what we can do about this nested function specifically.

third_party/ibis/ibis_impala/api.py
ngdav: The floats are covered in the else block. I'd also rather leave in the else just in case anything else might be added in the future.

Issue #397, opened 2022-03-22
Support column list for row hash validations


nehanene15:
Support including only specific column in row hash validations via CLI.
i.e. `data-validation -v validate row -sc hive_conn -tc bq_conn -tbls default.mascot=pso-kokoro-resources.hivetest.mascot --hash column_a,column_b -pk id`

Currently we only support excluding columns via YAML config.

PR #407, opened 2022-03-25
feat: Hash selective columns

nehanene15:
Closes Issue #397 

Adds support for selective columns for hashing and comparison-fields like so:
`data-validation -v validate row -sc my_bq_conn -tc my_bq_conn -tbls pso-kokoro-resources.hivetest.struct -pk id --hash id,name`

Specifically fixes excluding columns with unsupported pandas data types i.e structs, maps, etc


renzokuken (approved):
remove json files otherwise LGTM

PR #408, opened 2022-03-25
feat: hash filter support

renzokuken:
@tswast this is the hack I've used to get hive hashing to work with filters. For some reason it's passing through base sql instead of SQLAlchemy. I am worried this will have breaking changes as functionality of the tool grows


tswast (approved):
LGTM, assuming tests pass.

nehanene15 (approved):

Issue #412, opened 2022-03-30
Filter results to output only failed (or succeeded) records


FX-HAO:
This is a feature request. It might be good to have a flag to display only failed (or succeeded) report.

nehanene15:
This should be implemented as a `--filter-status` flag that accepts a comma-separated list of status' with the default being 'succes,fail'. This should apply to all types of validations (column, row, schema, etc.)

For example, to filter out only failed records in the output a user would pass in `--filter-status 'fail'`.

The filter itself should be implemented in the result handler and should apply to both text and BigQuery.
This would also require updating the documentation and adding an example of flag usage in docs/examples.

PR #414, opened 2022-03-31
feat: add --filter-status flag

FX-HAO:
#412

### DIFF
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -368,6 +368,12 @@ def _configure_validate_parser(subparsers):
     validate_parser.add_argument(
         "--result-handler-config", "-rc", help="Result handler config details"
     )
+    validate_parser.add_argument(
+        "--filter-status",
+        "-fs",
+        help="filter the validation status, Supported statuses are (success, fail). It defaults "
+        "to return all"
+    )
 
     validate_subparsers = validate_parser.add_subparsers(dest="validate_cmd")
 
--- a/data_validation/combiner.py
+++ b/data_validation/combiner.py
@@ -39,6 +39,7 @@ def generate_report(
     target,
     join_on_fields=(),
     is_value_comparison=False,
+    filter_status=None,
     verbose=False,
 ):
     """Combine results into a report.
@@ -91,6 +92,9 @@ def generate_report(
 
     result_df = client.execute(documented)
     result_df.status.fillna(consts.VALIDATION_STATUS_FAIL, inplace=True)
+
+    if filter_status:
+        result_df = result_df[result_df.status == filter_status]
 
     return result_df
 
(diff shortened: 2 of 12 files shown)


FX-HAO:
@dhercher updated with the latest code. please rerun the tests

data_validation/cli_tools.py
```diff
@@ -376,6 +376,12 @@ def _configure_validate_parser(subparsers):
     validate_parser.add_argument(
         "--result-handler-config", "-rc", help="Result handler config details"
     )
+    validate_parser.add_argument(
```
dhercher, line 379: I'm not sure about the binary (single value) design of this field.  Particularly, it's limiting going forward as it makes adding a third status (eg warning) difficult to fit into the design.

The value of ALL also ends up being an implicit specification throughout the entire codebase which can be difficult for readability and testing.

I would prefer if this flag allowed a comma separated list of status' and the default was `success,fail`.

data_validation/combiner.py
```diff
@@ -89,7 +90,11 @@ def generate_report(
         print(documented.compile())
 
     result_df = client.execute(documented)
-    result_df.validation_status.fillna(consts.VALIDATION_STATUS_FAIL, inplace=True)
+    result_df.status.fillna(consts.VALIDATION_STATUS_FAIL, inplace=True)
+
+    if filter_status:
```
dhercher, line 95: I'm not entirely sure this logic belongs in the combiner.  The status' are really more related to what is being served not what is being combined, and in that sense would make more sense in the result handler.

This would potentially also allow a result handler to make more tactical decisions, such as logging all results but only displaying/distributing the filtered results.

let me know what you think

setup.py
```diff
@@ -18,7 +18,7 @@
 
 name = "google-pso-data-validator"
 description = "A package to enable easy data validation"
-version = "1.7.0"
+version = "1.7.1"
```
dhercher, line 21: versions shouldn't be upgraded here or may need a merge

dhercher (changes requested):

data_validation/combiner.py
fuxinhao-crypto: okay, sounds reasonable. then should we filter the result in `TextResultHandler#print_formatted_()` or `TextResultHandler#execute()`?

FX-HAO: okay, sounds reasonable. then should we filter the result in `TextResultHandler#print_formatted_()` or `TextResultHandler#execute()`?

nehanene15:
@FX-HAO Any updates on this PR? We have seen some other users also request a `--feature-status` flag for column and row validations. Thanks!

nehanene15:
@FX-HAO As this feature has been requested by other users and the PR has gone stale, we will need to re-assign this issue and create a new PR to add this feature. Thanks!

PR #420, opened 2022-04-04
feat!: Changed result schema 'status' column to 'validation_status'

### LINKED ISSUE #417: Update result schema 'status' column to 'validation_status' to avoid duplicate column names
Currently if you validate a table with a column named 'status', Ibis will throw the following error: `ibis.common.exceptions.IntegrityError: Duplicate column name(s): ['status']`

Since 'status' is a common column name, we should update the result handler column name to 'validation_status' to avoid this error.

Raniksingh:
Changed result schema 'status' column to 'validation_status' to resolve duplicate column error.
Closes #417


nehanene15:
This change isn't backwards compatible, since it will break users' existing commands since DVT will now try to write to BQ with a different schema. We should catch the BQ load error and raise an exception with a descriptive error message describing that we have updated the schema and the user will need to update that on BigQuery.

ryanmcdowell:
> 

To this point... can we give the customer the actual query to execute in order to perform this upgrade. Since BigQuery doesn't currently support a `RENAME` clause for columns, we should make this as simple as possible for end users. Ideally just executing a provided script or template SQL.

nehanene15 (changes requested):
As per [Ryan's comment](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/420#issuecomment-1088956435), let's document the actual query users will have to run to rename this column. 

We can add this script under data_validation/samples/bq_utils/update_schema.sql and we can reference this script in the [error message](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/420/files#diff-87c04411a3636ea41fdf3062594a190cefbc4c35eca5ecf4387401c663fb9ca1R66).

Raniksingh:
@nehanene15  Added the script data_validation/samples/bq_utils/update_schema.sh to update bigquery schema from 'status' to 'validation_status'

data_validation/result_handlers/bigquery.py
```diff
@@ -59,6 +59,14 @@ def execute(self, config, result_df):
             table, result_df
         )
         if any(chunk_errors):
+            if (
+                chunk_errors[0][0]["errors"][0]["message"]
+                == "no such field: validation_status."
+            ):
+                raise RuntimeError(
+                    f"Please update your BigQuery results table schema using the script : data_validation/samples/bq_utils/update_schema.sh.\n"
```
nehanene15: This path is wrong - should be `samples/bq_utils/update_schema.sh`

samples/bq_utils/update_schema.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+#This script helps to perform an upgrade of column 'STATUS' to 'VALIDATION_STATUS' without losing existing data in Results table.
```
nehanene15:
```suggestion
# This script helps update the column name of the DVT results table from 'status' to 'validation_status' without losing existing data in the results table as per PR #420
```

samples/bq_utils/update_schema.sh
```diff
@@ -0,0 +1,19 @@
...
+#It takes backup of the existing biqquery table 'RESULTS' and creates a new table with column name as 'VALIDATION_STATUS'
```
nehanene15:
```suggestion
# It makes a copy of the existing BigQuery results table and creates a new table with the updated schema. The new table keeps the original table name to avoid downstream errors.
```

samples/bq_utils/update_schema.sh
```diff
@@ -0,0 +1,19 @@
...
+
+#PROJECT=pso-kokoro-resources
```
nehanene15: Can we delete this line since it's un-used?

samples/bq_utils/update_schema.sh
```diff
@@ -0,0 +1,19 @@
...
+DEPRECATED_TABLE=pso_data_validator.results_deprecated #--Deprecated results table with column name 'status'
```
nehanene15: nit: Can we update the comments to remove the '--'?

samples/bq_utils/update_schema.sh
```diff
@@ -0,0 +1,19 @@
...
+CURRENT_TABLE=pso_data_validator.results #--Current results table
+DEST_TABLE=pso_data_validator.results #--Destination results table with new column name 'validation_status'
+
+#To copy the existing table
```
nehanene15: nit: Can we add spaces after the '#'? i.e '# To copy the ...'

nehanene15 (changes requested):
Small changes in bash script + wrong path provided in error message.

samples/bq_utils/update_schema.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+#This script helps to perform an upgrade of column 'STATUS' to 'VALIDATION_STATUS' without losing existing data in Results table.
+#It takes backup of the existing biqquery table 'RESULTS' and creates a new table with column name as 'VALIDATION_STATUS'
+
+#PROJECT=pso-kokoro-resources
+DEPRECATED_TABLE=pso_data_validator.results_deprecated #--Deprecated results table with column name 'status'
+CURRENT_TABLE=pso_data_validator.results #--Current results table
```
ryanmcdowell: Forgive my ignorance... 

Since the `-bqrh` flag allows the user to specify the output table in `PROJECT_ID.DATASET.TABLE` format... wouldn't it be better to take this an input?

```
./update_schema.sh <current-bq-result-table>
```

nehanene15: True, although we aren't running the script for them just pointing them to the script in the error message.

Since we already will require the user to specify a env var for the deprecated table name, I think it would require the same amount of user effort to specify 2 env vars versus 1 env var and a script input.

nehanene15 (approved):

PR #422, opened 2022-04-04
chore: replace print statements with logging

### LINKED ISSUE #419: Implement uniform logging
DVT uses print statements and logging simultaneously. For example, we usually [print logs when verbose is enabled](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/data_validation/data_validation.py#L277), but in some cases we use the [python logging](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/data_validation/config_manager.py#L465).

We should standardize to only use the [Python logger](https://docs.python.org/3/howto/logging.html#when-to-use-logging) and categorize logs by level (DEBUG, INFO, WARNING, etc.)

Sync271:
closes #419

### DIFF
--- a/data_validation/__main__.py
+++ b/data_validation/__main__.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
 import os
 import json
 import sys
@@ -479,9 +480,9 @@ def main():
     elif args.command == "configs":
         run_validation_configs(args)
     elif args.command == "find-tables":
-        print(find_tables_using_string_matching(args))
+        logging.info(find_tables_using_string_matching(args))
     elif args.command == "query":
-        print(run_raw_query_against_connection(args))
+        logging.info(run_raw_query_against_connection(args))
     elif args.command == "validate":
         validate(args)
     elif args.command == "deploy":
--- a/data_validation/clients.py
+++ b/data_validation/clients.py
@@ -16,6 +16,7 @@
 import pandas
 import warnings
 import copy
+import logging
 
 import google.oauth2.service_account
 
@@ -181,7 +182,7 @@ def get_all_tables(client, allowed_schemas=None):
         try:
             tables = list_tables(client, schema_name)
         except Exception as e:
-            print(f"List Tables Error: {schema_name} -> {e}")
+            logging.exception(f"List Tables Error: {schema_name} -> {e}")
             continue
 
         for table_name in tables:
--- a/data_validation/app.py
+++ b/data_validation/app.py
@@ -17,6 +17,7 @@ import os
 from data_validation import data_validation
 import flask
 import pandas
+import logging
 
 app = flask.Flask(__name__)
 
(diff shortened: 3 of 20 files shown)


data_validation/__main__.py
```diff
@@ -479,9 +480,9 @@ def main():
     elif args.command == "configs":
         run_validation_configs(args)
     elif args.command == "find-tables":
-        print(find_tables_using_string_matching(args))
+        logging.info(find_tables_using_string_matching(args))
     elif args.command == "query":
```
latika-wadhwa, line 484: We could standardize the logging for standardize and easier debugging
for instance the above could be changed to 

logging.info(f"LInfo: {find_tables_using_string_matching(args)}")

Similarly we could use -
LWarn for Warnings
LExcept for Exceptions and so on

data_validation/__main__.py
```diff
@@ -479,9 +480,9 @@ def main():
     elif args.command == "configs":
         run_validation_configs(args)
     elif args.command == "find-tables":
-        print(find_tables_using_string_matching(args))
+        logging.info(find_tables_using_string_matching(args))
```
latika-wadhwa, line 483: We will need to set level for logging to 'INFO' for it to display the logging.info messages too. The format is to get the timestamps and the logging level along with the messages.
    logging.basicConfig(level=logging.INFO, format='%(asctime)s-%(levelname)s: %(message)s')

@nehanene15 - Could you please validate the format.

data_validation/clients.py
```diff
@@ -181,7 +182,7 @@ def get_all_tables(client, allowed_schemas=None):
         try:
             tables = list_tables(client, schema_name)
         except Exception as e:
-            print(f"List Tables Error: {schema_name} -> {e}")
+            logging.exception(f"List Tables Error: {schema_name} -> {e}")
             continue
```
latika-wadhwa, line 186: My understanding is this shouldn't be an exception level. We could use warning or info

latika-wadhwa (changes requested):

data_validation/__main__.py
latika-wadhwa: Also, we could add a variable coming out of a config file to for level in - logging.basicConfig(level=logging.INFO, format='%(asctime)s-%(levelname)s: %(message)s'). This will help us change it to different Level without any code changes.

data_validation/clients.py
dhercher: Warning would make sense to me

data_validation/__main__.py
dhercher: I'm worried about this change.  This is not actually a log, it is the output of the command.

Personally, I use this response to build validations for a whole schema when i know the names will be very similar.

In reality, my usage should be a feature which gets pulled into the validation command.  For now, changing this print is actually an API change and we should hold off until the feature for this is built.

nehanene15:
@Sync271 Following up on this PR - We'll need to pull the latest code and make the requested code changes to merge this

Sync271:
@nehanene15 you should probably take this away from me, won't be able to contribute.

data_validation/__main__.py
latika-wadhwa: Makes sense! Will hold on for these!

nehanene15:
This has been merged with PR #543. Thanks!

PR #429, opened 2022-04-07
fix: Bug with duplicate labels flag introduced with new PR merge

nehanene15:
This fixes a bug introduced where two parses register the `--labels` flag.


dmedora (approved):
LGTM

PR #430, opened 2022-04-07
feat: Allow aggregation over length of string columns

### LINKED ISSUE #424: Support aggregate validation on string columns with length(string_col)
DVT should support column level aggregates (sum, min, max, avg) on the length of string columns. This is supported via calculated fields in a YAML config, but it should also be supported through the CLI. 

Running a '*' aggregation should run over all _numeric_ columns.
`data-validation validate column -sc my_bq_conn -tc my_bq_conn -tbls pso-kokoro-resources.hivetest.mascot -sum '*'`

But specifying a certain string column name should implicitly convert run the aggregation over the length(string_col) 
`data-validation validate column -sc my_bq_conn -tc my_bq_conn -tbls pso-kokoro-resources.hivetest.mascot -sum name`

nehanene15:
Closes #424

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,8 +94,9 @@ With this flag, if a mismatch was found, DVT will dive deeper into the slice
 with the error and find the row (primary key value) with the inconsistency.
 
 You can specify a list of string columns for aggregations in order to calculate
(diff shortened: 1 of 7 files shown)


dmedora (approved):
LGTM

PR #440, opened 2022-04-12
feat!: Adds custom query row level hash validation feature.

### LINKED ISSUE #387: Custom query validation for row level hash
Ability to provide custom SQL queries and run a row level hash comparison on the results.

Robby29:
This PR adds the custom query row level hash validation feature with current support for Biqquery and Hive data sources.
Close #387


README.md
```diff
@@ -230,9 +230,9 @@ data-validation (--verbose or -v) validate schema
                         Defaults  to table.
 ```
 
-#### Custom Query Validations
+#### Custom Query Column Validations
```
dhercher: remove whitespace

README.md
```diff
@@ -273,6 +273,48 @@ The [Examples](docs/examples.md) page provides few examples of how this tool can
 used to run custom query validations.
 
 
+#### Custom Query Row Validations 
+
+(Note: Row hash validation is currently only supported for BigQuery and Imapala/Hive)
```
dhercher: maybe bold the note, parens look slight off in docs

data_validation/data_validation.py
```diff
@@ -291,10 +291,19 @@ def _execute_validation(self, validation_builder, process_in_memory=True):
             if self.config_manager.validation_type == consts.ROW_VALIDATION
             else set(validation_builder.get_group_aliases())
         )
+        if (
+            self.config_manager.validation_type == consts.CUSTOM_QUERY
+            and "hash__all" in self.run_metadata.validations
+        ):
+            join_on_fields = set(["hash__all"])
 
         # If row validation from YAML, compare source and target agg values
```
dhercher, line 300: not related to this PR, but IMO YAML / cli considerations should not make there way into the core logic of DVT

data_validation/cli_tools.py
```diff
@@ -609,6 +609,12 @@ def _configure_custom_query_parser(custom_query_parser):
         "-pk",
         help="Comma separated list of primary key columns 'col_a,col_b'",
     )
+    custom_query_parser.add_argument(
```
dhercher, line 618: I don't exactly understand this param from the description.  Does it apply across all agg types (sum, max, min)?

What is the purpose of the flag?

data_validation/query_builder/custom_query_row_builder.py
```diff
@@ -0,0 +1,214 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+""" The QueryBuilder for building custom query row validation."""
+
+
+class CustomQueryRowBuilder(object):
+    def __init__(self):
+        """Build a CustomQueryRowBuilder objct which is ready to build a custom query row validation nested query."""
```
dhercher: object not objct

data_validation/query_builder/custom_query_row_builder.py
```diff
@@ -0,0 +1,214 @@
...
+        )
+        rstrip_query = self.compile_rstrip_df_fields(calculated_columns, ifnull_query)
+        upper_query = self.compile_upper_df_fields(calculated_columns, rstrip_query)
+        concat_query = self.compile_concat_df_fields(
+            calculated_columns, upper_query, client_config
+        )
+        sha2_query = self.compile_sha2_df_fields(concat_query, client_config)
+        return sha2_query
+
+    def get_table_exapression(self, input_query, client_config):
```
dhercher: `expression`

dhercher (dismissed):

data_validation/cli_tools.py
Robby29: The way custom query logic has been written is that it uses most of the flow of the existing column and row validations , so in order to support that logic we need to map all the arguments to custom query as well. However in the README I have explicitly mentioned which arguments are supported.

README.md
Robby29: Done.

README.md
Robby29: Done.

data_validation/cli_tools.py
Robby29: @nehanene15 Should we change the description to something more intuitive as per Dylan's comment?

nehanene15: @dhercher This was added in PR #430 which allows for aggregation on len(string_col) on all agg types. The flag indicates if the user wants to include string columns when running a wildcard agg i.e `--sum '*'`

data_validation/combiner.py
```diff
@@ -210,7 +209,11 @@ def _calculate_differences(
 
 def _pivot_result(result, join_on_fields, validations, result_type):
     all_fields = frozenset(result.schema().names)
-    validation_fields = all_fields - frozenset(join_on_fields)
+    validation_fields = (
+        all_fields - frozenset(join_on_fields)
+        if "hash__all" not in join_on_fields
```
nehanene15, line 226: Instead of checking if `hash__all` exists, we should have a flag that indicates custom query column versus row i.e.`validate custom-query row` or `validate custom-query column`
This way someone can't provide a `--sum` and a `--hash` in the same validation which would break

data_validation/validation_builder.py
```diff
@@ -337,15 +338,21 @@ def get_source_query(self):
...
                 )
-            source_aggregate_query = self.get_wrapper_aggregation_query(
-                source_aggregate_query, source_input_query
-            )
-            query = self.source_client.sql(source_aggregate_query)
+                query = self.source_client.sql(calculated_query)
+            else:
+                source_aggregate_query = "SELECT "
+                for aggregate in self.config_manager.aggregates:
+                    source_aggregate_query += self.get_aggregation_query(
```
nehanene15: Ideally `get_aggregation_query` should also be in the CustomQueryBuilder() class. There can be a CustomQueryBuilder().compile_row_query() and another for CustomQueryBuilder().compile_column_query().

This way, all custom query logic can be in the Custom query builder

nehanene15 (changes requested):
Two architectural suggestions: Adding a `custom query row/column` in the CLI, and moving custom query column logic to the CustomQueryBuilder() class.

data_validation/cli_tools.py
dhercher: maybe clarify len(string) in the help text?

data_validation/combiner.py
Robby29: Done.

data_validation/validation_builder.py
Robby29: Done.

README.md
```diff
@@ -273,6 +276,51 @@ The [Examples](docs/examples.md) page provides few examples of how this tool can
 used to run custom query validations.
 
 
+### Custom Query Row Validations 
+
+#### (Note: Row hash validation is currently only supported for BigQuery and Imapala/Hive)
```
nehanene15: And Teradata!

data_validation/cli_tools.py
```diff
@@ -537,6 +537,11 @@ def _configure_schema_parser(schema_parser):
 def _configure_custom_query_parser(custom_query_parser):
     """Configure arguments to run custom-query validations."""
     _add_common_arguments(custom_query_parser)
+    custom_query_parser.add_argument(
```
nehanene15, line 540: We should make this a required argument with argparse `required=True`

data_validation/combiner.py
nehanene15, line 226: Should we check if custom_query_type = row field instead of if hash_all not in join_on_fields

data_validation/config_manager.py
```diff
@@ -150,6 +150,17 @@ def append_query_groups(self, grouped_column_configs):
             self.query_groups + grouped_column_configs
         )
 
+    @property
+    def custom_query_type(self):
+        """Return SQL Query File from Config"""
+        return self._config.get(consts.CONFIG_CUSTOM_QUERY_TYPE, [])
```
nehanene15: Shouldn't the default value here be 'None' instead of an empty list? The custom_query_type should only allow two string values: "row" or "column"

data_validation/config_manager.py
```diff
@@ -150,6 +150,17 @@ def append_query_groups(self, grouped_column_configs):
...
+
+    def append_custom_query_type(self, custom_query_type):
```
nehanene15, line 158: We shouldn't need to append the type to a list since custom_query_type should be a string value similar to [validation_type](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/data_validation/config_manager.py#L88)

nehanene15 (changes requested):
One change requested re: how we store the custom_query_type property as a string instead of a list

data_validation/combiner.py
Robby29: So I am actually setting this specifically in the dat_validation function. For rest all other conditions I am using if custom_query = row/column check

data_validation/config_manager.py
Robby29: Done

data_validation/config_manager.py
Robby29: Done

README.md
Robby29: Done

data_validation/cli_tools.py
Robby29: Done

README.md
```diff
@@ -246,7 +246,10 @@ data-validation (--verbose or -v) validate custom-query
                         Comma separated list of tables in the form schema.table=target_schema.target_table
                         Target schema name and table name are optional.
                         i.e 'bigquery-public-data.new_york_citibike.citibike_trips'
-  --source-query-file SOURCE_QUERY_FILE, -sqf SOURCE_QUERY_FILE
+  --custome-query-type CUSTOM_QUERY_TYPE, -cqt CUSTOM_QUERY_TYPE
```
nehanene15: typo

README.md
```diff
@@ -273,6 +276,51 @@ The [Examples](docs/examples.md) page provides few examples of how this tool can
...
+                        Source connection details
+                        See: *Data Source Configurations* section for each data source
+  --target-conn or -tc TARGET_CONN
+                        Target connection details
+                        See: *Connections* section for each data source
+  --tables-list or -tbls SOURCE_SCHEMA.SOURCE_TABLE=TARGET_SCHEMA.TARGET_TABLE
+                        Comma separated list of tables in the form schema.table=target_schema.target_table
+                        Target schema name and table name are optional.
+                        i.e 'bigquery-public-data.new_york_citibike.citibike_trips'
+  --custome-query-type CUSTOM_QUERY_TYPE, -cqt CUSTOM_QUERY_TYPE
```
nehanene15: typo

nehanene15 (approved):

README.md
Robby29: Done

README.md
Robby29: Done

data_validation/cli_tools.py
Robby29: Leaving this as is in this PR. Will change this if needed in a new PR

data_validation/query_builder/custom_query_row_builder.py
Robby29: Done.

PR #442, opened 2022-04-12
feat: Implement sum/avg/bit_xor aggs for Timestamp

### LINKED ISSUE #432: Support SUM/AVG on timestamp fields
DVT should convert timestamps to epoch seconds which will allow for the sum/avg of timestamps when provided in the CLI.

`data-valdiation validate column -sc bq_conn -tc bq_conn -tbls project.dataset.table -sum timestamp_col`

dmedora:
Closes #432


data_validation/__main__.py
```diff
@@ -53,7 +53,15 @@ def get_aggregate_config(args, config_manager):
         config_manager (ConfigManager): Validation config manager instance.
     """
     aggregate_configs = [config_manager.build_config_count_aggregate()]
-    supported_data_types = ["float64", "int8", "int16", "int32", "int64", "decimal"]
+    supported_data_types = [
```
nehanene15, line 56: Can we delete the `supported_data_types.append("timestamp")` for the min/max aggregates?

nehanene15 (approved):
LGTM, besides one comment about lines that need to be deleted

data_validation/__main__.py
dmedora: Lines removed, thanks for catching that.

Issue #441, opened 2022-04-12
Hive boolean converts to 'nan' in dataframe while BigQuery converts to 'None'


nehanene15:
When comparing Hive to BigQuery, if a boolean value is NULL, the value is expressed as 'nan' in Hive and 'None' in BigQuery, making the comparison a mismatch. 

This occurs when grouping by a boolean value.

Command to reproduce:
`data-validation -v validate column -sc hive_conn2 -tc bq_conn -tbls source=target --grouped-columns boolean_col`

PR #444, opened 2022-04-13
fix: Hive boolean nan to None, Unsupported ibis data types in structs and arrays

nehanene15:
Closes Issue #441 
- Converts NULL boolean values in Hive from returning 'nan' to 'None' to be compatible with BigQuery boolean NULLs
- Adds support for structs with unsupported ibis data types which resolves the `'{!r} cannot be parsed as a datatype'.format(value) ibis.common.exceptions.IbisTypeError` when running a column aggregation on a table with struct fields (the validation itself should exclude the struct fields)

### DIFF
--- a/third_party/ibis/ibis_impala/api.py
+++ b/third_party/ibis/ibis_impala/api.py
@@ -61,7 +61,14 @@ def parse_type(t):
             else:
                 return ValueError(t)
         elif "struct" in t or "array" in t or "map" in t:
-            return t.replace("int", "int32")
+            if "bigint" in t:
+                return t.replace("bigint", "int64")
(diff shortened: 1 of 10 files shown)


third_party/ibis/ibis_impala/api.py
```diff
@@ -61,7 +61,14 @@ def parse_type(t):
             else:
                 return ValueError(t)
         elif "struct" in t or "array" in t or "map" in t:
-            return t.replace("int", "int32")
+            if "bigint" in t:
```
dmedora, line 64: Should arrays of strings be accounted for here as well?

dmedora (approved):

Issue #452, opened 2022-04-19
NULL datetime or timestamp values result in errors


nehanene15:
When running a column validation on datetime or timestamp columns, NULL values result in the following error:
`ibis.common.exceptions.IbisTypeError: Value <class 'ibis.expr.datatypes.Interval'> is not a valid datatype`

Command to reproduce:
```
data-validation validate column -sc my_bq_conn -tc my_bq_conn -tbls pso-kokoro-resources.hivetest.test_timestamp -min order_date
```
where order_date is a DATETIME column with all NULL values. DVT needs to build a check for NULL values before comparing the query results. The immediate use case is comparing a Hive TIMESTAMP col to a BQ DATETIME col.

PR #460, opened 2022-04-28
fix: supports NULL datetime/timestamps in Hive to BQ

nehanene15:
Closes Issue #452 

- Allows NULL Hive datetime/timestamp to be compared in validations. Previously DVT would throw a `ibis.common.exceptions.IbisTypeError: Value <class 'ibis.expr.datatypes.Interval'> is not a valid datatype` error
- Fixes bug with "status" in system tests in PR #455


Robby29 (approved):
LGTM

Raniksingh (approved):

Issue #416, opened 2022-04-01
Allow for casting on columns before aggregations/validation functions


nehanene15:
Currently we allow for [casting through YAML](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/docs/examples.md#sample-yaml-file-grouped-column-validation) after an aggregation is complete. For example, we can do a ` CAST(SUM(column)) `. 

We should be able to also cast the field before any aggregations/functions are applied like so: `SUM(CAST(column))`. We can do this through calculated fields. Currently calc fields only support the [default cast to string](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/data_validation/query_builder/query_builder.py#L331). This should be configurable.

PR #461, opened 2022-04-29
feat: Support cast to BIGINT before aggregation

nehanene15:
Closes Issue #416 

- Adds support for specifying a default_cast for any column
- Adds flag `--cast-to-bigint` which automatically casts any int32 column into int64 before aggregations. This is useful when running into INT overflow errors.

### DIFF
--- a/data_validation/config_manager.py
+++ b/data_validation/config_manager.py
@@ -459,6 +459,8 @@ class ConfigManager(object):
             calc_func = "length"
         elif column_type == "timestamp":
             calc_func = "epoch_seconds"
+        elif column_type == "int32":
+            calc_func = "cast"
         else:
             raise ValueError(f"Unsupported column type: {column_type}")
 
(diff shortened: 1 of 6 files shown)


data_validation/config_manager.py
```diff
@@ -470,6 +472,9 @@ def append_pre_agg_calc_field(self, column, agg_type, column_type):
             consts.CONFIG_DEPTH: 0,
         }
 
+        if column_type == "int32":
```
dmedora, line 475: Should int8 and int16 also be included?

nehanene15: I was leaving it just to int32 since that's what an INTEGER column in Hive/Teradata translates to in Ibis. We can probably extend it to support int8/int16 if needed

dmedora (approved):

crfacundo:
Hi, when i use --cast-to-bigint i have an error because int64 is not a datatype in teradata. Is there any possibility to cast to bigint datatype?

nehanene15:
> Hi, when i use --cast-to-bigint i have an error because int64 is not a datatype in teradata. Is there any possibility to cast to bigint datatype?

You can see [this example](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/docs/examples.md#sample-yaml-with-calc-fields-cast-to-numeric-before-aggregation) on how to cast to a different data type such as decimal/numeric.

PR #469, opened 2022-05-04
feat: timezone standardization on clients

renzokuken:


third_party/ibis/ibis_teradata/client.py
```diff
@@ -66,7 +66,7 @@ class TeradataClient(SQLClient):
     table_class = TeradataTable
     dialect = compiler.TeradataDialect
 
-    def __init__(self, host, user_name, password, port=1025, logmech='TD2', use_no_lock_tables=False):
+    def __init__(self, host, user_name, password, port=1025, logmech='TD2', use_no_lock_tables=False, convert_to_utc=True):
```
dhercher, line 69: We could consider making this flag `default_time_zone="GMT"` and using that in the SQL below?  Then if someone actually uses a non-utc timezone they could specify that as well

dhercher (changes requested):

nehanene15:
@renzokuken Closing this out since we were unable to set a timezone via config variable. Instead, we suggest casting to a timestamp within a timezone ([reference](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/third_party/ibis/ibis_teradata/datatypes.py#L149
))

PR #470, opened 2022-05-06
feat: support float and decimal types in Hive

nehanene15:
- Added support for DECIMAL types in Hive for column and row level aggregation. Treats them as Decimal(10,0) which is the default
- Added support for FLOAT type in Hive (float32)
- Updated BigQuery NUMERIC to correlate to Decimal(38,9) in Ibis
- Added dependencies to Cloud Run Hive example


renzokuken (approved):
LGTM

PR #488, opened 2022-05-23
feat: Enable sum/avg/bit_xor for BigQuery datetime type

### LINKED ISSUE #451: SUM/AVG support of DATETIME columns
When running a sum/avg on a DATETIME column, DVT throws the following: 400 No matching signature for function UNIX_SECONDS for argument types: DATETIME. Supported signature: UNIX_SECONDS(TIMESTAMP).

The query that is generated uses UNIX_SECONDs which isn't supported for DATETIME columns:
```
SELECT count(*) AS `count`,
       sum(`epoch_seconds__unit_creation_date_time`) AS `sum__epoch_seconds__unit_creation_date_time`,
       avg(`epoch_seconds__unit_creation_date_time`) AS `avg__epoch_seconds__unit_creation_date_time`
FROM (
  SELECT *,
         UNIX_SECONDS(`unit_creation_date_time`) AS `epoch_seconds__unit_creation_date_time`
  FROM `project.db.table`
) t0
```

A possible fix would be to cast datetime columns to timestamp:
`SELECT UNIX_SECONDS(CAST(unit_creation_date_time AS TIMESTAMP))`

dmedora:
Resolves #451. 

All aggregations (sum, avg, bit_xor) involving the BigQuery client will cast any columns with the Ibis Timestamp type to Timestamp before converting to epoch seconds. This is because BigQuery's Datetime type is represented by Ibis as Timestamp and so first needs to be cast to Timestamp.


nehanene15 (approved):
LGTM!

PR #500, opened 2022-06-05
fix: Schema validation to make case insensitive column name comparision

### LINKED ISSUE #385: Schema validation generate wrong result if source column name is lower and target column name is upper
Hi,

I'm try to check schema validation in BigQuery source view vs target view. But in source i have lower name column and in target i have upper case name. If i run validation then i see "column<>COLUMN" and i have status FAILED :)

I made quic change in schema_validation.py and add some 2 lines and working:
```
def schema_validation_matching(source_fields, target_fields):
    """ Compare schemas between two dictionary objects """
    results = []
    # Go through each source and check if target exists and matches
    
   #----------- GW add START
    target_fields = {k.casefold(): v for k, v in target_fields.items()}
    source_fields** = {k.casefold(): v for k, v in source_fields.items()}
   #----------- GW add STOP

    for source_field_name, source_field_type in source_fields.items():
```

Please add a parameter to configuration maybe that I can choose if i want use lower or upper.
In documentation BigQuery - https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#case_sensitivity is see something like that:
`
Category | Case Sensitive?
Table names | Yes
Column names | No
`

Thanks,
Greg

latika-wadhwa:
Closes #385

### DIFF
--- a/data_validation/schema_validation.py
+++ b/data_validation/schema_validation.py
@@ -97,12 +97,26 @@ class SchemaValidation(object):
 def schema_validation_matching(source_fields, target_fields):
     """Compare schemas between two dictionary objects"""
     results = []
+    # Apply the casefold() function to lowercase the keys of source and target
+    source_fields_casefold = dict(
+        {
+            source_field_name.casefold(): source_field_type
+            for source_field_name, source_field_type in source_fields.items()
+        }
+    )
+    target_fields_casefold = dict(
+        {
+            target_field_name.casefold(): target_field_type
+            for target_field_name, target_field_type in target_fields.items()
+        }
+    )
+
     # Go through each source and check if target exists and matches
-    for source_field_name, source_field_type in source_fields.items():
+    for source_field_name, source_field_type in source_fields_casefold.items():
         # target field exists
-        if source_field_name in target_fields:
+        if source_field_name in target_fields_casefold:
             # target data type matches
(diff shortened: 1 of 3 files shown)


data_validation/schema_validation.py
```diff
@@ -97,12 +97,26 @@ def execute(self):
 def schema_validation_matching(source_fields, target_fields):
     """Compare schemas between two dictionary objects"""
     results = []
+    # Apply the casefold() function to lowercase the keys of source and target
+    source_fields_casefold = dict(
```
nehanene15: Do you need the dict()? Just the curly braces will instantiate a dictionary `{
            source_field_name.casefold(): source_field_type
            for source_field_name, source_field_type in source_fields.items()
        }`

nehanene15 (changes requested):
Let's also document under [Schema Validation](https://github.com/GoogleCloudPlatform/professional-services-data-validator#schema-validations) that this will be a case insensitive column match.

data_validation/schema_validation.py
latika-wadhwa: Don't need it. Cleaned it in the next commit.

renzokuken (approved):
LGTM

nehanene15 (approved):

PR #513, opened 2022-06-17
fix: fixing teradata doc

renzokuken:


README.md
```diff
@@ -464,6 +464,9 @@ and run a SHA256() hash and compare the source and target results. Since each ro
 be returned in the result set, it is recommended to utilize the `--use-random-row` feature
 to validate a subset of the table.
 
+Please note that SHA256 is not a supported function on teradata systems. If you wish to perform
```
nehanene15, line 469: Let's also add this disclaimer under the Notes for Row Validation CLI here: https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/README.md?plain=1#L137

nehanene15 (changes requested):
Think we should add the disclaimer in one more spot - otherwise LGTM.
Also, the link redirects to the TD community home page (https://support.teradata.com/community). Not sure if there is a way to avoid that..

nehanene15 (approved):

PR #521, opened 2022-06-23
fix: primary key casting

### LINKED ISSUE #494: Support casting primary keys
There should be support to cast a primary key block in the YAML like so:
```yaml
primary_keys:
  - cast: date
    field_alias: start_time
    source_column: start_time
    target_column: start_time
```

This issue is particularly focusing on a Teradata to BQ validaton

renzokuken:
Closes #494

### DIFF
--- a/data_validation/query_builder/query_builder.py
+++ b/data_validation/query_builder/query_builder.py
@@ -210,6 +210,7 @@ class ComparisonField(object):
         alias = self.alias or self.field_name
         if self.cast:
             comparison_field = comparison_field.cast(self.cast)
+        comparison_field = comparison_field.name(alias)
 
         return comparison_field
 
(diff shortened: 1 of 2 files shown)


data_validation/query_builder/query_builder.py
```diff
@@ -508,7 +512,7 @@ def compile(self, data_client, schema_name, table_name):
                 self.compile_aggregate_fields(filtered_table)
             )
         else:
-            query = grouped_table
+            query = filtered_table
```
nehanene15: Wouldn't we need to return grouped_table in case we want to add aggregate fields or a group_by field?

renzokuken: good catch. Reverting.

nehanene15 (approved):

PR #526, opened 2022-06-28
chore: remove support run command

### LINKED ISSUE #503: Remove support for 'data validation run' command
We have deprecated support for `data validation run --type column` syntax for versions >1.3.2 in support of `data-validation validate column/schema/row/custom-query` command. 

We should clean up the code to remove support for the 'run' option when validating through the CLI.

mloeberc:
changes to cli_tools.py file to remove suppoprt for run command.
Closes #503

### DIFF
--- a/data_validation/__main__.py
+++ b/data_validation/__main__.py
@@ -497,9 +497,7 @@ def main():
     # Create Parser and Get Deployment Info
     args = cli_tools.get_parsed_args()
 
-    if args.command == "run":
-        run(args)
-    elif args.command == "connections":
+    if args.command == "connections":
         run_connections(args)
     elif args.command == "run-config":
         run_config(args)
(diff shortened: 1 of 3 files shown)


nehanene15 (changes requested):
There is still a some downstream references that need to be deleted/updated like so - https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/data_validation/__main__.py#L500 

And here: https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/e3d42cc9a1dea092d15687e9995fa41f18f59aeb/data_validation/__main__.py#L203. We reference args.type which was only supported with the deprecated `data-validation run --type column`

mloeberc:
@nehanene15 Hi Neha, I made the changes and left the validations running. Lete me know if everything is ok now.

helensilva14 (approved):
LGTM!

nehanene15 (approved):

PR #537, opened 2022-07-12
feat: Remove dependency on tables list for custom query 

### LINKED ISSUE #496: Remove dependency on tables list for custom query
A user should be able to omit the `--tbls` flag and only specify the source and target SQL query for custom query validations. This way a user will not have to parse through the SQL to find the table name(s) when validating queries at scale.

sramsrinivasan:
Closes #496

### DIFF
--- a/data_validation/__main__.py
+++ b/data_validation/__main__.py
@@ -245,16 +245,13 @@ def build_config_managers_from_args(args):
     )
 
     is_filesystem = source_client._source_type == "FileSystem"
-    tables_list = cli_tools.get_tables_list(
-        args.tables_list, default_value=[], is_filesystem=is_filesystem
-    )
 
-    for table_obj in tables_list:
+    if args.tables_list is None and config_type == consts.CUSTOM_QUERY:
         config_manager = ConfigManager.build_config_manager(
             config_type,
             args.source_conn,
             args.target_conn,
-            table_obj,
+            {},
             labels,
             threshold,
             format,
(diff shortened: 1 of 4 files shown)


docs/examples.md
```diff
@@ -228,7 +228,7 @@ validations:
 
 #### Run a custom query column validation
 ````shell script
-data-validation validate --custom-query-type column custom-query --source-query-file source_query.sql --target-query-file target_query.sql -sc my_bq_conn -tc my_bq_conn -tbls bigquery-public-data.new_york_citibike.citibike_stations
+data-validation validate custom-query --custom-query-type column  --source-query-file source_query.sql --target-query-file target_query.sql -sc my_bq_conn -tc my_bq_conn -tbls bigquery-public-data.new_york_citibike.citibike_stations
```
helensilva14:
```suggestion
data-validation validate custom-query --custom-query-type column  --source-query-file source_query.sql --target-query-file target_query.sql -sc my_bq_conn -tc my_bq_conn
```

I think you missed to remove the tables list parameter here on the example

data_validation/__main__.py
```diff
@@ -245,16 +245,13 @@ def build_config_managers_from_args(args):
     )
 
     is_filesystem = source_client._source_type == "FileSystem"
-    tables_list = cli_tools.get_tables_list(
-        args.tables_list, default_value=[], is_filesystem=is_filesystem
-    )
 
-    for table_obj in tables_list:
+    if args.tables_list is None and validate_cmd == "Custom-query":
```
```suggestion
    if args.tables_list is None and config_type == consts.CUSTOM_QUERY:
```
It'd be nice to reference the constant here and use the config type variable as it is used on the `config_type != consts.SCHEMA_VALIDATION`

helensilva14 (changes requested):
LGTM! Since the dependency is being removed, the README can be updated on the sections about the custom query parameters:
- https://github.com/GoogleCloudPlatform/professional-services-data-validator#custom-query-column-validations
- https://github.com/GoogleCloudPlatform/professional-services-data-validator#custom-query-row-validations

data_validation/__main__.py
```diff
@@ -266,17 +263,43 @@ def build_config_managers_from_args(args):
...
-            if args.exclusion_columns is not None:
-                exclusion_columns = cli_tools.get_arg_list(args.exclusion_columns)
-                config_manager.append_exclusion_columns(
-                    [col.casefold() for col in exclusion_columns]
-                )
+        config_manager = build_config_from_args(args, config_manager)
 
         configs.append(config_manager)
 
+    else:
```
nehanene15, line 270: There should be a cleaner way to condense this instead of the large conditional statement with very similar code in each block. The cli.get_tables_list() method should already return an empty list if `--tbls` is not provided (default_value = []).

nehanene15 (changes requested):
This change breaks row level custom query validations since it needs the schema/table name when building the config for row validation.

helensilva14 (approved):

PR #538, opened 2022-07-12
fix: support for cast to timestamp in TD, support for random row

### LINKED ISSUE #536: Support random row for Teradata connections
Currently if a user tries to use `-rr` or `-rbs` to add random sampling/random row batch size to a row level validation, the will get the following warning:
"WARNING:root:Data Client <class 'third_party.ibis.ibis_teradata.client.TeradataClient'> Does Not Enforce Random Sort on Sample"

We enforce a LIMIT of 5, but do not randomize the sample. We should add support for random sampling for Teradata by adding support [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/data_validation/query_builder/random_row_builder.py#L44).

nehanene15:
Closes #536 

- Removes warning for Teradata random row validation as it is supported with SAMPLE 
- Adds support for casting to Teradata TIMESTAMP and including timezone cast
- Fixes bug for schema validation that throws error if column is not included in both source and target tables

### DIFF
--- a/data_validation/query_builder/random_row_builder.py
+++ b/data_validation/query_builder/random_row_builder.py
@@ -25,11 +25,15 @@ from ibis.backends.pandas.client import PandasClient
 import ibis.backends.pandas.execution.util as pandas_util
 
 from ibis.expr.signature import Argument as Arg
-from third_party.ibis.ibis_teradata.client import TeradataClient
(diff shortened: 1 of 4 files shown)


renzokuken (approved):
LGTM

PR #541, opened 2022-07-13
feat: Remove dependency on tables list for custom query

### LINKED ISSUE #496: Remove dependency on tables list for custom query

sramsrinivasan:
Closes #496

### DIFF
--- a/data_validation/__main__.py
+++ b/data_validation/__main__.py
@@ -246,7 +246,7 @@ def build_config_managers_from_args(args):
 
     is_filesystem = source_client._source_type == "FileSystem"
     tables_list = cli_tools.get_tables_list(
-        args.tables_list, default_value=[], is_filesystem=is_filesystem
+        args.tables_list, default_value=[{}], is_filesystem=is_filesystem
     )
 
     for table_obj in tables_list:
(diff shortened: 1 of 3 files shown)


sramsrinivasan:
This fix is a not a complete solution to the issues we have with custom query. For all custom-query row validations and custom-query column validations with aggregates (count, sum, max, min etc...), we still need -tbls argument. We will work on a change of design around custom-query next. This fix is intended to avoid any blockers associated with custom-query column validations via source_query.sql and target_query.sql files without having the need to specify -tbls arguement.

data_validation/__main__.py
```diff
@@ -246,7 +246,7 @@ def build_config_managers_from_args(args):
 
     is_filesystem = source_client._source_type == "FileSystem"
     tables_list = cli_tools.get_tables_list(
-        args.tables_list, default_value=[], is_filesystem=is_filesystem
+        args.tables_list, default_value=[{}], is_filesystem=is_filesystem
```
nehanene15, line 249: Is there a reason we need the default value to be a list with an empty dictionary? Can we just keep the default value as an empty list?

sramsrinivasan: when this piece of code is executed "for table_obj in tables_list:" for a default_value=[] it just does nothing as there is nothing to iterate and the return configs just returns an empty list. This was the original issue we had (no code gets executed). Having an empty dictionary at the least ensure the code "for table_obj in tables_list:" is iterated once and the code proceeds forward to run the custom queries listed in source and target files. 

Since its an empty dictionary, we assign table and schema to None in config_manager.py.

nehanene15 (approved):

PR #547, opened 2022-07-22
feat!: Added teradata custom query support

### LINKED ISSUE #501: Custom query row level support for Teradata
Custom query does not support Teradata for row level. Users should be able to pass in a source and target custom query for TD that allows comparing the row hash.

Yashfork:
Added Teradata Custom Query support for row validation

Updated logic to fetch schema from custom query


Yashfork:
In this release we are supporting teradata custom query row validation. The final mapping between hash values in custom query needs to be fixed. This will be taken care next.

nehanene15 (approved):

Raniksingh (approved):

Raniksingh (approved):

PR #564, opened 2022-08-18
feat!: Add Python 3.10 support

ngdav:
*Add python3.10
*remove python3.6 and it was not originally supported anyway.
*remove python3.6 references
*switch to new testing image because the old testing image assumed python3.6


renzokuken (approved):

sramsrinivasan (approved):

docs/installation.md
```diff
@@ -7,7 +7,7 @@ This tool can be natively installed on your machine or can be containerized and
 
 ## Prerequisites
 
-- Any machine with Python 3.6+ installed.
+- Any machine with Python 3.7+ installed.
```
dhercher, line 10: Maybe add a note on which version dropped 3.6 so anyone using it can peg to that if they need to

cloudbuild.yaml
```diff
@@ -18,31 +18,31 @@ options:
     name: 'projects/pso-kokoro-resources/locations/us-central1/workerPools/private-pool'
 steps:
 - id: lint
-  name: 'gcr.io/pso-kokoro-resources/python-multi'
+  name: 'gcr.io/pso-kokoro-resources/python-multi-test:staging'
```
dhercher, line 21: Have we documented how to (re)build `python-multi`?

I would ideally like to install Oracle+ in it for example

dhercher, line 21: Have we documented how to (re)build python-multi?

I would ideally like to install Oracle+ in it for example

dhercher (changes requested):

cloudbuild.yaml
ngdav: I will build a doc for this, I'm working on building custom images for testing starting with Db2, then moving to oracle and sql server.

docs/installation.md
ngdav: It was part of the build dependencies. I'm not sure we ever supported 3.6, this might have been a typo.

cloudbuild.yaml
ngdav: I will build a doc for this, I'm working on building custom images for testing starting with Db2, then moving to oracle and sql server.

dhercher (approved):

PR #566, opened 2022-08-23
feat: adding scaffold for concatenate as a cli operation

renzokuken:


data_validation/config_manager.py
```diff
@@ -660,6 +660,14 @@ def _build_dependent_aliases(self, calc_type, col_list=None):
                 "concat",
                 "hash",
             ]
+        if calc_type == "concat":
```
nehanene15, line 663: Should this be an 'elif'? Because we can't have someone pass in hash _and_ concat flags?

data_validation/cli_tools.py
```diff
@@ -337,6 +337,11 @@ def _configure_row_parser(row_parser):
         "-hash",
         help="Comma separated list of columns for hash 'col_a,col_b' or * for all columns",
     )
+    row_parser.add_argument(
```
nehanene15, line 340: Let's add this to the README. We'll also need to update [this documentation](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/README.md?plain=1#L144) since users can have either hash, comparison-fields, or concat.

data_validation/__main__.py
```diff
@@ -158,6 +166,12 @@ def get_calculated_config(args, config_manager):
                 ["hash__all"], depth=max_depth
             )
         )
+    if args.concat:
```
nehanene15: Should these be 'elif's as well?

data_validation/__main__.py
```diff
@@ -131,12 +131,20 @@ def get_calculated_config(args, config_manager):
         col_list = None if args.hash == "*" else cli_tools.get_arg_list(args.hash)
         fields = config_manager._build_dependent_aliases("hash", col_list)
         aliases = [field["name"] for field in fields]
-
+        config_manager.append_dependent_aliases(aliases)
```
nehanene15: We only need to append dependent aliases if a list of columns is provided. If someone runs `--hash '*'`, they don't want to exclude certain columns and we don't need to trim the final data frame.

I think we should keep this line in the if col_list loop.

nehanene15 (changes requested):

data_validation/config_manager.py
renzokuken: that's a good question. Let me test it out!

renzokuken: does not work. Removed.

README.md
```diff
@@ -172,6 +172,7 @@ data-validation (--verbose or -v) (--log-level or -ll) validate row
                         Comma separated list of columns to compare. Can either be a physical column or an alias
                         See: *Calculated Fields* section for details
   --hash COLUMNS        Comma separated list of columns to hash or * for all columns 
+  --concat COLUMNS        Comma separated list of columns to concatenate or * for all columns
```
nehanene15: nit
```suggestion
  --concat COLUMNS      Comma separated list of columns to concatenate or * for all columns 
```

data_validation/__main__.py
```diff
@@ -131,11 +131,23 @@ def get_calculated_config(args, config_manager):
         col_list = None if args.hash == "*" else cli_tools.get_arg_list(args.hash)
         fields = config_manager._build_dependent_aliases("hash", col_list)
         aliases = [field["name"] for field in fields]
-
+        # config_manager.append_dependent_aliases(aliases)
         # Add to list of necessary columns for selective hashing in order to drop
         # excess columns with invalid data types (i.e structs) when generating source/target DFs
         if col_list:
```
nehanene15, line 137: If a column list was provided, we should append the col_list _and_ the aliases to the dependent aliases. 
If not column list was provided, we don't need to populate dependent_aliases at all. 
So couldn't we keep the original code like so:
```
if col_list:
            config_manager.append_dependent_aliases(col_list)
            config_manager.append_dependent_aliases(aliases)
```

data_validation/__main__.py
```diff
@@ -131,11 +131,23 @@ def get_calculated_config(args, config_manager):
         col_list = None if args.hash == "*" else cli_tools.get_arg_list(args.hash)
         fields = config_manager._build_dependent_aliases("hash", col_list)
         aliases = [field["name"] for field in fields]
-
+        # config_manager.append_dependent_aliases(aliases)
```
nehanene15, line 134: nit: remove comment

data_validation/__main__.py
```diff
@@ -184,16 +202,15 @@ def build_config_from_args(args, config_manager):
...
-            if args.hash != "*":
+            if args.hash != "*" or args.concat != "*":
                 config_manager.append_dependent_aliases(comparison_fields)
 
     if args.primary_keys is not None:
         primary_keys = cli_tools.get_arg_list(args.primary_keys)
         config_manager.append_primary_keys(
             config_manager.build_column_configs(primary_keys)
         )
-        if args.hash != "*":
```
nehanene15, line 195: Shouldn't we do the same thing here i.e `if args.hash != "*" or args.concat != "*"`?
Since this way if the user wants to hash/concat all fields, we won't populate the dependent aliases at all since we don't need it

nehanene15 (approved):
LGTM! Only thing that may be good to add is some documentation on the difference between concat and hash (and how only one of the flags should be provided). I'm worried users may try to use both flags at once unaware that the --hash includes a concat.

PR #572, opened 2022-08-30
fix: support NUMBER with no precision/scale

nehanene15:
Currently, NUMBER in Oracle is only supported when a precision and scale are present in the schema i.e NUMBER (2,2). This adds support for NUMBER without precision/scale.


renzokuken (approved):

dhercher (commented):
looks good

PR #582, opened 2022-09-08
fix: random rows with filter option

### LINKED ISSUE #552: DVT output is not proper and giving ambiguous output with --filters,  --use-random-row and --random-row-batch-size options
We are running DVT with filters, --random-row-batch-size and  --use-random-row(for sampling) options. The source and target tables both are Teradata. The DVT execution is going thru properly. However the output generated is not correct and is ambiguous.
We are trying to run the DVT with 10,000 batch size, but the output  being generated is only for very handful rows(3/5/7).

kanhaPrayas:
Closes #552

### DIFF
--- a/data_validation/query_builder/random_row_builder.py
+++ b/data_validation/query_builder/random_row_builder.py
@@ -28,6 +28,7 @@ from ibis.expr.signature import Argument as Arg
 from typing import List
 from data_validation import clients
 from io import StringIO
+from data_validation.query_builder.query_builder import QueryBuilder
 
 try:
     from third_party.ibis.ibis_teradata.client import TeradataClient
@@ -90,7 +91,9 @@ class RandomRowBuilder(object):
         self.batch_size = batch_size
 
     def compile(
-        self, data_client: ibis.client, schema_name: str, table_name: str
+        self, data_client: ibis.client, 
+        schema_name: str, table_name: str,
+        query_builder: QueryBuilder
     ) -> ibis.Expr:
         """Return an Ibis query object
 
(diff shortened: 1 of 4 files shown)


nehanene15:
Just tested locally - LGTM!

data_validation/query_builder/random_row_builder.py
```diff
@@ -28,6 +28,7 @@
 from typing import List
 from data_validation import clients
 from io import StringIO
+from data_validation.query_builder.query_builder import QueryBuilder
```
dhercher, line 30: Nit can you order these public first then local

data_validation/query_builder/random_row_builder.py
```diff
@@ -100,7 +103,13 @@ def compile(
             table_name (String): The name of the table to query.
         """
         table = clients.get_ibis_table(data_client, schema_name, table_name)
-        randomly_sorted_table = self.maybe_add_random_sort(data_client, table)
+        # Build Query Expressions
+        calc_table = table
```
dhercher: This line looks unnecessary, we can use table in the places we use calctable

dhercher (approved):

data_validation/query_builder/random_row_builder.py
kanhaPrayas: Done

data_validation/query_builder/random_row_builder.py
kanhaPrayas: Modified

nehanene15:
@kanhaPrayas The lint is failing; You'll need to run black and flake8 before merging

PR #588, opened 2022-09-16
feat: Oracle random row

### LINKED ISSUE #584: Support random row validation for Oracle
This will require registering a SQL Alchemy random function inside the random row builder. This would also require ensuring the subquery for random row is compiled correctly.

Includes documentation.

nehanene15:
Supports random row for Oracle.
Closes #584

### DIFF
--- a/data_validation/query_builder/random_row_builder.py
+++ b/data_validation/query_builder/random_row_builder.py
@@ -16,6 +16,7 @@ import random
 import logging
 from typing import List
 from io import StringIO
+import sqlalchemy as sa
 import ibis
 import ibis.expr.operations as ops
 import ibis.expr.types as tz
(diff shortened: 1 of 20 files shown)


data_validation/query_builder/random_row_builder.py
```diff
@@ -16,19 +16,29 @@
...
 from ibis_bigquery import BigQueryClient
 from ibis.backends.impala.client import ImpalaClient
 from ibis.backends.pandas.client import PandasClient
 from ibis.expr.signature import Argument as Arg
 from data_validation import clients
 from data_validation.query_builder.query_builder import QueryBuilder
 
+# If you have an cx_Oracle driver installed
+try:
+    from third_party.ibis.ibis_oracle.client import OracleClient
```
dhercher: I would suggest using `clients.OracleClient` here even though you are using an import as an object in a less than natural way.

Might as well only have this whacky logic contained to one spot

dhercher: ditto for teradata

dhercher (changes requested):
Consider adding some integration testing

nehanene15:
> Consider adding some integration testing

Yeah, I'm waiting on an Oracle instance we can use for testing to build out integration tests for Oracle row + random row. I'll create a new issue to build those out once an image is ready.

dhercher (approved):

Issue #565, opened 2022-08-23
Custom query validation throwing error with sql files ending with semicolon(;)


Raniksingh:
If the source and target files that is being passed for custom query validation has ; in it, the validation fails.

Example : source.sql file --- select * from db.table1;  fails
Whereas source.sql file --- select * from db.table1 runs fine.

File "/Users/ranisingh/Documents/DVT/professional-services-data-validator/env/lib/python3.7/site-packages/teradatasql/__init__.py", line 933, in executemany
    raise OperationalError (sErr)
teradatasql.OperationalError: [Version 17.20.0.1] [Session 1318] [Teradata Database] [Error 3707] Syntax error, expected something like an 'EXCEPT' keyword or an 'UNION' keyword or a 'MINUS' keyword between the word 'DEPARTMENT' and ';'.
 at gosqldriver/teradatasql.formatError ErrorUtil.go:88

PR #590, opened 2022-09-19
fix: Custom query validation throwing error with sql files ending with semicolon(;)

celiajmd:
Closes issue #565 

- remove semicolon in query string


celiajmd:
Code need to be further adjusted.

celiajmd:
Code need to be further adjusted.

PR #593, opened 2022-09-21
feat: New flag to filter results by status in all supported validations

### LINKED ISSUE #412: Filter results to output only failed (or succeeded) records

helensilva14:
Closes #412 

- [x] It doesn't accept unknown statuses
- [x] Filters in the command line output
- [x] Filters the records before saving it to BigQuery
- [x] It's an optional flag, so if not provided all statuses (success, fail) are shown

Tested for:

- [x] Column
- [x] Row
- [x] Schema
- [x] YAML
- [ ] Custom query 
- [ ] Ad hoc SQL


data_validation/result_handlers/text.py
```diff
@@ -25,16 +25,23 @@
 from data_validation import consts
 
 
+def filter_validation_status(status_list, result_df):
+    return result_df[result_df.validation_status.isin(status_list)]
+
+
 class TextResultHandler(object):
-    def __init__(self, format, cols_filter_list=consts.COLUMN_FILTER_LIST):
+    def __init__(self, format, status_list, cols_filter_list=consts.COLUMN_FILTER_LIST):
```
nehanene15: We should add a default value of None `status_list=None` in case we don't always want to provide a status_list upon instantiation

docs/examples.md
```diff
@@ -46,6 +46,11 @@ data-validation validate column -sc my_bq_conn -tc my_bq_conn -tbls bigquery-pub
 data-validation validate row -sc my_bq_conn -tc my_bq_conn -tbls bigquery-public-data.new_york_citibike.citibike_stations --primary-keys station_id --hash '*'
 ````
 
+#### Run a row hash validation for all rows but filters to get only the failed records
```
nehanene15: nit
```suggestion
#### Run a row hash validation for all rows but filter only the failed records 
```

data_validation/__main__.py
```diff
@@ -266,6 +266,14 @@ def build_config_managers_from_args(args):
         args.tables_list, default_value=[{}], is_filesystem=is_filesystem
     )
 
+    filter_status = consts.VALIDATION_STATUSES
```
nehanene15: I think this should be instantiated to 'None' instead of to [success,fail]. This way, if no filters are provided, we don't call the filter_validation_status function in the result handler to avoid extra computation per validation run.

data_validation/__main__.py
```diff
@@ -266,6 +266,14 @@ def build_config_managers_from_args(args):
...
+    if args.filter_status is not None:
+        arg_list = cli_tools.get_arg_list(args.filter_status)
+        if all(arg in filter_status for arg in arg_list):
```
nehanene15: once we instantiate filter_status to None, we need to use consts here instead
```suggestion
        if all(arg in consts.VALIDATION_STATUSES for arg in arg_list):
```

data_validation/config_manager.py
```diff
@@ -286,6 +286,11 @@ def exclusion_columns(self):
         """Return the exclusion columns from Config"""
         return self._config.get(consts.CONFIG_EXCLUSION_COLUMNS, [])
 
+    @property
+    def filter_status(self):
+        """Return filter status list from Config"""
+        return self._config.get(consts.CONFIG_FILTER_STATUS)
```
nehanene15:
```suggestion
        return self._config.get(consts.CONFIG_FILTER_STATUS, None)
```

data_validation/result_handlers/bigquery.py
```diff
@@ -32,13 +33,16 @@ class BigQueryResultHandler(object):
             destination table for results.
     """
 
-    def __init__(self, bigquery_client, table_id="pso_data_validator.results"):
+    def __init__(
+        self, bigquery_client, status_list, table_id="pso_data_validator.results"
```
nehanene15: Let's set a default value of status_list=None here so we don't require a status_list parameter when using BQResultHandler

data_validation/result_handlers/bigquery.py
```diff
@@ -32,13 +33,16 @@ class BigQueryResultHandler(object):
...
+    ):
         self._bigquery_client = bigquery_client
         self._table_id = table_id
+        self._status_list = status_list
 
     @staticmethod
     def get_handler_for_project(
-        project_id, table_id="pso_data_validator.results", credentials=None
+        project_id, status_list, table_id="pso_data_validator.results", credentials=None
```
nehanene15: Same as above, default to None

data_validation/result_handlers/bigquery.py
```diff
@@ -48,20 +52,22 @@ def get_handler_for_project(
...
         info = client_info.get_http_client_info()
         client = bigquery.Client(
             project=project_id, client_info=info, credentials=credentials
         )
-        return BigQueryResultHandler(client, table_id=table_id)
+        return BigQueryResultHandler(client, status_list, table_id=table_id)
 
-    def execute(self, config, result_df):
+    def execute(self, result_df):
+        result_df = filter_validation_status(self._status_list, result_df)
```
nehanene15: We should only filter validation status if self.status_list exists.

data_validation/result_handlers/text.py
```diff
@@ -25,16 +25,23 @@
...
         self.format = format
         self.cols_filter_list = cols_filter_list
+        self.status_list = status_list
 
     def print_formatted_(self, result_df):
         """
         Utility for printing formatted results
         :param result_df
         """
+        result_df = filter_validation_status(self.status_list, result_df)
```
nehanene15: Same as above, only filter if self.status_list is not None

data_validation/result_handlers/text.py
nehanene15: Same as above, default this to None

tests/local_check.sh
```diff
@@ -0,0 +1,8 @@
+# execute it after the venv3-9 activation
```
nehanene15: Nice! I would add our copyright to the top (we have it on all the files in our repo). And maybe another comment on when to use this.

We can also reference this in our CONTRIBUTING.md guide: https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/CONTRIBUTING.md#local-testing

tests/system/data_sources/test_bigquery.py
```diff
@@ -66,6 +66,7 @@
         },
     ],
     consts.CONFIG_FORMAT: "table",
+    consts.CONFIG_FILTER_STATUS: consts.VALIDATION_STATUSES,
```
nehanene15: The default values here will need to be `None` 😢

tests/unit/result_handlers/test_text.py
```diff
@@ -16,11 +16,13 @@
```
nehanene15, line 16: We should add a test here to check that filter_validation_status works as intended.

tests/unit/test_data_validation.py
```diff
@@ -67,6 +67,7 @@
     consts.CONFIG_THRESHOLD: 0.0,
     consts.CONFIG_RESULT_HANDLER: None,
     consts.CONFIG_FORMAT: "table",
+    consts.CONFIG_FILTER_STATUS: consts.VALIDATION_STATUSES,
```
nehanene15: All of the tests will need to update the default value to `None`.

tests/unit/result_handlers/test_text.py
```diff
@@ -16,11 +16,13 @@
...
 from pandas import DataFrame
 
+from data_validation.consts import VALIDATION_STATUSES
+
 SAMPLE_CONFIG = {}
 SAMPLE_RESULT_DATA = [
-    [0, 1, 2, 3, "Column", "source", "target"],
-    [4, 5, 6, 7, "Column", "source", "target"],
-    [8, 9, 10, 11, "Column", "source", "target"],
+    [0, 1, 2, 3, "Column", "source", "target", "success"],
```
nehanene15, line 27: After changing the TextResultHandler() to have a default value of None, you won't need to change this test data or the `test_columns_to_print` test.

nehanene15 (dismissed):

tests/unit/result_handlers/test_text.py
helensilva14: I needed to change this data to have at least one failed result for the new unit test

data_validation/result_handlers/bigquery.py
```diff
@@ -48,20 +55,25 @@ def get_handler_for_project(
...
                 Explicit credentials to use in case default credentials
                 aren't working properly.
+            status_list (list): provided status to filter the results with
         """
         info = client_info.get_http_client_info()
         client = bigquery.Client(
             project=project_id, client_info=info, credentials=credentials
         )
-        return BigQueryResultHandler(client, table_id=table_id)
+        return BigQueryResultHandler(client, status_list, table_id=table_id)
```
nehanene15: nit: to keep the same format, we can pass in the keyword arg for status_list or remove keyword args from table_id:

return BigQueryResultHandler(client, status_list=status_list, table_id=table_id)

tests/unit/result_handlers/test_bigquery.py
```diff
@@ -29,7 +29,7 @@ def test_get_handler_for_project_sets_user_agent(module_under_test, monkeypatch)
     mock_client = mock.create_autospec(bigquery.Client)
     monkeypatch.setattr(bigquery, "Client", value=mock_client)
     module_under_test.BigQueryResultHandler.get_handler_for_project(
-        "test-project", table_id="some_dataset.some_table"
+        "test-project", None, table_id="some_dataset.some_table"
```
nehanene15: nit: technically here you don't need to pass in 'None' since it's already the default. You should be able to leave this test as is with "test-project", table_id="some_dataset.some_table"

nehanene15 (approved):
Only have two nits, but approved otherwise!

PR #594, opened 2022-09-28
test: fix image in tests to use managed multi image

dhercher:
Using the public image supplied in Google's DevRel repo

https://github.com/googleapis/testing-infra-docker/blob/main/python/cloudbuild.yaml


ngdav (approved):
LGTM

PR #599, opened 2022-10-06
fix: Row validation optimization to avoid select all columns

### LINKED ISSUE #466: Optimize row validation to avoid SELECT * in each level
Currently a row validation generates a query like so:
```
SELECT *, TO_HEX(SHA256(`concat__all`)) AS `hash__all`
FROM (
  SELECT *,
         ARRAY_TO_STRING([`upper__rstrip__ifnull__cast__id`, `upper__rstrip__ifnull__cast__ts`, `upper__rstrip__ifnull__cast__order_date`], '') AS `concat__all`
  FROM (
    SELECT *,
           upper(`rstrip__ifnull__cast__id`) AS `upper__rstrip__ifnull__cast__id`,
           upper(`rstrip__ifnull__cast__ts`) AS `upper__rstrip__ifnull__cast__ts`,
           upper(`rstrip__ifnull__cast__order_date`) AS `upper__rstrip__ifnull__cast__order_date`
    FROM (
      SELECT *, rtrim(`ifnull__cast__id`) AS `rstrip__ifnull__cast__id`,
             rtrim(`ifnull__cast__ts`) AS `rstrip__ifnull__cast__ts`,
             rtrim(`ifnull__cast__order_date`) AS `rstrip__ifnull__cast__order_date`
      FROM (
        SELECT *,
               IFNULL(`cast__id`, 'DEFAULT_REPLACEMENT_STRING') AS `ifnull__cast__id`,
               IFNULL(`cast__ts`, 'DEFAULT_REPLACEMENT_STRING') AS `ifnull__cast__ts`,
               IFNULL(`cast__order_date`, 'DEFAULT_REPLACEMENT_STRING') AS `ifnull__cast__order_date`
        FROM (
          SELECT *, CAST(`id` AS STRING) AS `cast__id`,
                 CAST(`ts` AS STRING) AS `cast__ts`,
                 CAST(`order_date` AS STRING) AS `cast__order_date`
          FROM `pso-kokoro-resources.hivetest.test_timestamp`
        ) t4
      ) t3
    ) t2
  ) t1
) t0
```

We can optimize the results returned by the query if we get rid of the SELECT * at each level.

sramsrinivasan:
Closes #466

### DIFF
--- a/data_validation/query_builder/query_builder.py
+++ b/data_validation/query_builder/query_builder.py
@@ -466,7 +466,7 @@ class QueryBuilder(object):
         # else:
         #     return [field.compile(table) for field in self.calculated_fields]
 
-    def compile(self, data_client, schema_name, table_name):
+    def compile(self, validation_type, data_client, schema_name, table_name):
         """Return an Ibis query object
 
         Args:
@@ -486,8 +486,16 @@ class QueryBuilder(object):
                 calc_table = calc_table.mutate(
                     self.compile_calculated_fields(calc_table, n)
                 )
-        if self.comparison_fields:
-            calc_table = calc_table.mutate(self.compile_comparison_fields(calc_table))
+
+        if validation_type == consts.ROW_VALIDATION:
+            calc_table = calc_table.projection(
+                self.compile_comparison_fields(calc_table)
+            )
+        else:
+            if self.comparison_fields:
+                calc_table = calc_table.mutate(
+                    self.compile_comparison_fields(calc_table)
+                )
         compiled_filters = self.compile_filter_fields(calc_table)
         filtered_table = (
             calc_table.filter(compiled_filters) if compiled_filters else calc_table
--- a/data_validation/validation_builder.py
+++ b/data_validation/validation_builder.py
@@ -383,7 +383,7 @@ class ValidationBuilder(object):
                     f"Input custom query type: {self.config_manager.custom_query_type}"
                 )
         else:
-            query = self.source_builder.compile(**source_config)
+            query = self.source_builder.compile(self.validation_type, **source_config)
         if self.verbose:
             logging.info(source_config)
             logging.info("-- ** Source Query ** --")
@@ -427,7 +427,7 @@ class ValidationBuilder(object):
                     f"Input custom query type: {self.config_manager.custom_query_type}"
                 )
         else:
-            query = self.target_builder.compile(**target_config)
+            query = self.target_builder.compile(self.validation_type, **target_config)
         if self.verbose:
             logging.info(target_config)
             logging.info("-- ** Target Query ** --")


data_validation/query_builder/query_builder.py
```diff
@@ -486,8 +486,16 @@ def compile(self, data_client, schema_name, table_name):
                 calc_table = calc_table.mutate(
                     self.compile_calculated_fields(calc_table, n)
                 )
-        if self.comparison_fields:
-            calc_table = calc_table.mutate(self.compile_comparison_fields(calc_table))
+
+        if validation_type == consts.ROW_VALIDATION:
+            calc_table = calc_table.projection(
```
nehanene15, line 491: Does there need to be an `if self.comparison_fields` conditional here? Or will row validations always have a comparison field populated here?

nehanene15: After testing, it seems it will always have a comparison_field

data_validation/query_builder/query_builder.py
```diff
@@ -486,8 +486,16 @@ def compile(self, data_client, schema_name, table_name):
                 calc_table = calc_table.mutate(
                     self.compile_calculated_fields(calc_table, n)
                 )
-        if self.comparison_fields:
-            calc_table = calc_table.mutate(self.compile_comparison_fields(calc_table))
+
+        if validation_type == consts.ROW_VALIDATION:
```
nehanene15, line 490: I don't think we need to check for validation_type since comparison fields are only populated for row validation. We should keep the existing conditional for 'if self.comparison_fields' and make the QueryBuilder unaware of validation_type.

nehanene15 (changes requested):

data_validation/query_builder/query_builder.py
nehanene15: Found that we need this if a user wants to use random rows or PKs in column validation

sramsrinivasan: here is one use case for which we need validation_type comparison. 
data-validation validate column -sc my_bq_conn -tc my_bq_conn -tbls bigquery-public-data.new_york_citibike.citibike_trips --grouped-columns usertype -sum num_bikes_available --primary-keys station_id -rr -rbs 10

below are the other use cases that gets corrected with this conditional check - 

FAILED tests/unit/test_data_validation.py::test_grouped_column_level_validation_perfect_match - ibis.common.exceptions.IbisTypeError: 'date_value' is not a field in ['id']
FAILED tests/unit/test_data_validation.py::test_calc_field_validation_calc_match - ibis.common.exceptions.IbisTypeError: 'date_value' is not a field in ['id']
FAILED tests/unit/test_data_validation.py::test_grouped_column_level_validation_non_matching - ibis.common.exceptions.IbisTypeError: 'date_value' is not a field in ['id']
FAILED tests/unit/test_data_validation.py::test_grouped_column_level_validation_smart_count - ibis.common.exceptions.IbisTypeError: 'date_value' is not a field in ['id']
FAILED tests/unit/test_data_validation.py::test_grouped_column_level_validation_multiple_aggregations - ibis.common.exceptions.IbisTypeError: 'date_value' is not a field in ['id']
FAILED tests/unit/test_data_validation.py::test_random_row_level_validation - ibis.common.exceptions.IbisTypeError: 'int_value' is not a field in ['id']

nehanene15 (approved):

PR #607, opened 2022-10-13
refactor: custom query code

### LINKED ISSUE #539: Refactor custom query validation to use query builder
Currently the custom query validations uses separate logic to build the column level query. This should be refactored to use the existing query_builder logic used for standard column validations.

### LINKED ISSUE #575: Deprecate the usage of --tables-list/-tbls argument from DVT command for custom-query validation

pshrivastava27:
### This PR includes following changes:

1. Refactoring custom query (row and column validation) code to use ibis framework (fix #539  )
Instead of constructing nested queries manually using string concatenation, we now use ibis framework similar to how we construct queries for regular validations.
2. Removing table list dependency for custom query row validation (fix #575 )
3. Resolving issues specific to TD custom query row validation (trailing whitespaces after hash value)
4. Making flags `source-query-file` and `target-query-file` mandatory for custom-query validation commands.
5. Making `primary-keys`flag mandatory for custom-query row validation.


data_validation/__main__.py
```diff
@@ -31,15 +31,6 @@
 # by default yaml dumps lists as pointers. This disables that feature
 Dumper.ignore_aliases = lambda *args: True
 
-# Log level mappings for the input argument of log level string
-LOG_LEVEL_MAP = {
```
sramsrinivasan: @pshrivastava27  - did you get a chance to check on these deletes? we have to revert and check the test cases before merging

data_validation/__main__.py
```diff
@@ -71,14 +62,6 @@ def get_aggregate_config(args, config_manager):
         "int64",
         "decimal",
         "timestamp",
-        "float64[non-nullable]",
```
sramsrinivasan: @pshrivastava27  - Same as above

data_validation/__main__.py
```diff
@@ -183,6 +166,35 @@ def build_config_from_args(args, config_manager):
...
+        if args.custom_query_type is not None:
+            config_manager.append_custom_query_type(args.custom_query_type)
+        else:
+            raise ValueError(
+                "Expected custom query type to be given, got empty string."
+            )
+
+        if args.source_query_file is not None:
+            query_file = cli_tools.get_arg_list(args.source_query_file)
+            source_query_str = config_manager.get_query_from_file(query_file[0])
```
sramsrinivasan: @pshrivastava27  - Can you perform a test case in which the syntax of the sql queries supplied in the file is incorrect? something that is as simple as select * form (typo) or a select with WHERE col_a === 'c'. Pls record the output in the excel. 

Reason I am requesting this is to check if there is anyway we can detect sql query syntax error or any other issues associated with the sql query itself much earlier rather than at the final execution stage

data_validation/__main__.py
```diff
@@ -279,8 +277,14 @@ def build_config_managers_from_args(args):
     )
 
     is_filesystem = source_client._source_type == "FileSystem"
+    # is_custom_query = validate_cmd == consts.CUSTOM_QUERY
+    try:
```
sramsrinivasan: @pshrivastava27  - Can this be handled via IF ELSE condition instead of try catch? Any particular reasoning behind using try catch?

data_validation/__main__.py
```diff
@@ -534,15 +538,13 @@ def validate(args):
 
 
 def main():
-
-    # Create Parser and Get Deployment Info
-    args = cli_tools.get_parsed_args()
     logging.basicConfig(
-        level=LOG_LEVEL_MAP[args.log_level],
```
sramsrinivasan, line 522: @pshrivastava27  - What is the reason for this change? Should we keep log level to INFO?

Raniksingh:
Can you also update the README with the following: 
  1. Remove tbls from custom query syntax(as this PR will remove tbls dependency from custom query
  2. add PK in the syntax for custom query as its a mandatory argument now
  3. add the syntax for concat option for both regular and custom query row validation syntax.

data_validation/cli_tools.py
```diff
@@ -498,11 +498,13 @@ def _configure_custom_query_parser(custom_query_parser):
     custom_query_parser.add_argument(
```
nehanene15, line 499: For the --custom-query-type argument on line 492, we should check that the user can only provide 'row' or 'column' using argparse [choices](https://docs.python.org/3/library/argparse.html#choices).

data_validation/__main__.py
```diff
@@ -183,6 +183,35 @@ def build_config_from_args(args, config_manager):
...
+    if config_manager.validation_type == consts.CUSTOM_QUERY:
+        if args.custom_query_type is not None:
+            config_manager.append_custom_query_type(args.custom_query_type)
+        else:
+            raise ValueError(
+                "Expected custom query type to be given, got empty string."
+            )
+
+        if args.source_query_file is not None:
+            query_file = cli_tools.get_arg_list(args.source_query_file)
```
nehanene15: Why are we using 'get_arg_list' here if we don't expect a list of query files? I feel like we could just remove this line and refer to 'args.source_query_file' in line 196.

data_validation/__main__.py
```diff
@@ -183,6 +183,35 @@ def build_config_from_args(args, config_manager):
...
+            source_query_str = config_manager.get_query_from_file(query_file[0])
+            config_manager.append_source_query(source_query_str)
+
+        if args.target_query_file is not None:
+            query_file = cli_tools.get_arg_list(args.target_query_file)
```
nehanene15: Same as above - not sure why we need to use 'get_arg_list'

data_validation/__main__.py
```diff
@@ -183,6 +183,35 @@ def build_config_from_args(args, config_manager):
     Args:
         config_manager (ConfigManager): Validation config manager instance.
     """
+    if config_manager.validation_type == consts.CUSTOM_QUERY:
```
nehanene15, line 173: We should move this whole conditional block to line 235 and make it an elif after checking for column and row validation:
`elif config_manager.validation_type == consts.CUSTOM_QUERY:`

data_validation/__main__.py
```diff
@@ -209,24 +238,10 @@ def build_config_from_args(args, config_manager):
         config_manager.append_primary_keys(
             config_manager.build_column_configs(primary_keys)
         )
+        calc_type = args.hash or args.concat
```
nehanene15: Why is this here? Shouldn't we move this into the 'elif config_manager.validation_type == consts.CUSTOM_QUERY_VALIDATION' block similar to line 225?

data_validation/__main__.py
```diff
@@ -279,8 +294,14 @@ def build_config_managers_from_args(args):
     )
 
     is_filesystem = source_client._source_type == "FileSystem"
+    # is_custom_query = validate_cmd == consts.CUSTOM_QUERY
+    try:
```
nehanene15: This try/except is confusing.. we set tables_list = args.tables_list and then override it on line 303. Not sure what the goal is here. The cli_tools.get_tables_list() already returns a default value of [{}] if args.tables_list doesn't exist.

data_validation/cli_tools.py
nehanene15, line 499: On line 491, shouldn't this be `_add_common_arguments(custom_query_parser, query_type="Custom-query")` ? Or when do we make use of the 'query_type' parameter?

data_validation/config_manager.py
```diff
@@ -305,10 +319,20 @@ def get_source_ibis_table(self):
...
+        self._source_ibis_table = clients.get_ibis_query(
+            self.source_client, self.source_query
+        )
+        return self._source_ibis_table
+
     def get_source_ibis_calculated_table(self, depth=None):
         """Return mutated IbisTable from source
         n: Int the depth of subquery requested"""
-        table = self.get_source_ibis_table()
+        if self.source_query:
```
nehanene15: Would it be more readable to check if self.validation_type = custom_query instead of checking if a source_query is populated?

data_validation/config_manager.py
```diff
@@ -324,10 +348,20 @@ def get_target_ibis_table(self):
...
+        self._target_ibis_table = clients.get_ibis_query(
+            self.target_client, self.target_query
+        )
+        return self._target_ibis_table
+
     def get_target_ibis_calculated_table(self, depth=None):
         """Return mutated IbisTable from target
         n: Int the depth of subquery requested"""
-        table = self.get_target_ibis_table()
+        if self.target_query:
```
nehanene15: Same as above..

data_validation/config_manager.py
```diff
@@ -713,3 +747,20 @@ def _build_dependent_aliases(self, calc_type, col_list=None):
                     column_aliases[name] = i
                     col_names.append(col)
         return col_names
+
+    def get_query_from_file(self, filename):
+        """Return query from input file"""
+        query = ""
+        try:
+            file = open(filename, "r")
+            query = file.read()
```
nehanene15, line 745: We added a line here in PR #591 :
`query = query.rstrip(";\n")`

data_validation/validation_builder.py
```diff
@@ -397,37 +380,21 @@ def get_target_query(self):
...
+            table = clients.get_ibis_table(
+                self.target_client,
+                self.config_manager.target_schema,
+                self.config_manager.target_table,
+            )
+
+        query = self.target_builder.compile(self.validation_type, table)
         if self.verbose:
             logging.info(target_config)
             logging.info("-- ** Target Query ** --")
```
nehanene15, line 388: We should remove the 'get_query_from_file` method on line 414 since you moved it to config_manager

data_validation/query_builder/query_builder.py
```diff
@@ -466,15 +466,13 @@ def compile_calculated_fields(self, table, n=0):
...
         """Return an Ibis query object
 
         Args:
-            data_client (IbisClient): The client used to validate the query.
-            schema_name (String): The name of the schema for the given table.
-            table_name (String): The name of the table to query.
+            table (IbisTable): The Ibis Table expression.
         """
-        calc_table = clients.get_ibis_table(data_client, schema_name, table_name)
+        calc_table = table
```
nehanene15: Instead of renaming this, let's replace the instances of 'calc_table' with 'table'.

data_validation/config_manager.py
```diff
@@ -281,6 +281,20 @@ def threshold(self):
         """Return threshold from Config"""
         return self._config.get(consts.CONFIG_THRESHOLD, 0.0)
 
+    @property
+    def source_query(self):
```
nehanene15, line 274: This should have a default value on None:
`return self._config.get("source_query", None)`

We may also want to have a consts here for SOURCE_QUERY and TARGET_QUERY so we can easily change the name in the future.

data_validation/config_manager.py
```diff
@@ -281,6 +281,20 @@ def threshold(self):
...
+        return self._config.get("source_query")
+
+    def append_source_query(self, source_query):
+        self._config["source_query"] = source_query
+
+    @property
+    def target_query(self):
+        return self._config.get("target_query")
```
nehanene15: See above for adding a default value.

data_validation/query_builder/query_builder.py
```diff
@@ -14,7 +14,7 @@
```
nehanene15, line 14: Can we delete the query_builder/custom_query_builder.py file with this refactor?

nehanene15 (changes requested):
Have we tested with saving the custom query validation to YAML and then running it?
Does this support comparison field validations or only hash/concat?

data_validation/__main__.py
pshrivastava27: This has been reverted back ([here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/510d651c5070785dcf71140f29e016aade407a46)) to how it currently exists in the develop branch, and tested too.
We'll have a dict of log levels, and the current log level will be decided by what is passed in the args.

data_validation/__main__.py
pshrivastava27: reverted back to the current state in develop branch

data_validation/__main__.py
pshrivastava27: In case of a typo or similar syntax error, the query will fail during execution. This is documented [here](https://docs.google.com/spreadsheets/d/1lwtblCauq83fnGb2JgiU-rJAD6una2BI5BWCwQ9bdqE/edit#gid=0&range=15:15)

We can possibly do a syntax check using sqlparse or similar libraries but I'm not sure if we should do it now or at a later stage. We can discuss this in our connect.

data_validation/query_builder/query_builder.py
pshrivastava27: Sure. We'll also need to remove the testcase associated with it (tests/unit/query_builder/test_custom_query_builder.py), otherwise the nox test will fail.

We'll anyway add a new test case for custom query later, so I think it should be okay. I'll remove both the custom_query_builder.py as well as the test.
Does that sound good?

data_validation/config_manager.py
pshrivastava27: Fixed [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/730c8ca04214ce3790d68a6564caa69af86a876d)

data_validation/validation_builder.py
pshrivastava27: fixed in this [commit](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/730c8ca04214ce3790d68a6564caa69af86a876d)
Had to remove testing of get_query_from_file() from  test-case of validation_builder.py and add it to test-case of config_manager.py

data_validation/config_manager.py
pshrivastava27: Fixed in this [commit](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/78cbe136cd70a5208799a61a6c5602baff080e74)

data_validation/config_manager.py
pshrivastava27: Fixed in this [commit](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/78cbe136cd70a5208799a61a6c5602baff080e74)

data_validation/config_manager.py
pshrivastava27: Fixed [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/78cbe136cd70a5208799a61a6c5602baff080e74)

data_validation/config_manager.py
pshrivastava27: Fixed [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/78cbe136cd70a5208799a61a6c5602baff080e74)

data_validation/__main__.py
pshrivastava27: This has been reverted back ([here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/510d651c5070785dcf71140f29e016aade407a46)) to how it currently exists in the develop branch, and tested too.
We'll have a dict of log levels, and the current log level will be decided by what is passed in the args.

data_validation/__main__.py
pshrivastava27: I guess try catch was introduced only while local debugging. This has been fixed through this [commit](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/0e6ac6965696b40cbcf577c17983bee8ae24c92a)

data_validation/cli_tools.py
pshrivastava27: Thanks for the suggestion. Fixed [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/0e6ac6965696b40cbcf577c17983bee8ae24c92a)

data_validation/__main__.py
pshrivastava27: Made the necessary changes in this [commit](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/0e6ac6965696b40cbcf577c17983bee8ae24c92a)

data_validation/__main__.py
pshrivastava27: Made the necessary changes in this [commit](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/0e6ac6965696b40cbcf577c17983bee8ae24c92a)

data_validation/__main__.py
pshrivastava27: Correct. This is now fixed through this [commit](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/0e6ac6965696b40cbcf577c17983bee8ae24c92a)

data_validation/__main__.py
pshrivastava27: Done, made the changes [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/0e6ac6965696b40cbcf577c17983bee8ae24c92a)

data_validation/__main__.py
pshrivastava27: I guess try catch was introduced only while local debugging. This has been fixed through this [commit](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/0e6ac6965696b40cbcf577c17983bee8ae24c92a)

data_validation/cli_tools.py
pshrivastava27: missed this part. Thanks for the input. This has been fixed through this [commit](I guess try catch was introduced only while local debugging. This has been fixed through this [commit](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/0e6ac6965696b40cbcf577c17983bee8ae24c92a))

data_validation/query_builder/query_builder.py
pshrivastava27: Fixed through this [commit](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/0e6ac6965696b40cbcf577c17983bee8ae24c92a)

data_validation/query_builder/query_builder.py
pshrivastava27: Deleted the py file and its testcase in this [commit](https://github.com/GoogleCloudPlatform/professional-services-data-validator/commit/f5b17d9760f8748c42852ed7ce30767344bbd020)

pshrivastava27:
> Have we tested with saving the custom query validation to YAML and then running it? Does this support comparison field validations or only hash/concat?

Yes. We tested generating the yaml config and then running the validation using config file. Works as expected.

pshrivastava27:
> Can you also update the README with the following:
> 
> 1. Remove tbls from custom query syntax(as this PR will remove tbls dependency from custom query
> 2. add PK in the syntax for custom query as its a mandatory argument now
> 3. add the syntax for concat option for both regular and custom query row validation syntax.

Done [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/42097549f3aff7c9dd820306cf7643eb159831f5)

data_validation/__main__.py
```diff
@@ -279,9 +294,14 @@ def build_config_managers_from_args(args):
     )
 
     is_filesystem = source_client._source_type == "FileSystem"
-    tables_list = cli_tools.get_tables_list(
-        args.tables_list, default_value=[{}], is_filesystem=is_filesystem
-    )
+    if config_type == consts.CUSTOM_QUERY:
+        tables_list = cli_tools.get_tables_list(
```
sramsrinivasan: On line 298, can we avoid calling the fun cli_tools.get_tables_list()? Since we are sure that tables_list is None for CUSTOM_QUERY

data_validation/__main__.py
```diff
@@ -183,7 +183,6 @@ def build_config_from_args(args, config_manager):
     Args:
         config_manager (ConfigManager): Validation config manager instance.
     """
-    config_manager.append_calculated_fields(get_calculated_config(args, config_manager))
```
sramsrinivasan, line 172: Why are we deleting this line of code?

data_validation/validation_builder.py
```diff
@@ -397,37 +380,21 @@ def get_target_query(self):
             "data_client": self.target_client,
             "schema_name": self.config_manager.target_schema,
             "table_name": self.config_manager.target_table,
+            "target_query": self.config_manager.target_query,
         }
+
         if self.validation_type == consts.CUSTOM_QUERY:
-            target_input_query = self.get_query_from_file(
-                self.config_manager.target_query_file[0]
+            table = clients.get_ibis_query(
```
sramsrinivasan: line 361-364 in config_manager.py fetches the table object. Do we need this again here?
@nehanene15  - Can you vet this?

data_validation/validation_builder.py
```diff
@@ -397,37 +380,21 @@ def get_target_query(self):
...
-                )
-                query = self.target_client.sql(target_aggregate_query)
-            else:
-                raise ValueError(
-                    "Expected custom query type to be column or row, got an unacceptable value. "
-                    f"Input custom query type: {self.config_manager.custom_query_type}"
-                )
         else:
-            query = self.target_builder.compile(self.validation_type, **target_config)
+            table = clients.get_ibis_table(
```
sramsrinivasan: line 361-364 in config_manager.py fetches the table object. Do we need this again here?
@nehanene15  - Can you vet this?

sramsrinivasan (changes requested):

data_validation/__main__.py
pshrivastava27: @sramsrinivasan  we are not deleting this line of code, this is just moved down a little bit.
We need to have a few fields (custom-query specific) already populated in the config_manager object, before the execution flow reaches the function `get_calculated_config(args, config_manager)`

data_validation/validation_builder.py
pshrivastava27: This is correct for normal as well as custom-query validations.
We already have the source (and target) ibis table expression generated in `config_manager.py`, and then we calculate it again in `validation_builder.py`.
This is because we are not adding source (and target) table expressions in the ConfigManager object (function `build_config_manager()` in `config_manager.py`)

@nehanene15 Let us know if this is correct understanding, and if we want to tackle this now or maybe at a later stage

data_validation/validation_builder.py
pshrivastava27: Same as above

data_validation/__main__.py
pshrivastava27: agreed, and thanks for the suggestion!
This has been updated [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/4f2b8efafb92762eb1a33ad5a32991f820c3ed59)

data_validation/validation_builder.py
nehanene15: We should keep this as clients.get_ibis_query since config manager methods are source/target specific i.e. get_source_ibis_table()

In the current query_builder we also use clients.get_ibis_query so we can keep this logic for now.

data_validation/__main__.py
nehanene15: Currently this line of code is before the if/elif block - will this affect column/row validations if we move it down?

nehanene15, line 172: I think moving this line down may affect regular row validation... since on line 296 we check 'if args.comparison_fields is not None' and get_calculated_config() populates comparison fields.

data_validation/__main__.py
```diff
@@ -203,29 +202,45 @@ def build_config_from_args(args, config_manager):
...
             config_manager.append_custom_query_type(args.custom_query_type)
         else:
             raise ValueError(
                 "Expected custom query type to be given, got empty string."
             )
+
         if args.source_query_file is not None:
-            query_file = cli_tools.get_arg_list(args.source_query_file)
-            config_manager.append_source_query_file(query_file)
+            query_file = args.source_query_file
```
nehanene15: nit: we don't need to assign a variable to this since we aren't updating it. We should instead change line 216 to be `config_manager.get_query_from_file(args.source_query_file)`

Same with line 220 below.

data_validation/__main__.py
```diff
@@ -279,8 +294,12 @@ def build_config_managers_from_args(args):
...
 
     is_filesystem = source_client._source_type == "FileSystem"
-    tables_list = cli_tools.get_tables_list(
-        args.tables_list, default_value=[{}], is_filesystem=is_filesystem
+    tables_list = (
+        cli_tools.get_tables_list(
+            args.tables_list, default_value=[{}], is_filesystem=is_filesystem
+        )
+        if config_type != consts.CUSTOM_QUERY
+        else [{}]
```
nehanene15: See my comment below on cli_tools.py - if we default args.tables_list to None instead of removing the argument itself from the parser, args.tables_list will be set to None for custom query. 
Then, we can remove this if/else and it will be set to the default_value of [{}]

data_validation/cli_tools.py
```diff
@@ -581,18 +589,22 @@ def _configure_custom_query_parser(custom_query_parser):
...
         "--target-conn", "-tc", required=True, help="Target connection name"
     )
-    parser.add_argument(
-        "--tables-list",
-        "-tbls",
-        help="Comma separated tables list in the form 'schema.table=target_schema.target_table'",
-    )
+
+    ##  remove tables-list from custom query validation
+    if query_type != consts.CUSTOM_QUERY:
```
nehanene15: Instead of removing the argument, why don't we just default it to None?
```
parser.add_argument(
            "--tables-list",
            "-tbls",
            default=None,
            help="Comma separated tables list in the form 'schema.table=target_schema.target_table'",
        )
```

data_validation/validation_builder.py
nehanene15: Actually, after looking at this further we should use the config_manager for this - we should do table = self.config_manager. get_target_ibis_table_from_query()

data_validation/validation_builder.py
nehanene15: Let's use the config_manager for this too - `table = self.config_manager.get_target_ibis_table()`

data_validation/validation_builder.py
```diff
@@ -353,37 +352,21 @@ def get_source_query(self):
             "data_client": self.source_client,
             "schema_name": self.config_manager.source_schema,
             "table_name": self.config_manager.source_table,
+            "source_query": self.config_manager.source_query,
         }
+
         if self.validation_type == consts.CUSTOM_QUERY:
-            source_input_query = self.get_query_from_file(
-                self.config_manager.source_query_file[0]
+            table = clients.get_ibis_query(
```
nehanene15: Same with below - let's use config_manager for these

tests/unit/test_validation_builder.py
```diff
@@ -242,15 +242,3 @@ def test_custom_query_validation(module_under_test):
     assert (
```
nehanene15: Can we delete this test too on line 232? This shouldn't be in the test_validation_builder() and it's not asserting anything useful.

data_validation/validation_builder.py
pshrivastava27: Agreed. This is incorporated [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/f0eb38bd9e472989bcb8a8ed43a69350c1d28973)

data_validation/validation_builder.py
pshrivastava27: Agreed. This is incorporated [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/f0eb38bd9e472989bcb8a8ed43a69350c1d28973)

data_validation/__main__.py
pshrivastava27: Done [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/f0eb38bd9e472989bcb8a8ed43a69350c1d28973)

data_validation/validation_builder.py
pshrivastava27: Done [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/f0eb38bd9e472989bcb8a8ed43a69350c1d28973)

tests/unit/test_validation_builder.py
pshrivastava27: Deleted the testcase through this [commit](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/660486c00925197c5fff7ee387be7376821f488d)

data_validation/__main__.py
pshrivastava27: Had to adjust the conditions a bit, the code for normal validation remains as it was before.
Fixed [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/f0eb38bd9e472989bcb8a8ed43a69350c1d28973#diff-ec3232fc84a73e24bf5cf68f7beb86e56e968c565509c7e07b420d026fd4f09f)

pshrivastava27: Reverted it back to the previous state, however for custom query validation, we need to have source query and target query string appended in the config_manager object, before calling `get_calculated_config`.
Fixed [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/f0eb38bd9e472989bcb8a8ed43a69350c1d28973)

data_validation/__main__.py
pshrivastava27: Fixed through this [commit](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/0ab8891e84ce18fa43e38d5b69a413e5d10b520d)

data_validation/cli_tools.py
pshrivastava27: Fixed through this [commit](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/0ab8891e84ce18fa43e38d5b69a413e5d10b520d)

data_validation/__main__.py
```diff
@@ -169,6 +169,38 @@ def build_config_from_args(args, config_manager):
...
+            )
+            config_manager.append_source_query(source_query_str)
+
+        if args.target_query_file is not None:
+            target_query_str = config_manager.get_query_from_file(
+                args.target_query_file
+            )
+            config_manager.append_target_query(target_query_str)
+
+        if args.custom_query_type.lower() == consts.COLUMN_VALIDATION.lower():
```
nehanene15: The user can only input 'row' or 'column' due to the argparse choices. So we don't need to lower() before checking.

data_validation/__main__.py
```diff
@@ -169,6 +169,38 @@ def build_config_from_args(args, config_manager):
...
+            config_manager.append_aggregates(get_aggregate_config(args, config_manager))
+
+        if (
```
nehanene15: Let's make this an 'elif' as part of the conditional above.. 
```
if args.custom_query_type == consts.COLUMN_VALIDATION.lower():
   code
elif not args.primary_keys:
   raise ValueError(...)
```

data_validation/__main__.py
```diff
@@ -169,6 +169,38 @@ def build_config_from_args(args, config_manager):
     Args:
         config_manager (ConfigManager): Validation config manager instance.
     """
+
+    if config_manager.validation_type == consts.CUSTOM_QUERY:
+        if args.custom_query_type is not None:
+            config_manager.append_custom_query_type(args.custom_query_type)
+        else:
```
nehanene15: We can remove the checks for all of these being 'None' and throwing an error since custom_query_type, source_query_file, and target_query_file are all required arguments.

data_validation/__main__.py
```diff
@@ -224,6 +241,10 @@ def build_config_managers_from_args(args):
         config_type = consts.ROW_VALIDATION
     elif validate_cmd == "Custom-query":
         config_type = consts.CUSTOM_QUERY
+        if args.tables_list is not None:
```
nehanene15: I would move this down to line 284 to keep the tables_list logic near each other. 
```
if args.tables_list and config_type == consts.CUSTOM_QUERY: 
   logging.warning(
                "Tables-list/tbls flag is not supported for custom-query validations, supplied tables list will be ignored."
            )
```

nehanene15 (commented):
Logic looks good, just suggested some code style changes.

data_validation/__main__.py
pshrivastava27: Fixed this [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/d39c63d2ab626b5368dedafa5169876d37c4baf0)

data_validation/__main__.py
pshrivastava27: Fixed this [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/d39c63d2ab626b5368dedafa5169876d37c4baf0)

data_validation/__main__.py
pshrivastava27: Fixed this [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/d39c63d2ab626b5368dedafa5169876d37c4baf0)

data_validation/__main__.py
pshrivastava27: Fixed this [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/d39c63d2ab626b5368dedafa5169876d37c4baf0)

data_validation/cli_tools.py
```diff
@@ -581,18 +589,21 @@ def _configure_custom_query_parser(custom_query_parser):
     )
 
 
-def _add_common_arguments(parser):
+def _add_common_arguments(parser, query_type="ANY"):
```
nehanene15: We don't need the query_type parameter any more.

data_validation/cli_tools.py
```diff
@@ -488,21 +488,24 @@ def _configure_schema_parser(schema_parser):
 
 def _configure_custom_query_parser(custom_query_parser):
     """Configure arguments to run custom-query validations."""
-    _add_common_arguments(custom_query_parser)
+    _add_common_arguments(custom_query_parser, consts.CUSTOM_QUERY)
```
nehanene15: See below; don't need to pass in the query_type.

nehanene15 (approved):
Approved. Suggested one last comment to delete the query_type in the arg parser since it is not used.

data_validation/cli_tools.py
pshrivastava27: Fixed [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/7667b5f603a4796bb96360ec48b7c66a9611179e)

data_validation/cli_tools.py
pshrivastava27: Fixed [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/607/commits/7667b5f603a4796bb96360ec48b7c66a9611179e)

sramsrinivasan (approved):

Issue #596, opened 2022-10-04
bare data-validation command throws exception


vicenteg:
I would have expected the following would provide output similar to `--help`:

```
$ data-validation
Traceback (most recent call last):
  File "/home/vincegonzalez/dv/venv/bin/data-validation", line 8, in <module>
    sys.exit(main())
  File "/home/vincegonzalez/dv/venv/lib/python3.8/site-packages/data_validation/__main__.py", line 537, in main
    raise ValueError(f"Positional Argument '{args.command}' is not supported")
ValueError: Positional Argument 'None' is not supported
```

```
$ pip freeze | grep data-val
google-pso-data-validator==2.3.0
```

PR #627, opened 2022-11-02
fix: bare data-validation command throws exception

piyushsarraf:
Closes issue #596 
* Prints data-validation --help output when bare data-validation command is executed.

### DIFF
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -141,7 +141,8 @@ CONNECTION_SOURCE_FIELDS = {
 def get_parsed_args():
     """Return ArgParser with configured CLI arguments."""
     parser = configure_arg_parser()
-    return parser.parse_args()
+    args = None if sys.argv[1:] else ["--help"]
+    return parser.parse_args(args)
 
 
 def configure_arg_parser():
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -141,7 +141,7 @@ CONNECTION_SOURCE_FIELDS = {
 def get_parsed_args():
     """Return ArgParser with configured CLI arguments."""
     parser = configure_arg_parser()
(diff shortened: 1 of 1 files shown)


data_validation/cli_tools.py
```diff
@@ -141,7 +141,8 @@
 def get_parsed_args():
     """Return ArgParser with configured CLI arguments."""
     parser = configure_arg_parser()
-    return parser.parse_args()
+    args = None if sys.argv[1:] else ["--help"]
```
nehanene15: This breaks the code after testing with `data-validation`. I'm not understanding the logic here...

We want to do something similar to [this](https://stackoverflow.com/questions/4042452/display-help-message-with-python-argparse-when-script-is-called-without-any-argu) where if no arguments are provided, we print the help menu.

nehanene15 (changes requested):

data_validation/cli_tools.py
piyushsarraf: The logic is when you go to `parse_args()` function it has a args parameter which is default to None, so earlier we were calling parse_args() with empty parameter, in which args parameter will be default to None. Now `sys.argv[1:]` returns a list of arguments passed with data-validation command, so if we pass any argument it will work with our original logic, but if we don't pass any argument, `sys.argv[1:]` will return null and we will pass the --help flag in args parameter.

nehanene15: Makes sense. Only suggestion is to check the length instead of list slicing for efficiency. ie `if len(sys.argv) == 1`

piyushsarraf: Sure, fixed it with this [commit](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/627/commits/dd615958b1e10101240bda629cf538ffca5ef9f9).

nehanene15 (approved):

PR #643, opened 2022-12-07
feat: Logic to add allow-list to support datatype matching with a provided list in case of mismatched datatypes between source and target

### LINKED ISSUE #606: Enable an 'allowlist' for data type mappings for schema validations
There are use cases when users expect a certain data type to be mapped to a different target data type in migrations. For example, when using the Data Transfer Service, users will expect a Redshift non-nullable string to be converted to a BigQuery nullable string data type. 

Another example is if users changed the precision/scale from Redshift DECIMAL(12,2) to BigQuery BIGNUMERIC/DECIMAL(38,9).

DVT schema validations should let users provide an allowlist of data type mappings that should be accepted as a successful validation.

ex.
```
data-validation validate schema -sc redshift -tc bq -tbls schema.table=schema.table --allow-list decimal(12,2):decimal(38,9),string[non-nullable]:string
```

The allowlist should accept a comma-separated list of type mappings or a file with the mappings.
```
[--allow-list MAPPING_FILE | MAPPING] Comma-separated list of data type mappings to be considered successful matches between source and target.
```

kanhaPrayas:
This PR is under progress. This has been raised for parallel review. So that useful review comments can be added as part of development.

Closes #606


data_validation/consts.py
```diff
@@ -119,10 +119,12 @@
 VALIDATION_STATUS = "validation_status"
 VALIDATION_STATUS_SUCCESS = "success"
 VALIDATION_STATUS_FAIL = "fail"
+VALIDATION_STATUS_WARNING = "warning"
 # TODO: update if we start to support other statuses
 VALIDATION_STATUSES = [
     VALIDATION_STATUS_SUCCESS,
     VALIDATION_STATUS_FAIL,
+    VALIDATION_STATUS_WARNING,
```
nehanene15: We shouldn't add another status as 'warning', but rather throw a warning with logging.warning if a data type isn't an exact match but still fits within the target type (i.e. decimal(1,1) --> decimal(38,9) OR int32 --> int64)
In this case, it should still be marked as a validation success.

data_validation/schema_validation.py
```diff
@@ -136,6 +139,40 @@ def schema_validation_matching(source_fields, target_fields, exclusion_fields):
...
+                            [
+                                source_field_name,
+                                source_field_name,
+                                str(source_field_type),
+                                str(target_field_type),
+                                consts.VALIDATION_STATUS_SUCCESS,
+                            ]
+                        ) 
+                    else:
+                        results.append(
```
nehanene15: See above comment - we can just log a warning and mark as success.

data_validation/cli_tools.py
```diff
@@ -485,6 +485,11 @@ def _configure_schema_parser(schema_parser):
         "-ec",
         help="Comma separated list of columns 'col_a,col_b' to be excluded from the schema validation",
     )
+    schema_parser.add_argument(
+        "--allow-list",
+        "-al",
+        help="Comma separated list of datatype mappings due to incompatible datatypes in source and columns. e.g: decimal(12,2):decimal(38,9),string[non-nullable]:string",
```
nehanene15: Should this be "incompatible datatypes in source and target"?

data_validation/schema_validation.py
```diff
@@ -172,3 +200,96 @@ def schema_validation_matching(source_fields, target_fields, exclusion_fields):
                 ]
             )
     return results
+
+def is_number(val):
```
nehanene15: Python has a _string_.isdigit() method that also returns a boolean.

nehanene15 (commented):
LGTM besides two small comments.

data_validation/cli_tools.py
kanhaPrayas: done

data_validation/schema_validation.py
kanhaPrayas: done

data_validation/cli_tools.py
```diff
@@ -791,7 +796,9 @@ def get_arg_list(arg_value, default_value=None):
             arg_value = str(arg_value)
         # arg_value = "hash_all"
         arg_list = json.loads(arg_value)
+        print(799)
```
nehanene15: Are these print statements supposed to be here?

README.md
```diff
@@ -226,7 +226,10 @@ data-validation (--verbose or -v) (--log-level or -ll) validate schema
   [--filter-status or -fs STATUSES_LIST]
                         Comma separated list of statuses to filter the validation results. Supported statuses are (success, fail). If no list is provided, all statuses are returned.
   [--exclusion-columns or -ec EXCLUSION_COLUMNS]
-                        Comma separated list of columns to be excluded from the schema validation, i.e col_a,col_b.                    
+                        Comma separated list of columns to be excluded from the schema validation, i.e col_a,col_b.
+
+  [--allow-list or -al ALLOW_LIST]
+                        Comma separated list of data-type mappings of source and destination data sources which will be validated in case of missing data types in destination data source.
```
nehanene15: We should provide an example here like you did in cli_tools ("e.g: decimal(12,2):decimal(38,9),string[non-nullable]:string")

data_validation/schema_validation.py
```diff
@@ -172,3 +216,98 @@ def schema_validation_matching(source_fields, target_fields, exclusion_fields):
...
+
+def validate_typeb_vals(source, target):
+    if source[0] > target[0] or source[1] > target[1]:
+        return False, True
+    elif source[0] == target[0] and source[1] == target[1]:
+        return False, False
+    return True, False
+
+
+"""
```
nehanene15: Which function does this comment refer to?

nehanene15 (commented):
LGTM besides some minor doc changes!

README.md
kanhaPrayas: done

data_validation/cli_tools.py
kanhaPrayas: removed the print(). Those were for debugging.

data_validation/schema_validation.py
kanhaPrayas: Fixed the comment.

nehanene15 (approved):

Issue #502, opened 2022-06-07
Custom Query row level validation support for Oracle


nehanene15:
DVT should support custom query row level validation for Oracle databases.

PR #646, opened 2022-12-16
feat: Added custom query support for Oracle

piyushsarraf:
- feat: Added custom query support for Oracle(#502)

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -45,6 +45,21 @@ data-validation validate column -sc my_bq_conn -tc my_bq_conn -tbls bigquery-pub
 ````shell script
 data-validation validate row -sc my_bq_conn -tc my_bq_conn -tbls bigquery-public-data.new_york_citibike.citibike_stations --primary-keys station_id --hash '*'
 ````
+#### Run a row hash validation for all rows when database is Oracle and primary key is of Numeric Datatype
+Typecast Numeric to String by either of following: 
+* Adding the code snippet above [this](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/63d6303ae08e4f8eb9a663a14d6d60828947f027/data_validation/config_manager.py#L187):
+    ````shell script
+primary_key_configs[0]['cast'] = 'string'
+````
+* When validating with yaml file update ````cast: null```` to ````cast: string````
+    ````shell script
+primary_keys:
+  - cast: null
+    field_alias: dept_no
+    source_column: dept_no
+    target_column: dept_no
+````
+
 
 #### Run a row hash validation for all rows but filter only the failed records 
 ````shell script
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -45,20 +45,20 @@ data-validation validate column -sc my_bq_conn -tc my_bq_conn -tbls bigquery-pub
 ````shell script
 data-validation validate row -sc my_bq_conn -tc my_bq_conn -tbls bigquery-public-data.new_york_citibike.citibike_stations --primary-keys station_id --hash '*'
 ````
-#### Run a row hash validation for all rows when database is Oracle and primary key is of Numeric Datatype
-Typecast Numeric to String by either of following: 
+#### Run a row hash validation with a **NUMERIC** primary key
+For Oracle and Postgres connections, numeric primary keys may not match between source and target data types due to precision or scale discrepancies. You may need to cast the primary key to a string for a valid match which can be done by either of following: 
 * Adding the code snippet above [this](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/63d6303ae08e4f8eb9a663a14d6d60828947f027/data_validation/config_manager.py#L187):
     ````shell script
-primary_key_configs[0]['cast'] = 'string'
-````
+    primary_key_configs[0]['cast'] = 'string'
+    ````
 * When validating with yaml file update ````cast: null```` to ````cast: string````
     ````shell script
-primary_keys:
-  - cast: null
-    field_alias: dept_no
-    source_column: dept_no
-    target_column: dept_no
-````
+    primary_keys:
+      - cast: null
+        field_alias: dept_no
+        source_column: dept_no
+        target_column: dept_no
+    ````
 
 
 #### Run a row hash validation for all rows but filter only the failed records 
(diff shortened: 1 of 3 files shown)


third_party/ibis/ibis_oracle/client.py
```diff
@@ -242,3 +244,36 @@ def list_tables(self, like=None, database=None, schema=None):
...
+        -------
+        table : TableExpr
+        """
+        limited_query = "SELECT * FROM ({}) t0 WHERE ROWNUM <= 1".format(query)
+        schema = self._get_schema_using_query(limited_query)
+        return ops.SQLQueryResult(query, schema, self).to_expr()
+
+    def _get_schema_using_query(self, limited_query):
+        type_map = {
+            "DB_TYPE_NUMBER": "int64",
```
nehanene15: Should this be 'float64' since NUMBER can include precision/scale?

docs/examples.md
```diff
@@ -45,6 +45,21 @@ data-validation validate column -sc my_bq_conn -tc my_bq_conn -tbls bigquery-pub
...
+#### Run a row hash validation for all rows when database is Oracle and primary key is of Numeric Datatype
+Typecast Numeric to String by either of following: 
+* Adding the code snippet above [this](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/63d6303ae08e4f8eb9a663a14d6d60828947f027/data_validation/config_manager.py#L187):
+    ````shell script
+primary_key_configs[0]['cast'] = 'string'
+````
+* When validating with yaml file update ````cast: null```` to ````cast: string````
+    ````shell script
+primary_keys:
+  - cast: null
```
nehanene15: Can't we just update this to `cast: string` instead of also updating the code snippet above?

third_party/ibis/ibis_oracle/client.py
piyushsarraf: The custom query validation fails if we use float64, because of same precision issue that we were facing in normal row validation.

docs/examples.md
```diff
@@ -45,6 +45,21 @@ data-validation validate column -sc my_bq_conn -tc my_bq_conn -tbls bigquery-pub
 ````shell script
 data-validation validate row -sc my_bq_conn -tc my_bq_conn -tbls bigquery-public-data.new_york_citibike.citibike_stations --primary-keys station_id --hash '*'
 ````
+#### Run a row hash validation for all rows when database is Oracle and primary key is of Numeric Datatype
```
nehanene15:
```suggestion
#### Run a row hash validation with a NUMERIC primary key
```

docs/examples.md
```diff
@@ -45,6 +45,21 @@ data-validation validate column -sc my_bq_conn -tc my_bq_conn -tbls bigquery-pub
...
+Typecast Numeric to String by either of following:
```
nehanene15: Let's add some details here i.e "For Oracle and Postgres connections, numeric primary keys may not match between source and target data types due to precision or scale discrepancies. You may need to cast the primary key to a string for a valid match."

docs/examples.md
```diff
@@ -45,6 +45,21 @@ data-validation validate column -sc my_bq_conn -tc my_bq_conn -tbls bigquery-pub
 ````shell script
 data-validation validate row -sc my_bq_conn -tc my_bq_conn -tbls bigquery-public-data.new_york_citibike.citibike_stations --primary-keys station_id --hash '*'
 ````
+#### Run a row hash validation for all rows when database is Oracle and primary key is of Numeric Datatype
+Typecast Numeric to String by either of following: 
+* Adding the code snippet above [this](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/63d6303ae08e4f8eb9a663a14d6d60828947f027/data_validation/config_manager.py#L187):
+    ````shell script
```
nehanene15: These indents mess up the code blocks. You can preview the text in GitHub or VisualStudio to see how the markdown renders

nehanene15 (changes requested):

docs/examples.md
```diff
@@ -45,6 +45,21 @@ data-validation validate column -sc my_bq_conn -tc my_bq_conn -tbls bigquery-pub
 ````shell script
 data-validation validate row -sc my_bq_conn -tc my_bq_conn -tbls bigquery-public-data.new_york_citibike.citibike_stations --primary-keys station_id --hash '*'
 ````
+#### Run a row hash validation with a **NUMERIC** primary key
+For Oracle and Postgres connections, numeric primary keys may not match between source and target data types due to precision or scale discrepancies. You may need to cast the primary key to a string for a valid match which can be done by either of following: 
+* Adding the code snippet above [this](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/63d6303ae08e4f8eb9a663a14d6d60828947f027/data_validation/config_manager.py#L187):
```
nehanene15: I think we should remove this option since users would need to clone the repo and locally package this change to use it.

docs/examples.md
```diff
@@ -45,6 +45,21 @@ data-validation validate column -sc my_bq_conn -tc my_bq_conn -tbls bigquery-pub
...
+    ````shell script
+    primary_key_configs[0]['cast'] = 'string'
+    ````
+* When validating with yaml file update ````cast: null```` to ````cast: string````
```
nehanene15: We can get rid of this bullet and just have the YAML example under the text on line 49

docs/examples.md
```diff
@@ -45,6 +45,21 @@ data-validation validate column -sc my_bq_conn -tc my_bq_conn -tbls bigquery-pub
...
+    ````shell script
```
nehanene15:
```suggestion
    ````yaml
```

docs/examples.md
```diff
@@ -45,6 +45,21 @@ data-validation validate column -sc my_bq_conn -tc my_bq_conn -tbls bigquery-pub
...
+    primary_keys:
+      - cast: null
```
nehanene15:
```suggestion
      - cast: string
```

nehanene15 (changes requested):

nehanene15 (approved):
Thanks for making the changes, LGTM!

third_party/ibis/ibis_oracle/client.py
nehanene15: [Here](https://cx-oracle.readthedocs.io/en/latest/user_guide/sql_execution.html#fetch-data-types) all the supported data types in cx_oracle. 

Based on [this](https://cx-oracle.readthedocs.io/en/latest/api_manual/cursor.html#Cursor.description) Cursor description should include precision/scale. So we could use that to check whether to map NUMBER to float64 or int64.
```
for row in cur.proxy._cursor_description():
      precision, scale = row[4], row[5]
```

PR #653, opened 2022-12-21
feat: Issue619 determine the method for generating batches efficiently

### LINKED ISSUE #619: Determine the method for generating batches efficiently
Sub-issue of issue #598 

The fastest solution will be to use a MIN/MAX split of a numeric primary key and compute the batches external to the database in order to generate config files for efficient partitioned execution.

Longer term we may need to look into more complex solutions such as RANK_ORDER and MOD division in order to generate batches across combo analytical keys.

Exit criteria for this ticket should be the generation of multiple YAML files that contain partitioned filters.

### LINKED ISSUE #662: test: Add unit test cases for generate-table-partitions command
Add unit test cases for file `partition_builder.py`

mohdt786:
Closes issue #619, #662
Added Partition support to generate multiple yaml config files

New: 

**- Arguments - file: cli_tools.py**

    1. New command 'generate-partitions' added to generate partitions for the following Validation types:
        1. row 
    2. --partition-type: Specify the type of partition logic:
        1. primary_key
        2. primary_key_mod(TODO)
        3. hash_mod(TODO)
    3. --partition-num: Number of partitions/config files to create.
        Range=[1,1000]
        If specified value is greater than count(*), value if coalesced to count(*)
    4. --config-dir: Directory Path to store YAML Config Files
    5. Added required arguments group to distinguish from optional arguments
    6. Added mutually exclusive arguments group for --hash and --concat
    Example: data-validation generate-partitions row \
                -sc BQ_CONN \
                -tc BQ_CONN \ 
                -tbls bigquery-public-data.new_york_citibike.citibike_stations,mohammedturky-sql.dvt.citibike_stations \   
                --primary-keys station_id,region_id \ 
                --hash \* \
                --filter-status fail \
                --filters 'station_id>3000:station_id>3000' \
                --config-dir partitions_dir \
                --partition-type primary_key \
                --partition-num 20

    For details, use: data-validation generate-partitions -h

**- Partition methods - file: __main__.py**
```
partition_and_store_config_files(): Build ConfigManager and Route based on Validate command
```
**- Partition methods - file: partition_builder.py**

```
1. _get_arg_partition_type(): Get config Directory from args
2. _get_arg_partition_type(): Get partition type from args
3. _get_yaml_from_config(): Convert ConfigManager object to YAML
4. partition_configs(): Route according to partition type
5. _get_primary_key_partition_filters(): Get partition filters for Primary key type partition logic
6. _add_partition_filters_and_store(): Add partition filters to generate multiple yamls and store
```
**- Partition methods - file: partition_row_builder.py**
```
1. _compile_query(): Return Ibis Query object
2. get_max_query(): Build max query
3. get_min_query(): Build min query
4. get_count_query(): Build count query
```

**- Partition methods - file: cli_tools.py**

    get_target_table_folder_path(config_dir, target_folder_name): Create and return target directory

**- Partition methods - file: state_manager.py**

    create_partition_config_directory(config_dir: str,target_folder_name: str): Create target directory for each table

**- Type Hints and Doc string:**

    1. Added Type Hints to the above methods
    7. Added Doc string with desc, args and return type for above methods


mohdt786:
cc: @Raniksingh

data_validation/cli_tools.py
```diff
@@ -331,6 +363,176 @@ def _configure_validate_parser(subparsers):
...
+        "-hash",
+        help="Comma separated list of columns for hash 'col_a,col_b' or * for all columns",
+    )
+    mutually_exclusive_arguments.add_argument(
+        "--concat",
+        "-concat",
+        help="Comma separated list of columns for concat 'col_a,col_b' or * for all columns",
+    )
+
+    row_partition_parser.add_argument(
```
nehanene15: Comparison fields should also be in the mutually_exclusive_arguments since a user can only do hash, concat, or comp-fields

data_validation/cli_tools.py
```diff
@@ -169,9 +170,40 @@ def configure_arg_parser():
...
     _configure_raw_query(subparsers)
     _configure_beta_parser(subparsers)
+    _configure_get_partition_parser(subparsers)
     return parser
 
 
+def _configure_get_partition_parser(subparsers):
+    """Configure arguments to generate partitioned config files."""
+    validate_parser = subparsers.add_parser(
+        "get-partitions",
```
nehanene15: Can we rename this 'generate-partitions'?

data_validation/cli_tools.py
```diff
@@ -169,9 +170,40 @@ def configure_arg_parser():
     _configure_find_tables(subparsers)
     _configure_raw_query(subparsers)
     _configure_beta_parser(subparsers)
+    _configure_get_partition_parser(subparsers)
     return parser
 
 
+def _configure_get_partition_parser(subparsers):
+    """Configure arguments to generate partitioned config files."""
+    validate_parser = subparsers.add_parser(
```
nehanene15: Can we rename this partition_parser? This is confusing since we have a validate_parser for the `data-validation validate` command

data_validation/cli_tools.py
```diff
@@ -169,9 +170,40 @@ def configure_arg_parser():
...
+        help="Generate partitions for validation and store the Config files in a directory",
+    )
+
+    # Keep these in order to support data-validation run command for backwards-compatibility
+    validate_parser.add_argument("--type", "-t", help="Type of Validation")
+    validate_parser.add_argument(
+        "--result-handler-config", "-rc", help="Result handler config details"
+    )
+
+    validate_subparsers = validate_parser.add_subparsers(
```
nehanene15: See above - let's rename this to partition_subparsers

data_validation/cli_tools.py
```diff
@@ -331,6 +363,176 @@ def _configure_validate_parser(subparsers):
...
+        "-comp-fields",
+        help="Individual columns to compare. If comparing a calculated field use the column alias.",
+    )
+    row_partition_parser.add_argument(
+        "--threshold",
+        "-th",
+        type=threshold_float,
+        help="Float max threshold for percent difference",
+    )
+    row_partition_parser.add_argument(
```
nehanene15: We can't have grouped-columns with row validation

data_validation/cli_tools.py
```diff
@@ -331,6 +363,176 @@ def _configure_validate_parser(subparsers):
...
+        "--random-row-batch-size",
+        "-rbs",
+        help="Row batch size used for random row filters (default 10,000).",
+    )
+    _add_common_partition_arguments(row_partition_parser, required_arguments)
+
+
+def _configure_custom_query_partition_parser(custom_query_partition_parser) -> None:
```
nehanene15: I think we shouldn't add a custom_query partition parser in this PR if we don't plan on supporting it yet

data_validation/state_manager.py
```diff
@@ -220,3 +221,26 @@ def _list_gcs_directory(self, directory_path: str) -> List[str]:
         ]
 
         return blobs
+
+    def create_partition_config_directory(
```
nehanene15: If the user supplies a directory with 'gs://' we can create the directory in GCS, otherwise it will be local. PR #654 adds support for reading from a GCS or local directory that you can see as an example

data_validation/__main__.py
```diff
@@ -472,10 +501,284 @@ def store_yaml_config_file(args, config_managers):
...
+    filters for primary_key type Partition.
 
+    Args:
+        args(Namespace): User specified Arguments.
+
+    Returns:
+        A list of type ConfigManager to get count, min and max of primary_key column.
+    """
+    agg_config_managers = []
+    config_type = consts.COLUMN_VALIDATION
```
nehanene15: Why is this being set to column validation?

data_validation/__main__.py
```diff
@@ -472,10 +501,284 @@ def store_yaml_config_file(args, config_managers):
...
+    filter_config, labels, threshold = [], [], 0.0
+    if args.filters:
+        filter_config = cli_tools.get_filters(args.filters)
+
+    labels = []
+    mgr = state_manager.StateManager()
+    source_client = clients.get_data_client(mgr.get_connection_config(args.source_conn))
+    target_client = clients.get_data_client(mgr.get_connection_config(args.target_conn))
+
+    format = "table"
```
nehanene15: We should be getting all these variables from the args parameters

data_validation/__main__.py
```diff
@@ -472,10 +501,284 @@ def store_yaml_config_file(args, config_managers):
...
+            target_client=target_client,
+            result_handler_config=result_handler_config,
+            filter_config=filter_config,
+            filter_status=filter_status,
+            verbose=args.verbose,
+        )
+
+        primary_key = primary_keys[0]
+        aggregate_configs = [config_manager.build_config_count_aggregate()]
+        aggregate_configs += config_manager.build_config_column_aggregates(
```
nehanene15: Why are we building column aggregates?

data_validation/__main__.py
```diff
@@ -472,10 +501,284 @@ def store_yaml_config_file(args, config_managers):
...
+    elif partition_type == "hash_mod":
+        # TODO: Add support for Hash + Mod
+        raise ValueError(f"Partition Type: '{partition_type}' is not supported")
+
+    _add_partition_filters_and_store(
+        config_managers, partition_filters, config_dir, args
+    )
+
+
+def _get_primary_key_partition_filters(args: Namespace) -> List[List[str]]:
```
nehanene15: We want to build this query using Ibis, not pandas ideally. This logic should also go in a new file query_builder/partitioned_row_builder.py or something similar. It will be very similar to random_row_builder where it [builds the query](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/data_validation/query_builder/random_row_builder.py#L110) using Ibis.

For example, in Ibis you can do table.count() or table[primary_key].min() and then execute the query on source or target.

nehanene15 (commented):
The biggest suggestion is to use Ibis instead of pandas to generate the partitions similar to the pattern in random row builder

data_validation/cli_tools.py
mohdt786: Done!

data_validation/cli_tools.py
mohdt786: Yes sounds better.
Done!

data_validation/cli_tools.py
mohdt786: Yes.
Done!

data_validation/cli_tools.py
mohdt786: Done!

data_validation/cli_tools.py
mohdt786: `grouped-columns` was part of the previous parser, hence had added the same arguments
```    
row_parser.add_argument(
        "--grouped-columns",
        "-gc",
        help="Comma separated list of columns to use in GroupBy 'col_a,col_b'",
    )
```

mohdt786: Removed `grouped-columns` from arguments for row validation in `row_partition_parser`

data_validation/cli_tools.py
mohdt786: Yes, makes sense
Removed custom_query partition parser

data_validation/state_manager.py
mohdt786: While creating connections, user specifies `PSO_DV_CONFIG_HOME`, if user specifies it as a GCS path, connection details are saved on GCS. Moreover, the YAML Config files are then also saved in the same path.

Reference: https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/docs/connections.md

> The connections can also be stored in GCS using PSO_DV_CONFIG_HOME. To do so simply add the GCS path to the environment. Note that if this path is set, query validation configs will also be saved here.
> 
> eg. export PSO_DV_CONFIG_HOME=gs://my-bucket/my/connections/path/

Hence, to save the partitions, user only needs to specify the Directory name, and the location will be decided by `PSO_DV_CONFIG_HOME`

mohdt786: Do we need a separate check, which, irrespective of `PSO_DV_CONFIG_HOME` decides the location of Directory based on the path supplied by the user at `--config-dir`?

.gitignore
```diff
@@ -84,7 +84,14 @@ terraform.rc
 
 # Custom
 *.yaml
+partitions_dir
 
 # Test temp files
 source_table_data.json
 target_table_data.json
+data_validation/cli_tools.py
```
nehanene15: We should remove this since we usually want to include changes to cli_tools

data_validation/cli_tools.py
```diff
@@ -168,9 +172,34 @@ def configure_arg_parser():
...
+    partition_parser.add_argument(
+        "--result-handler-config", "-rc", help="Result handler config details"
+    )
+
+    partition_subparsers = partition_parser.add_subparsers(
+        dest="validate_cmd",
+        help="Type of Validation.",
+    )
+
+    row_partition_parser = partition_subparsers.add_parser(
```
nehanene15: I think we should only support row validations. Column validations aggregate results so memory isn't an issue (i.e count * or sum col_a will always return one row as a result, but row validations return multiple).

We can get rid of this subparser: `data-validation generate-table-partitions -sc bq -tc bq ...`

data_validation/cli_tools.py
```diff
@@ -168,9 +172,34 @@ def configure_arg_parser():
...
     _configure_raw_query(subparsers)
     _configure_beta_parser(subparsers)
+    _configure_partition_parser(subparsers)
     return parser
 
 
+def _configure_partition_parser(subparsers):
+    """Configure arguments to generate partitioned config files."""
+    partition_parser = subparsers.add_parser(
+        "generate-partitions",
```
nehanene15: Maybe we can change this to `--generate-table-partitions` so it's easy to understand what it's partitioning and how it's different from random row

docs/examples.md
```diff
@@ -30,6 +30,12 @@ Above command creates a YAML file named citibike.yaml that can be used to run va
 data-validation run-config -c citibike.yaml
 ````
 Above command executes validations stored in a config file named citibike.yaml. 
+
+#### Generate partitions and save as multiple configuration files
+````shell script
+data-validation generate-partitions row -sc my_bq_conn -tc my_bq_conn -tbls bigquery-public-data.new_york_citibike.citibike_stations --primary-keys station_id --hash '*' --filters 'station_id>3000' -cdir partitions_dir --partition-type primary_key --partition-num 200
+````
+Above command creates multiple partitions based on primary key. Number of configuration files is decided by `--partition-num`
```
nehanene15: Let's also add the CLI and a description of this feature in the README similar to [this](https://github.com/GoogleCloudPlatform/professional-services-data-validator#row-validations). We can have a new section titled 'Generate Table Partitions for Large Table Row Validations' with the command description.
We should also include that we only support monotonically increasing keys currently

data_validation/partition_builder.py
```diff
@@ -0,0 +1,263 @@
...
+
+            target_partition_row_builder = PartitionRowBuilder(
+                primary_key,
+                config_manager.target_client,
+                config_manager.target_schema,
+                config_manager.target_table,
+                validation_builder.target_builder,
+            )
+
+            # accepted_datatypes = ("int64", "int32")
```
nehanene15: Should this be uncommented?

data_validation/cli_tools.py
```diff
@@ -641,6 +733,71 @@ def _add_common_arguments(parser):
...
+    required_arguments.add_argument(
+        "--target-conn", "-tc", required=True, help="Target connection name"
+    )
+    required_arguments.add_argument(
+        "--config-dir",
+        "-cdir",
+        required=True,
+        help="Directory Path to store YAML Config Files",
+    )
+    required_arguments.add_argument(
```
nehanene15: I think we should get rid of this flag and instead just support a --partition-key flag. This way the partition key and PK don't necessarily need to be the same column (if no --partition-key is supplied, we can use the PK). In the future when we support hashmod or partitionkeymod we can add flags depending on what user input we expect.

.gitignore
mohdt786: Done

data_validation/cli_tools.py
mohdt786: Yes. Done

data_validation/partition_builder.py
mohdt786: There is a datatype check below, will remove commented block
```
# If Primary key is non numeric, raise Type Error
if not (source_min.is_integer() and target_min.is_integer()):
    raise TypeError(
        f"Supplied Primary key is not of type Numeric: {primary_key}"
    )
```

data_validation/cli_tools.py
mohdt786: Done. Removed `--partition-type` flag. Default partitions would be generated on `--partition-key`. 
`--partition-key` has been made optional, hence in its absence, it will default to first `--primary-keys`

data_validation/cli_tools.py
mohdt786: Done

docs/examples.md
mohdt786: Done

README.md
```diff
@@ -197,7 +197,73 @@ data-validation (--verbose or -v) (--log-level or -ll) validate row
   [--filter-status or -fs STATUSES_LIST]
                         Comma separated list of statuses to filter the validation results. Supported statuses are (success, fail). If no list is provided, all statuses are returned.
 ```
+#### Generate Table Partitions for Large Table Row Validations
 
+Below is the command syntax for Generating Tabe Partitions for performing Row Validation on large Tables.
```
nehanene15:
```suggestion
Below is the command syntax for generating table partitions in order to perform row validations on large tables with memory constraints. 
```

README.md
```diff
@@ -197,7 +197,73 @@ data-validation (--verbose or -v) (--log-level or -ll) validate row
...
+The following command splits Row Validation into multiple steps by generating multiple Config files which can then be run in Parallel or independently.
```
nehanene15:
```suggestion
The command generates and stores multiple YAML configs that represent chunks of the large table using filters (`WHERE partition_key > X AND partition_key < Y`). You can then run the configs in the directory serially with the `data-validation configs run --config-dir PATH` command as described [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator#yaml-configuration-files).
```

README.md
```diff
@@ -197,7 +197,73 @@ data-validation (--verbose or -v) (--log-level or -ll) validate row
...
+
+The command takes the same parameters as required for `Row Validation` *plus* few commands to implement the partitioning logic.
```
nehanene15: Since we already describe the flags needed below, I don't think we need lines 207-213. We should keep the note that only monotonically increasing keys are supported though.

README.md
nehanene15: Does this PR support GCS based config-dir paths?

README.md
```diff
@@ -197,7 +197,73 @@ data-validation (--verbose or -v) (--log-level or -ll) validate row
...
+                        i.e 'bigquery-public-data.new_york_citibike.citibike_trips'
+  --primary-keys PRIMARY_KEYS, -pk PRIMARY_KEYS
+                        Comma separated list of primary key columns 'col_a,col_b'
+  --comparison-fields or -comp-fields FIELDS
+                        Comma separated list of columns to compare. Can either be a physical column or an alias
+                        See: *Calculated Fields* section for details
+  --hash COLUMNS        Comma separated list of columns to hash or * for all columns 
+  --concat COLUMNS      Comma separated list of columns to concatenate or * for all columns (use if a common hash function is not available between databases)
+  --config-dir CONFIG_DIR, -cdir CONFIG_DIR
+                        Directory Path to store YAML Config Files
```
nehanene15, line 231: We should specify if this path is local, GCS, or either.

README.md
```diff
@@ -197,7 +197,73 @@ data-validation (--verbose or -v) (--log-level or -ll) validate row
...
+  --partition-num [1-1000], -pn [1-1000]
+                        Number of partitions/config files to generate
+                        In case this value exceeds the row count of the source/target table, its will be decreased to max(source_row_count, target_row_count)
+  [--partition-key PARTITION_KEY, -partkey PARTITION_KEY]
+                        Column on which the partitions would be generated. Column type must be integer. Defaults to Primary key
+  [--threshold THRESHOLD, -th THRESHOLD]
```
nehanene15: We shouldn't support this threshold flag as we aren't running a validation and can't compare percent difference.

README.md
```diff
@@ -197,7 +197,73 @@ data-validation (--verbose or -v) (--log-level or -ll) validate row
...
+                        Float max threshold for percent difference
+  [--filters SOURCE_FILTER:TARGET_FILTER]
+                        Colon spearated string values of source and target filters.
+                        If target filter is not provided, the source filter will run on source and target tables.
+                        See: *Filters* section
+  [--use-random-row, -rr]
+                        Finds a set of random rows of the first primary key supplied.
+  [--random-row-batch-size RANDOM_ROW_BATCH_SIZE, -rbs RANDOM_ROW_BATCH_SIZE]
+                        Row batch size used for random row filters (default 10,000).
+  [--result-handler-config RESULT_HANDLER_CONFIG, -rc RESULT_HANDLER_CONFIG]
```
nehanene15: result-handler-config is deprecated so we shouldn't document it. We just keep it in cli_tools for backwards compatibility

README.md
```diff
@@ -197,7 +197,73 @@ data-validation (--verbose or -v) (--log-level or -ll) validate row
...
+                        Result handler config details
+  [--bq-result-handler or -bqrh PROJECT_ID.DATASET.TABLE]
```
nehanene15: We shouldn't support result handler since we don't write any validation results to BQ.

README.md
```diff
@@ -197,7 +197,73 @@ data-validation (--verbose or -v) (--log-level or -ll) validate row
...
+                        BigQuery destination for validation results. Defaults to stdout.
+                        See: *Validation Reports* section
+  [--labels LABELS, -l LABELS]
+                        Key value pair labels for validation run
+  [--service-account SERVICE_ACCOUNT, -sa SERVICE_ACCOUNT]
+                        Path to SA key file for result handler output
+  [--format FORMAT, -fmt FORMAT]
+                        Set the format for printing command output, Supported formats are (text, csv, json, table). Defaults to table
+  [--filter-status FILTER_STATUS, -fs FILTER_STATUS]
```
nehanene15: We don't need any of the above flags (line 252-265) as we aren't running validations, just creating configs.

README.md
```diff
@@ -197,7 +197,73 @@ data-validation (--verbose or -v) (--log-level or -ll) validate row
...
+                        In case this value exceeds the row count of the source/target table, its will be decreased to max(source_row_count, target_row_count)
+  [--partition-key PARTITION_KEY, -partkey PARTITION_KEY]
+                        Column on which the partitions would be generated. Column type must be integer. Defaults to Primary key
+  [--threshold THRESHOLD, -th THRESHOLD]
+                        Float max threshold for percent difference
+  [--filters SOURCE_FILTER:TARGET_FILTER]
+                        Colon spearated string values of source and target filters.
+                        If target filter is not provided, the source filter will run on source and target tables.
+                        See: *Filters* section
+  [--use-random-row, -rr]
```
nehanene15: Supporting random row here wouldn't make sense since we want to break a full table into chunks. Random row would select rows and add a WHERE clause to select random rows; we want to cater this command to only full table row validations

data_validation/__main__.py
```diff
@@ -217,11 +210,15 @@ def build_config_from_args(args, config_manager):
     return config_manager
 
 
-def build_config_managers_from_args(args):
+def build_config_managers_from_args(args: Namespace, validate_cmd: str = None):
     """Return a list of config managers ready to execute."""
     configs = []
 
-    validate_cmd = args.validate_cmd.capitalize()
+    # Since `generate-table-partitions` is default to `validate_cmd=row`,
```
nehanene15:
```suggestion
    # Since `generate-table-partitions` defaults to `validate_cmd=row`,
```

data_validation/__main__.py
```diff
@@ -491,10 +488,34 @@ def store_yaml_config_file(args, config_managers):
...
+
+    Args:
+        args (Namespace): User specified Arguments
 
+    Returns:
+        None
+    """
+    # Default Validate Type
+    validate_cmd = consts.ROW_VALIDATION
+    config_managers = build_config_managers_from_args(args, validate_cmd)
```
nehanene15:
```suggestion
    config_managers = build_config_managers_from_args(args, consts.ROW_VALIDATION)
```

data_validation/__main__.py
```diff
@@ -491,10 +488,34 @@ def store_yaml_config_file(args, config_managers):
...
+    """Build multiple YAML Config files using user specified partition logic
+
+    Args:
+        args (Namespace): User specified Arguments
 
+    Returns:
+        None
+    """
+    # Default Validate Type
+    validate_cmd = consts.ROW_VALIDATION
```
nehanene15: Instead of declaring this separately, we can pass in the constant directly to the next line.

data_validation/cli_tools.py
```diff
@@ -168,9 +172,104 @@ def configure_arg_parser():
...
+        "generate-table-partitions",
+        help=(
+            "Generate partitions for validation and store the Config files in "
+            "a directory"
+        ),
+    )
+
+    # Group all optional arguments together
+    optional_arguments = partition_parser.add_argument_group("optional arguments")
+    optional_arguments.add_argument(
```
nehanene15, line 191: We don't need threshold, use random row, or random row batch size since we aren't running a typical validation

data_validation/cli_tools.py
```diff
@@ -641,6 +740,75 @@ def _add_common_arguments(parser):
...
+        "-pn",
+        required=True,
+        help="Number of partitions/config files to generate",
+        type=int,
+        choices=range(1, 1001),
+        metavar="[1-1000]",
+    )
+
+    # Optional arguments
+    optional_arguments.add_argument(
```
nehanene15, line 779: Don't need flags from line 770 - 801 as mentioned above

data_validation/consts.py
```diff
@@ -90,6 +90,8 @@
     CUSTOM_QUERY,
 ]
 
+# Partition types
+PARTITION_TYPES = ["primary_key", "primary_key_mod", "hash_mod"]
```
nehanene15: We can remove this and line 71 since it's unused for now.

data_validation/partition_builder.py
```diff
@@ -0,0 +1,241 @@
...
+                    "target": filter_list[pos],
+                }
+                # Append partition new filter
+                config_manager.filters.append(filter_dict)
+
+                # Save partition
+                yaml_configs = self._get_yaml_from_config(config_manager)
+                target_file_name = "0" * (4 - len(str(pos))) + str(pos) + ".yaml"
+                config_file_path = os.path.join(target_folder_path, target_file_name)
+                cli_tools.store_validation(config_file_path, yaml_configs)
```
nehanene15: Each store_validation call logs a success message for the file being written [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/data_validation/state_manager.py#L175) - if a user has 1000 partitions this would log a lot of times... did you encounter this while testing?

README.md
mohdt786: Done

README.md
mohdt786: Done

README.md
mohdt786: Done

README.md
mohdt786: Done

README.md
mohdt786: Done

data_validation/__main__.py
mohdt786: Done

data_validation/__main__.py
mohdt786: Done

data_validation/__main__.py
mohdt786: Done

data_validation/cli_tools.py
mohdt786: Done

README.md
mohdt786: Done

README.md
mohdt786: Done

data_validation/consts.py
mohdt786: Done

data_validation/cli_tools.py
mohdt786: Removed from readme.md but kept required ones in cli_tools.py to build ConfigManager object

data_validation/cli_tools.py
```diff
@@ -646,6 +728,79 @@ def _add_common_arguments(parser):
...
+    required_arguments.add_argument(
+        "--target-conn", "-tc", required=True, help="Target connection name"
+    )
+    required_arguments.add_argument(
+        "--config-dir",
+        "-cdir",
+        required=True,
+        help="Directory Path to store YAML Config Files. "
+        "GCS: Provide a full gs:// path of the target directory. "
+        "Eg: `gs://<BUCKET>/partiitons_dir`. "
```
nehanene15, line 764:
```suggestion
        "Eg: `gs://<BUCKET>/partitons_dir`. "
```

data_validation/__main__.py
```diff
@@ -258,12 +261,18 @@ def build_config_managers_from_args(args):
...
 
-    use_random_rows = (
-        None if config_type == consts.SCHEMA_VALIDATION else args.use_random_row
-    )
-    random_row_batch_size = (
-        None if config_type == consts.SCHEMA_VALIDATION else args.random_row_batch_size
-    )
+    # Random row validation is not supported for schema validation and
+    # generate-table-partitions
+    use_random_rows = None
```
nehanene15: Instead of doing this, we should just leave the random_rows/random_row_batch_size/threshold as an argument in the partition parser. We should just remove it from documentation so people don't use it.

mohdt786: Okay, done

data_validation/cli_tools.py
mohdt786: Fixed the typo

tests/unit/test_partition_builder.py
```diff
@@ -0,0 +1,491 @@
...
+
+
+@pytest.fixture
+def module_under_test():
+    import data_validation.partition_builder
+
+    return data_validation.partition_builder
+
+
+def teardown_module(module):
```
nehanene15, line 263: Where do we use this function?

tests/unit/test_partition_builder.py
```diff
@@ -0,0 +1,491 @@
...
+    _create_table_file(SOURCE_TABLE_FILE_PATH, source_json_data)
+    _create_table_file(TARGET_TABLE_FILE_PATH, target_json_data)
+
+    config_manager = _generate_config_manager("my_table")
+    config_managers = [config_manager]
+
+    parser = cli_tools.configure_arg_parser()
+    mock_args = parser.parse_args(CLI_ARGS_JSON_SOURCE)
+
+    mock_partition_filters = _generate_fake_partition_filters(
```
nehanene15: Instead of generating fake partition filters, we can just manually check that the partition_filters_list has the correct length/attributes that we expect so we don't have to copy the logic over. 
```
expected_partition_filters_list = ["key >= 0 and key < 10", "key >= 11 and key < 20"]
assert len(partition_filters_list[0]) == 20
assert partition_filters_list[0] == expected_partition_filters_list
```
This can also be done for _generate_fake_yaml_configs() to avoid having any logic in the tests.

nehanene15 (commented):
LGTM besides a 2 comments on the tests - thanks!

tests/unit/test_partition_builder.py
mohdt786: It is a clean up method called automatically by PyTest at the end of all the test methods. Used here to delete the directory created while testing `test_store_yaml_partitions_local()`

tests/unit/test_partition_builder.py
mohdt786: Removed generation of `partition_filters_list` and `yaml_configs_list` via function.

Added expected `PARTITION_FILTERS_LIST` and `YAML_CONFIGS_LIST` to a json file `tests/unit/test_inputs/test_partition_builder.json` since expected `YAML_CONFIGS_LIST` is too large to store in `tests/unit/test_partition_builder.py` and would make it less readable.

tests/unit/test_inputs/test_partition_builder.json
```diff
@@ -0,0 +1,1956 @@
+{
+    "PARTITION_FILTERS_LIST": [
```
nehanene15: This file is really long... can we adjust to 3-5 partitions instead of 20?

mohdt786: Done. Reduced test input size

nehanene15 (approved):

PR #654, opened 2022-12-22
feat: Support config directory for running validations and add multithreading for DB queries

dmedora:
- Adds support for specifying a directory to run validation yamls from (either local or GCS).
- Adds multithreading to source & target database query requests, so these happen simultaneously, effectively halving total time waiting for external query execution.


data_validation/__main__.py
```diff
@@ -314,12 +317,27 @@ def build_config_managers_from_args(args):
...
+    else:
+        config_file_path = _get_arg_config_file(args)
+        config_managers = build_config_managers_from_yaml(args, config_file_path)
+
+    run_validations(args, config_managers)
+
+
+def build_config_managers_from_yaml(args, config_file_path):
+    """Returns List[ConfigManager] instances ready to be executed."""
+    if "config_dir" in args and args.config_dir:
```
nehanene15, line 337: Why do we need the first conditional? Can't we just do `if args.config_dir`

nehanene15 (approved):
Looks perfect!

data_validation/__main__.py
dmedora: It's because of the (still supported) `run-config` legacy option - I've only added the `config-dir` subparser to the newer `configs run`, so when `run-config` is used, the `config-dir` variable is not initialized to null and so it errors.

PR #656, opened 2022-12-26
test: Add system integration test for Postgres row validation with different datatypes

### LINKED ISSUE #655: Postgres (add different data types for row hash in system Integration test)
This is sub-issue for issue#611. This should involve adding different data types test in the row validation (string/varchar, integer, decimal, & date)

sharangagarwal:
Closes #655

Adds system integration test for Postgres row validation with different datatypes

- Modified cloud_sql function in tests/system/data_sources/test_postgres.py to add postgres_data_row.sql to generate dummy data with serial, int, text, char, varchar, double precision, numeric, timestamp and date datatypes
- Modified test_postgres_row in tests/system/data_sources/test_postgres.py to change the consts.CONFIG_CALCULATED_FIELDS to use the hash for these columns

### DIFF
--- a/tests/system/data_sources/test_postgres.py
+++ b/tests/system/data_sources/test_postgres.py
@@ -413,6 +413,14 @@ def test_postgres_row(cloud_sql):
                 "depth": 5,
             },
         ],
+        consts.CONFIG_PRIMARY_KEYS: [
+            {
+                "source_column": "serial_col",
+                "target_column": "serial_col",
+                "field_alias": "serial_col",
+                "cast": None,
+            }
+        ],
         consts.CONFIG_FORMAT: "table",
     }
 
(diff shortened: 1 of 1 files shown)


tests/system/data_sources/test_postgres.py
```diff
@@ -60,6 +60,7 @@ def cloud_sql(request):
         # If instance already exists, returns host IP and does not add new data
         postgres_instance.setup()
         postgres_instance.add_data("gs://pso-kokoro-resources/postgres_data.sql")
+        postgres_instance.add_data("gs://pso-kokoro-resources/postgres_data_row.sql")
```
Raniksingh: Can we create folders for each source and place the specific sql files inside it?

sharangagarwal: Moved the new file postgres_data_row.sql to 
**gs://pso-kokoro-resources/test/postgres/system_integration/**
I also copied gs://pso-kokoro-resources/postgres_data.sql to the above folder but since it might be used in other  development branches, I haven't deleted it for now

Raniksingh (approved):

PR #663, opened 2023-01-19
fix: Error handling for batch processing of config files

kanhaPrayas:


data_validation/__main__.py
```diff
@@ -478,7 +478,21 @@ def run_validations(args, config_managers):
     """
     # TODO(issue/31): Add parallel execution logic
     for config_manager in config_managers:
-        run_validation(config_manager, verbose=args.verbose)
+        if config_manager.config and consts.CONFIG_FILE in config_manager.config:
```
nehanene15, line 481: Should this be checking for config_dir in config_manager.config instead of config_file? 
[This](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/data_validation/__main__.py#L327) is where we iterate through the individual files in the config dir. Maybe we should just log the file names within the config_runner function instead of pushing this logic to run_validation() so we keep track of each file we iterate through

nehanene15 (approved):

Issue #670, opened 2023-01-23
Making primary key flag compulsory in custom-query-row validation


piyushsarraf:
Earlier primary key was not mandatory because the hash values were compared with `hash_all` column as reference, but after recent [change](https://github.com/GoogleCloudPlatform/professional-services-data-validator/commit/a7f800d9958b3e971ea99eade679cfc77c192c7c#diff-e55665966f6165967c646d72cbfde78c62ce8c1cf32149dfbfe0a5c3e3a4fda9R295-R299) where hash values are getting compared with `primary key` column as reference, DVT throws error of None primary  key if `--primary_key` flag is not present.

nehanene15:
This can be a sub-issue of Issue #673

Issue #673, opened 2023-01-25
Optimize CLI tools arg parser


nehanene15:
Currently the arg parser does not make use of all the features of [argparse](https://docs.python.org/3/library/argparse.html). We should ensure that for the subparsers, we can fail fast if some required args are not provided, or if only certain values are accepted by using the `required` or `choices` or `default` parameters ([Reference](https://docs.python.org/3/library/argparse.html#quick-links-for-add-argument))

We should also remove support for the data-validation run command. ([Example](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/data_validation/cli_tools.py#L213))

nehanene15:
Currently, the column level parser also accepts hash, concat, comp-fields which makes the `data-validation validate column -h` command invalid.

mohdt786:
Issues solved:

1. deprecated: Remove support for data-validation run-config command
2. test: 🎨 Remove tests for run-config command
3. fix: 🔥 Add a subparser for validate custom-query: validate custom-query row and validate custom-query column
4. fix: Remove random-row and added --comparison-fields support for custom-query row validation
5. fix: Make --hash, --concat, and --comparison-fields required and mutually exclusive for validate row command.
6. fix: Remove --hash, --concat, and --comparison-fields from validate column command
7. fix: Split optional and required arguments into different groups for better visibility while using -h flag
8. deprecated: Deprecate --result-handler-config and --type in validate and generate-table-partitions command
9. fix: Remove data-validation run-config -c ex_yaml.yaml from cli usage doc_string
10. fix: Remove --tables-list from validate custom-query command
11. fix: Remove --grouped-columns from validate row command
12. fix: Remove --use-random-row and --random-row-batch-size for generate-table-partitions command
13. fix: Remove primary-keys argument for validate column and validate custom-query column
14. feat: Add support for --comparison-fields for validate custom-query row
15. refactor: 🎨 Refactored cli_tools.py and main.py
16. fix: Reorganise function build_config_managers_from_args()
17. docs: Update custom-query CLI command in README.md

mohdt786:
Updated:

1. Remove `run-config` command
2. Make `--hash`, `--concat`, and `--comparison-fields` required and mutually exclusive for `validate row` command.
3. Remove `--hash`, `--concat`, and `--comparison-fields` from `validate column` command
4. Split optional and required arguments into different groups for better visibility while using `-h` flag
[...]

PR #690, opened 2023-02-01
fix: Making primary key compulsory for Custom Query Row Validation

piyushsarraf:
* fix: Making primary key compulsory for Custom Query Row Validation (#670) 

* Adding more datatype support for oracle custom query row validation.

### DIFF
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -585,6 +585,7 @@ def _configure_custom_query_parser(custom_query_parser):
         "--primary-keys",
         "-pk",
         help="Comma separated list of primary key columns 'col_a,col_b'",
+        required=True,
     )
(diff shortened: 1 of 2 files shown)


data_validation/cli_tools.py
```diff
@@ -585,6 +585,7 @@ def _configure_custom_query_parser(custom_query_parser):
         "--primary-keys",
         "-pk",
         help="Comma separated list of primary key columns 'col_a,col_b'",
+        required=True,
```
nehanene15, line 588: The issue with adding required here is that custom query column parser doesn't require a PK. PR #673 is doing something similar

nehanene15:
Closing this as stale since we merged PR #673 that makes PKs required for custom query row validations.

PR #701, opened 2023-02-03
fix!: issue673 optimize CLI tools arg parser

### LINKED ISSUE #673: Optimize CLI tools arg parser

mohdt786:
### Summary:

> CLI command for data-validation custom-query has changed. A new subparser has been added to support `data-validation custom-query row` and `data-validation custom-query column` validations. Please refer [README.md](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/README.md#custom-query-column-validations)

> Support for `data-validation run-config -c <configs.yaml>` command has been deprecated. Please use `data-validation configs run -c <configs.yaml>` command. Refer [README.md](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/README.md#yaml-configuration-files)

> Added support for `--comparison-fields` for validate custom-query row

### Detailed changes:
1. deprecated: Remove support for data-validation run-config command
2. test: 🎨 Remove tests for run-config command
3. fix: 🔥 Add a subparser for validate custom-query: validate custom-query row and validate custom-query column
4. fix: Remove random-row and added `--comparison-fields` support for custom-query row validation
5. fix: Make --hash, --concat, and --comparison-fields required and mutually exclusive for validate row command.
6. fix: Remove --hash, --concat, and --comparison-fields from validate column command
7. fix: Split optional and required arguments into different groups for better visibility while using -h flag
8. deprecated: Deprecate --result-handler-config and --type in validate and generate-table-partitions command
9. fix: Remove data-validation run-config -c ex_yaml.yaml from cli usage doc_string
10. fix: Remove --tables-list from validate custom-query command
11. fix: Remove --grouped-columns from validate row command
12. fix: Remove --use-random-row and --random-row-batch-size for generate-table-partitions command
13. fix: Remove primary-keys argument for validate column and validate custom-query column
14. feat: Add support for --comparison-fields for validate custom-query row
15. refactor: 🎨 Refactored cli_tools.py and __main__.py
16. fix: Reorganise function build_config_managers_from_args()
17. docs: Update custom-query CLI command in README.md


data_validation/cli_tools.py
```diff
@@ -436,316 +395,351 @@ def _configure_validate_parser(subparsers):
...
+        required=True,
+        help="File containing the source sql query",
+    )
+    required_arguments.add_argument(
+        "--target-query-file",
+        "-tqf",
+        required=True,
+        help="File containing the target sql query",
+    )
+    required_arguments.add_argument(
```
nehanene15: PKs are only required for custom query row validations... maybe we should make it optional

mohdt786: Done.
Added a fail fast check in `cli_tools.py` by calling: `def apply_post_parse_checks(args: Namespace, parser) -> Namespace:`

mohdt786: PKs would be a required argument in the case of `validate custom-query row` command

data_validation/cli_tools.py
```diff
@@ -146,15 +145,36 @@ def get_parsed_args() -> Namespace:
     """Return ArgParser with configured CLI arguments."""
     parser = configure_arg_parser()
     args = ["--help"] if len(sys.argv) == 1 else None
-    return parser.parse_args(args)
+    parsed_args = apply_post_parse_checks(parser.parse_args(args), parser)
+    return parsed_args
+
+
+def apply_post_parse_checks(args: Namespace, parser) -> Namespace:
```
nehanene15: Instead of doing this, maybe we should split out custom query row versus column into separate subparsers like we do for data-validation validate (row | column). This way we can have the correct set of flags for each type of validation. We can also require PKs for `data-validation validate custom-query row` and support hash/concat/comp-fields only for row.

This would be a breaking change so we should add the exclamation point 'fix!' to the PR title as per the [conventional commits guidelines](https://www.conventionalcommits.org/en/v1.0.0/), and update our documentation to support `data-validation validate custom-query (row|column)`

mohdt786: Done!
Added a subparser for `data-validation validate custom-query` to separate `row` and `column`.
Removed `--primary_keys` arg for `validate column` and `validate custom-query column`

data_validation/__main__.py
```diff
@@ -159,13 +160,24 @@ def get_calculated_config(args, config_manager):
...
-    """Return config manager object ready to execute.
+def build_config_from_args(args: Namespace, config_manager: ConfigManager):
+    """Append build configs to ConfigManager object.
 
     Args:
+        args (Namespace): User specified Arguments
         config_manager (ConfigManager): Validation config manager instance.
     """
+    # Append SCHEMA_VALIDATION configs
+    if config_manager.validation_type == consts.SCHEMA_VALIDATION:
```
nehanene15, line 171: We don't need to add a condition for schema validation since this method only gets called for non-schema validations: https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/data_validation/__main__.py#L304

data_validation/__main__.py
```diff
@@ -175,24 +187,42 @@ def build_config_from_args(args, config_manager):
...
             config_manager.append_aggregates(get_aggregate_config(args, config_manager))
 
-        elif not args.primary_keys:
-            raise ValueError(
-                "Expected valid primary keys for custom query row validation, got None."
+        # For custom-query row command
+        if args.custom_query_type == consts.ROW_VALIDATION.lower():
+
+            # Append calculated fields: --hash/--concat
+            config_manager.append_calculated_fields(
```
nehanene15, line 207: We should keep the append_calculated_fields() outside of the `if config_manager.validation_type == consts.CUSTOM_QUERY` block since calculated fields can be applied to column validations as well, i.e doing a cast before a SUM.

data_validation/__main__.py
```diff
@@ -175,24 +187,42 @@ def build_config_from_args(args, config_manager):
...
             config_manager.append_query_groups(
                 config_manager.build_column_configs(grouped_columns)
             )
-    elif config_manager.validation_type == consts.ROW_VALIDATION:
+
+    # Append ROW_VALIDATION configs
+    if config_manager.validation_type == consts.ROW_VALIDATION:
+
+        # Append calculated fields: --hash/--concat
+        config_manager.append_calculated_fields(
```
nehanene15, line 230: See above comment - we should keep the append_calculated_fields outside the conditional

tests/system/data_sources/test_bigquery.py
```diff
@@ -342,7 +341,7 @@
     "--config-file",
     CLI_CONFIG_FILE,
 ]
-EXPECTED_NUM_YAML_LINES_TIMESTAMP_MIN_MAX = 30
+EXPECTED_NUM_YAML_LINES_TIMESTAMP_MIN_MAX = 29
```
nehanene15, line 344: Why did this number go down one?

data_validation/cli_tools.py
```diff
@@ -436,316 +394,414 @@ def _configure_validate_parser(subparsers):
...
+        "-filters",
+        help="Filters in the format source_filter:target_filter",
+    )
+    optional_arguments.add_argument(
+        "--threshold",
+        "-th",
+        type=threshold_float,
+        help="Float max threshold for percent difference",
+    )
+    optional_arguments.add_argument(
```
nehanene15: We don't support random-row for custom query

data_validation/cli_tools.py
```diff
@@ -436,316 +394,414 @@ def _configure_validate_parser(subparsers):
...
+    )
+    required_mutually_exclusive.add_argument(
+        "--hash",
+        "-hash",
+        help="Comma separated list of columns for hashing a concatenate 'col_a,col_b' or * for all columns",
+    )
+    required_mutually_exclusive.add_argument(
+        "--concat",
+        "-concat",
+        help="Comma separated list of columns for concat 'col_a,col_b' or * for all columns",
```
nehanene15, line 657: Let's add comparison-field here (and to README doc) too..it should work in theory, just hasn't been added to the CLI

data_validation/__main__.py
mohdt786: I have put all the post build configs under `build_config_from_args(args, config_manager)`: 
[Reference](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/issue673-optimize-cli-tools-arg-parser/data_validation/__main__.py#L257)

Hence all the 4 types of validations are checked and compared here including schema-validations
The method is now being called for `schema validation` as well

data_validation/__main__.py
mohdt786: In the definition of `get_calculated_config(args, config_manager)`, `calculated_configs` is only populated in case when `--hash` or `--concat` are present. So it returns an empty list in case of `column validation`
[Reference](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/data_validation/__main__.py#L118)

Adding `config_manager.append_calculated_fields()` would result in an empty list being appended in case of `validate column` and `validate custom-query column`

data_validation/__main__.py
mohdt786: Since `--hash` and `--concat` args are not present for `validate custom-query column` and `validate column`, it would throw an error while checking for `--hash` and `---concat`
[Reference](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/issue673-optimize-cli-tools-arg-parser/data_validation/__main__.py#L127)

tests/system/data_sources/test_bigquery.py
mohdt786: Since `calculated fields` is not appended in case of `validate column`
Previously, it was an empty list

data_validation/cli_tools.py
mohdt786: I see there is no check present for `comp-fields` in `__main__.py` for `custom-query`, should I add there too?

mohdt786: Added this block of [code](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/data_validation/__main__.py#L196) under `validate custom-query row`

```
      if args.comparison_fields is not None:
          comparison_fields = cli_tools.get_arg_list(
              args.comparison_fields, default_value=[]
          )
          config_manager.append_comparison_fields(
              config_manager.build_config_comparison_fields(comparison_fields)
          )
```

Tested few examples and `--comparison_fields` is supported for `validate custom-query row`

data_validation/cli_tools.py
mohdt786: Done. Removed `random-row` for `custom-query`

nehanene15 (approved):
LGTM!

PR #704, opened 2023-02-06
feat: gcp secret manger support for DVT 

mokhahmed:
/gcbrun


data_validation/cli_tools.py
```diff
@@ -368,16 +368,28 @@ def _configure_validation_config_parser(subparsers):
...
+        "--secret-manger-type",
+        "-sm",
+        default=None,
+        help="Secret manger type to store credentials by default will be plain ",
+    )
+    add_parser.add_argument(
+        "--secret-manger-project-id",
+        "-sm-prj-id",
+        default=None,
+        help="Secret manger type to store credentials by default will be plain ",
```
nehanene15: This description is for type rather than project-id

data_validation/secret_manger.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class GCPSecretManger:
+    """
+    GCPSecretManger: client to access secrets stored at GCP secret manger
+    """
+
+    def __init__(self):
+        # Import the Secret Manager client library.
+        from google.cloud import secretmanager
```
nehanene15: Do we need to add google-cloud-secret-manager in our requirements?

data_validation/secret_manger.py
```diff
@@ -0,0 +1,53 @@
+class SecretMangerBuilder:
```
nehanene15: Let's add the Google copyright at the top of this - you can just copy/paste from the other files

data_validation/secret_manger.py
```diff
@@ -0,0 +1,53 @@
...
+            print(e)
+            return secret_id
+
+
+if __name__ == "__main__":
+    import os
+
+    os.environ[
+        "GOOGLE_APPLICATION_CREDENTIALS"
+    ] = "/Users/moukhtar/ws/ma-sabre-sandbox-01-dfe8f33bb3ad.json"
```
nehanene15: We should redact this path and the secret on line 52

docs/connections.md
```diff
@@ -55,11 +60,11 @@ The data validation tool supports the following connection types.
 * [DB2](#DB2)
 * [AlloyDB](#AlloyDB)
 
-As you see above, Teradata and BigQuery have different sets of custom arguments (for example project-id for BQ versus host for Teradata).
+As you see above, Teradata and BigQuery have different sets of custom arguments (for example project_id for BQ versus host for Teradata).
```
nehanene15: We want to keep the hyphen instead of the underscores in this doc since user can understand the CLI config better. Some folks mistake the underscores and use `--project_id` instead of `--project-id`

nehanene15 (commented):
Left some minor comments. Could we also add a unit test for the SecretManagerBuilder?

data_validation/cli_tools.py
```diff
@@ -368,16 +368,28 @@ def _configure_validation_config_parser(subparsers):
...
     )
     connect_subparsers = connection_parser.add_subparsers(dest="connect_cmd")
     _ = connect_subparsers.add_parser("list", help="List your connections")
-
     add_parser = connect_subparsers.add_parser("add", help="Store a new connection")
     add_parser.add_argument(
         "--connection-name", "-c", help="Name of connection used as reference"
     )
+    add_parser.add_argument(
+        "--secret-manger-type",
```
nj1973: There's a type here. "manger" should be "manager".
This has propagated throughout the changes, if you search for "manger" throughout all changes there are 90 occurrences that are missing an "a".

data_validation/secret_manger.py
nj1973: Also the file is named "secret_manger.py" rather than "secret_manager.py".

data_validation/cli_tools.py
mokhahmed: fixed the typo to manager instead of manger

data_validation/secret_manger.py
mokhahmed: removed

data_validation/secret_manger.py
mokhahmed: adding copyright headers

data_validation/secret_manger.py
mokhahmed: added the dependencies to setup.py

nehanene15 (commented):
Logic looks good; we should add unit tests for the secret manager class.

Raniksingh (approved):

PR #710, opened 2023-02-08
fix: schema validation fix for different base names of source and des…

### LINKED ISSUE #683: validate schema: --allow-list option prevents some required mappings
There is a secondary check in the code behind `--allow-list` that ensures the base data type is the same in source and target.
It will accept mapping of `decimal(10,0)` to `decimal(8,0)` but rejects `decimal(10,0)` to `int64` (`bigint`). This is rejected because the base data type of `decimal` is `decimal` and that is not the same as the base datatype of `int64` which is `int`.

This prevents these useful mappings:

- date -> timestamp
- timestamp -> date
- decimal(?,0) -> int[16/32/64]

We need to understand why these mappings were prevented and, ideally, extend the accepted base types to include those above.

kanhaPrayas:
…tination data types

### DIFF
--- a/data_validation/schema_validation.py
+++ b/data_validation/schema_validation.py
@@ -143,20 +143,20 @@ def schema_validation_matching(
                         consts.VALIDATION_STATUS_SUCCESS,
                     ]
                 )
-            elif string_val(source_field_type) in allow_list_map:
+            elif (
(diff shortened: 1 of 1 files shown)


nehanene15 (approved):
LGTM!

PR #720, opened 2023-02-13
feat: Use centralized date/time format in order to compare row data across engines

### LINKED ISSUE #680: validate row: Oracle DATE format doesn't match PostgreSQL TIMESTAMP format
The example that triggered this issue is an Oracle to PostgreSQL validation.
DVT casts date based columns to string without a format, formatting is left to the underlying engine. The Oracle default is often out of our control. Even if we can control it, Oracle DATE is to second precision where most other engines have DATE with day precision or a TIMESTAMP with fractional second precision. Again we are at the mercy of the underlying engines as to how they format the output of the CAST().

A suggested workaround for this specific Oracle to PostgreSQL project is to edit the config YAML file and for each Oracle DATE or TIMESTAMP column:

- Change type `cast` to `to_char`
- Add `default_to_char_fmt` of `"YYYY-MM-DD HH24:MI:SS"`

The project is to migrate 1000s of database therefore a manual process is not viable. Also the workaround only works for this specific project because Oracle and PostgreSQL share date formatting masks. An Oracle to BigQuery project would still be a problem.

I think a better solution is to redirect `cast` to `strftime` for any date based columns.

Approximate tasks would be:

- Support a type of `strftime` in YAML config file and support a format token too (either re-use `default_to_char_fmt` or add a new one specific to `strftime`)
- Change `config_manager.build_config_calculated_fields` to support a format parameter
- Add code to generate a suitable `strftime` format based on the column type. For example:
```
    def _strftime_format(self, column_type: Union[Date, Timestamp]) -> str:
        if isinstance(column_type, Timestamp):
            return "%Y-%m-%d %H:%M:%S"
        if isinstance(self.source_client, OracleClient):
            # Oracle DATE is a DateTime
            return "%Y-%m-%d %H:%M:%S"
        return "%Y-%m-%d"
```
- Use the above code on each source and target date-based column and pick the shortest mask for the config file. For example, if Oracle DATE is migrated to PostgreSQL date then we validate with the shorter `"%Y-%m-%d"` mask.
- Change `data_validation/config_manager.py build_config_calculated_fields()` to allow `cast` to be overridden when appropriate

I've mocked this up and it works for combinations of Oracle DATE/TIMESTAMP to PostgreSQL DATE/TIMESTAMP.

One vulnerability is that `strftime` does not currently cater for fractional seconds. Therefore the above validations are all at second precision. If we complete this issue then a follow on issue to support fractional seconds may be required.

nj1973:
Before this change casts to string of date based columns were done without a format, formatting was left to the underlying engine. This caused a few problems with `validate row` validations.

- Most engines have different default formats
- Some migrations will switch from date to timestamp or vice versa which introduces incompatible formats

This PR introduces code to use `strftime()` for all `Date` / `Timestamp` row data comparisons in place of a default `cast()`. By using a canonical format any engine specifics should be irrelevant.

Change summary:

- `config_manager.py` has been changed to get a suitable `strftime()` for any date based columns (source and target)
- `config_manager.py` will choose the shortest format out of the source and target formats. This caters for comparisons where the source is `Timestamp` but the target is `Date` (or vice versa)
- `third_party/ibis/ibis_mssql/compiler.py` has been change to have a very basic implementation of `strftime()`. Because SQL Server's `convert()` function is too restrictive for a full `strftime()` implementation we've done the bare minumum to support the canonical formats used for  `Date` and `Timestamp`

A downside is that the canonical `Timestamp` format does not support fractional seconds, there are a couple of reasons for this:

- Ibis `Timestamp` does not support `scale` in the version we use
- Existing `strftime` functions do not support fractional second formats

Once we've upgraded Ibis version we could revisit this to choose a correct fraction second format based on source and target data type.


data_validation/config_manager.py
```diff
@@ -14,9 +14,12 @@
 
 import copy
 import logging
+from typing import Optional, Union
 
 import google.oauth2.service_account
 from ibis_bigquery.client import BigQueryClient
+import ibis.expr.datatypes as dt
+from third_party.ibis.ibis_oracle.client import OracleClient
```
nehanene15: We try/catch the import of OracleClient ([example](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/data_validation/clients.py#L69)) since we don't require cx_oracle to be installed for non-Oracle validations. 

Instead of importing OracleClient here, you can do `from data_validation import clients` and refer to `clients.OracleClient` in line 717

nj1973: Perfect thanks. I've made that change and integration tests now pass.

nehanene15 (approved):
LGTM!

PR #721, opened 2023-02-13
feat: Divert cast of PostgreSQL decimal with scale>0 to to_char

### LINKED ISSUE #685: validate row: Oracle NUMBER with scale doesn't match PostgreSQL DECIMAL with scale
Fractional NUMBER column values do not match equivalent values in PostgreSQL due to differences in CAST behaviour.

For example, in Oracle:
```
SELECT CAST(CAST(100 AS NUMBER(5,2)) AS VARCHAR2(10)) FROM dual;

CAST(CAST(
----------
100
```

PostgreSQL equivalent:
```
# SELECT CAST(CAST(100 AS DECIMAL(5,2)) AS VARCHAR(10));
 varchar
---------
 100.00
```

The upshot of this is row data mismatches when using default validation type `cast` behaviour.

A suggested workaround for this specific Oracle to PostgreSQL project is to edit the config YAML file and for each Oracle NUMBER column with scale > 0:

- Change type cast to `to_char`
- Add `default_to_char_fmt` of "99999999.99" (matching p and s with the number of 9s)

The project is to migrate 1000s of database therefore a manual process is not viable. Also the workaround only works for this specific combination of source/target engines because Oracle and PostgreSQL share number formatting masks. An MSSQL to PostgreSQL project could still be a problem.

This is very similar to issue 680 (https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/680) but I've raised it as an independent issue because there is no common alternative already coded. Issue 680 can utilise existing `strftime`  functionality. This issue either needs a new engine agnostic alternative to cast or engine specific code to say "if the source and target are Oracle/PostgreSQL then use to_char and build a suitable mask".

nj1973:
Example of PostgreSQL default CAST behaviour (note the unnecessary `.00`):

```
# SELECT CAST(CAST(100 AS DECIMAL(5,2)) AS VARCHAR(10));
 varchar
---------
 100.00
```

Most other engines would return a string of `'100'` from the above operation.

This change overrides the PostgreSQL `cast()` function in `third_party/ibis/ibis_addon/operations.py`. The new code diverts the operation to `rtrim(to_char(),'.')` for `Decimal` columns with a scale > 0. The example about then looks like this:

```
# SELECT RTRIM(TO_CHAR(CAST(100 AS DECIMAL(5,2)),'FM999.99'),'.');
 rtrim
-------
 100
```

### DIFF
--- a/third_party/ibis/ibis_addon/operations.py
+++ b/third_party/ibis/ibis_addon/operations.py
@@ -220,11 +220,12 @@ def sa_cast_postgres(t, expr):
         #     100.00
         # This doesn't match most engines which would return "100".
         # Using to_char() function instead of cast to return a more typical value.
+        # We've wrapped to_char in rtrim(".") due to whole numbers having a trailing ".".
         # Would have liked to use trim_scale but this is only available in PostgreSQL 13+
         #     return (sa.cast(sa.func.trim_scale(sa_arg), sa_type))
         precision = arg.type().precision or 38
         fmt = "FM" + ("9" * (precision - arg.type().scale)) + "." + ("9" * arg.type().scale)
-        return sa.func.to_char(sa_arg, fmt)
+        return sa.func.rtrim(sa.func.to_char(sa_arg, fmt), ".")
 
     if arg.type().equals(dt.binary) and typ.equals(dt.string):
         return sa.func.encode(sa_arg, 'escape')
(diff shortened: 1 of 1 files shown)


nehanene15:
LGTM! This is great.

nehanene15 (approved):

PR #723, opened 2023-02-15
fix: Add support for numeric and precision with length and precision in Postgres Custom Query

### LINKED ISSUE #681: validate custom-query: Exception when PostgreSQL column data type has length or precision
When validating by query between Oracle and PostgreSQL we get an exception if any validated column has a length or precision.

Demonstrating below with a PostgreSQL to PostgreSQL comparison to keep things simple.

Test table:
```
CREATE TABLE dvt_test.qtab (
  id bigint
, col1 numeric(4)
) ;
INSERT INTO dvt_test.qtab VALUES (1, 111);
```

Query file:
```
$ echo "select id,col1 from dvt_test.qtab" > /tmp/q.sql
```

Validation command:
```
$ data-validation validate custom-query --custom-query-type row \
--source-query-file /tmp/q.sql --target-query-file /tmp/q.sql \
-sc pg_conn -tc pg_conn --hash '*' --primary-keys id
...
  File "/usr/local/google/home/neiljohnson/github/professional-services-data-validator/.venv/lib/python3.10/site-packages/ibis/expr/datatypes.py", line 1573, in default
    raise com.IbisTypeError('Value {!r} is not a valid datatype'.format(value))
ibis.common.exceptions.IbisTypeError: Value None is not a valid datatype
```

Column data types are decoded from the query in `third_party/ibis/ibis_postgres/client.py` in `_get_schema_using_query()`.
The data type string, for example `'numeric(4,0)'` is not matched in the list of the accepted types because type mappings do not include the precision/length in parentheses.

sharangagarwal:
Closes #681 

- Added _parse_numeric function from the [new ibis postgres datatype update](https://github.com/ibis-project/ibis/blob/master/ibis/backends/postgres/datatypes.py) and made it compatible with the current version
- added parsy pypi library dependency

### DIFF
--- a/third_party/ibis/ibis_postgres/client.py
+++ b/third_party/ibis/ibis_postgres/client.py
@@ -47,73 +47,73 @@ ORDER BY attnum
     tuples = [(col, self._get_type(typestr)) for col, typestr in type_info]
     return sch.Schema.from_tuples(tuples)
 
-def _get_type(self,typestr: str) -> dt.DataType:
-    _type_mapping = {
-    # "boolean": dt.bool,
-    "boolean": dt.boolean,
-    "boolean[]": dt.Array(dt.boolean),
-    "bytea": dt.binary,
-    "bytea[]": dt.Array(dt.binary),
-    "character(1)": dt.string,
-    "character(1)[]": dt.Array(dt.string),
-    "bigint": dt.int64,
-    "bigint[]": dt.Array(dt.int64),
-    "smallint": dt.int16,
-    "smallint[]": dt.Array(dt.int16),
-    "integer": dt.int32,
-    "integer[]": dt.Array(dt.int32),
-    "text": dt.string,
-    "text[]": dt.Array(dt.string),
-    "json": dt.json,
-    "json[]": dt.Array(dt.json),
-    "point": dt.point,
-    "point[]": dt.Array(dt.point),
-    "polygon": dt.polygon,
-    "polygon[]": dt.Array(dt.polygon),
-    "line": dt.linestring,
-    "line[]": dt.Array(dt.linestring),
-    "real": dt.float32,
-    "real[]": dt.Array(dt.float32),
-    "double precision": dt.float64,
-    "double precision[]": dt.Array(dt.float64),
-    "macaddr8": dt.macaddr,
-    "macaddr8[]": dt.Array(dt.macaddr),
-    "macaddr": dt.macaddr,
-    "macaddr[]": dt.Array(dt.macaddr),
-    "inet": dt.inet,
-    "inet[]": dt.Array(dt.inet),
-    "character": dt.string,
-    "character[]": dt.Array(dt.string),
-    "character varying": dt.string,
-    "character varying[]": dt.Array(dt.string),
-    "date": dt.date,
-    "date[]": dt.Array(dt.date),
-    "time without time zone": dt.time,
-    "time without time zone[]": dt.Array(dt.time),
-    "timestamp without time zone": dt.timestamp,
-    "timestamp without time zone[]": dt.Array(dt.timestamp),
-    "timestamp with time zone": dt.Timestamp("UTC"),
-    "timestamp with time zone[]": dt.Array(dt.Timestamp("UTC")),
-    "interval": dt.interval,
-    "interval[]": dt.Array(dt.interval),
-    # NB: this isn"t correct, but we try not to fail
-    "time with time zone": "time",
-    "numeric": dt.Decimal,
-    "numeric[]": dt.Array(dt.Decimal),
-    "uuid": dt.uuid,
-    "uuid[]": dt.Array(dt.uuid),
-    "jsonb": dt.jsonb,
-    "jsonb[]": dt.Array(dt.jsonb),
-    "geometry": dt.geometry,
-    "geometry[]": dt.Array(dt.geometry),
-    "geography": dt.geography,
-    "geography[]": dt.Array(dt.geography),
-    }
-    _BRACKETS = "[]"
-    is_array = typestr.endswith(_BRACKETS)
-    if (typ := _type_mapping.get(typestr.replace(_BRACKETS, ""))) is not None:
-        return dt.Array(typ) if is_array else typ
-    return _parse_numeric(typestr)
+# def _get_type(self,typestr: str) -> dt.DataType:
+#     _type_mapping = {
+#     # "boolean": dt.bool,
+#     "boolean": dt.boolean,
+#     "boolean[]": dt.Array(dt.boolean),
+#     "bytea": dt.binary,
+#     "bytea[]": dt.Array(dt.binary),
+#     "character(1)": dt.string,
+#     "character(1)[]": dt.Array(dt.string),
+#     "bigint": dt.int64,
+#     "bigint[]": dt.Array(dt.int64),
+#     "smallint": dt.int16,
+#     "smallint[]": dt.Array(dt.int16),
+#     "integer": dt.int32,
+#     "integer[]": dt.Array(dt.int32),
+#     "text": dt.string,
+#     "text[]": dt.Array(dt.string),
+#     "json": dt.json,
+#     "json[]": dt.Array(dt.json),
+#     "point": dt.point,
+#     "point[]": dt.Array(dt.point),
+#     "polygon": dt.polygon,
+#     "polygon[]": dt.Array(dt.polygon),
+#     "line": dt.linestring,
+#     "line[]": dt.Array(dt.linestring),
+#     "real": dt.float32,
+#     "real[]": dt.Array(dt.float32),
+#     "double precision": dt.float64,
+#     "double precision[]": dt.Array(dt.float64),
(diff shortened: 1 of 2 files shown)


third_party/ibis/ibis_postgres/client.py
```diff
@@ -40,72 +49,106 @@ def _get_schema_using_query(self, query: str) -> sch.Schema:
...
+
+COMMA = spaceless_string(",")
+COLON = spaceless_string(":")
+SEMICOLON = spaceless_string(";")
+
+RAW_STRING = parsy.regex(_STRING_REGEX).map(ast.literal_eval)
+FIELD = parsy.regex("[a-zA-Z_][a-zA-Z_0-9]*")
+
+
+def _parse_numeric(
```
nj1973, line 87: This feels over-engineered to me. I looked at how the Impala client does this and it uses `dt.dtype` like this:
```
        elif "decimal" in t:
            result = dt.dtype(t)
```

If we followed that approach for PostgreSQL your new `_get_type` function could look like this:
```
def _get_type(self, typestr: str) -> dt.DataType:
    typestr = typestr.lower()
    is_array = typestr.endswith(_BRACKETS)
    typ = _type_mapping.get(typestr.replace(_BRACKETS, ""))
    if typ is not None:
        return dt.Array(typ) if is_array else typ
    if typestr.startswith("numeric"):
        # numeric is a synonym for decimal
        typestr = typestr.replace("numeric", "decimal")
    if typestr.lower().startswith("decimal"):
        return dt.dtype(typestr)
    return _type_mapping[typestr]
```

The code is bit more primitive, no AST etc, but it means you don't need `parsy`,  `_parse_numeric` or any of the new constants/imports.

I think the intention would be a lot clearer.

Did I miss something?

third_party/ibis/ibis_postgres/client.py
```diff
@@ -13,11 +13,20 @@
...
+
+from __future__ import annotations
+import parsy
+import re
+import ast
+import toolz
 import ibis.expr.datatypes as dt
 import ibis.expr.schema as sch
+
+# import ibis.expr.types as ir
```
nj1973: Is this to be removed?

nj1973 (changes requested):
Have you tested your code with non-numeric columns? I tried it with a date and got an error:
```
  File "/usr/local/google/home/neiljohnson/github/professional-services-data-validator2/third_party/ibis/ibis_postgres/client.py", line 107, in _parse_numeric
    return ty.parse(text)
  File "/usr/local/google/home/neiljohnson/github/professional-services-data-validator2/.venv/lib/python3.10/site-packages/parsy/__init__.py", line 87, in parse
    (result, _) = (self << eof).parse_partial(stream)
  File "/usr/local/google/home/neiljohnson/github/professional-services-data-validator2/.venv/lib/python3.10/site-packages/parsy/__init__.py", line 101, in parse_partial
    raise ParseError(result.expected, stream, result.furthest)
parsy.ParseError: expected one of 'decimal', 'numeric' at 0:0
```

I think you should add a unit test feeding each common data type string into the `_get_type`  function to be sure we don't get any exceptions.

third_party/ibis/ibis_postgres/client.py
sharangagarwal: Can you provide the table/query you used?

We have[ integration test](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/tests/system/data_sources/test_postgres.py) written already for major datatypes and it works for them. I have checked it as well.

**table creation:**
DROP TABLE IF EXISTS test_data_types_postgres_row;
CREATE TABLE test_data_types_postgres_row(
	serial_col serial PRIMARY KEY,
    int_col int,
	text_col  text,
    char_col  char,
    varchar_col  varchar,
	float_col double precision,
	numeric_col  numeric(5, 2),
	timestamp_col timestamp,
    date_col date
);
INSERT INTO test_data_types_postgres_row(int_col, text_col, char_col, varchar_col, float_col, numeric_col, timestamp_col, date_col) values (10, 'row1_text', '1', 'row1_varchar', 10.10, 10.101, '2016-06-22 19:10:25-07','2020-01-01');
INSERT INTO test_data_types_postgres_row(int_col, text_col, char_col, varchar_col, float_col, numeric_col, timestamp_col, date_col) values (20, 'row2_text', '2', 'row2_varchar', 20.20, 20.202, '2016-06-22 19:20:25-07','2020-02-02');
INSERT INTO test_data_types_postgres_row(int_col, text_col, char_col, varchar_col, float_col, numeric_col, timestamp_col, date_col) values (30, 'row3_text', '3', 'row3_varchar', 30.20, 30.203, '2016-06-22 19:30:25-07','2020-03-03');
INSERT INTO test_data_types_postgres_row(int_col, text_col, char_col, varchar_col, float_col, numeric_col, timestamp_col, date_col) values (40, 'row4_text', '4', 'row4_varchar', 40.20, 40.404, '2016-06-22 19:30:45-07','2020-04-04');
INSERT INTO test_data_types_postgres_row(int_col, text_col, char_col, varchar_col, float_col, numeric_col, timestamp_col, date_col) values (50, 'row5_text', '5', 'row5_varchar', 50.20, 50.404, '2016-06-22 19:50:44-0','2020-05-05');
INSERT INTO test_data_types_postgres_row(int_col, text_col, char_col, varchar_col, float_col, numeric_col, timestamp_col, date_col) values (60, 'row6_text', '6', 'row6_varchar', 60.20, 60.404, '2016-01-22 19:51:44-0','2020-06-06');
INSERT INTO test_data_types_postgres_row(int_col, text_col, char_col, varchar_col, float_col, numeric_col, timestamp_col, date_col) values (70, 'row7_text', '7', 'row7_varchar', 70.20, 70.404, '2017-01-22 19:51:44-0','2020-06-07');
INSERT INTO test_data_types_postgres_row(int_col, text_col, char_col, varchar_col, float_col, numeric_col, timestamp_col, date_col) values (80, 'row8_text', '8', 'row8_varchar', 80.20, 80.404, '2018-01-22 19:51:44-0','2020-06-08');
INSERT INTO test_data_types_postgres_row(int_col, text_col, char_col, varchar_col, float_col, numeric_col, timestamp_col, date_col) values (90, 'row9_text', '9', 'row9_varchar', 90.20, 90.404, '2019-01-22 19:51:44-0','2020-06-09');
INSERT INTO test_data_types_postgres_row(int_col, text_col, char_col, varchar_col, float_col, numeric_col, timestamp_col, date_col) values (100, 'row10_text', '1', 'row10_varchar', 100.20, 100.404, '2020-01-22 19:51:44-0','2021-06-09');

**dvt execution**
data-validation validate custom-query --custom-query-type row --source-query-file source_query.sql --target-query-file source_query.sql -sc MY_PG_CONN -tc MY_PG_CONN --hash '*' --primary-keys serial_col

sharangagarwal: Also, I picked this code almost entirely from the [postgres new ibis framework code](https://github.com/ibis-project/ibis/blob/master/ibis/backends/postgres/datatypes.py) and took _parse_numeric and _get_type because of that. I do think we should go with the original implementation since changing the code would make it harder to clean up once we migrate to the new framework later this year. (**At that point third party folder for postgres would simply go away**)

third_party/ibis/ibis_postgres/client.py
sharangagarwal: Yes, I will clean this up. I added it during development but ir wasn't needed in the final edit.

third_party/ibis/ibis_postgres/client.py
nj1973: Ok, fair enough regarding the code.

nj1973 (changes requested):
The test column triggering the exception is actually timestamp rather than date.
```
col_ts          timestamp(6)
```

And looks like this from a describe:
```
    Column    |              Type              | Collation | Nullable | Default
--------------+--------------------------------+-----------+----------+---------
 id           | integer                        |           | not null |
...
 col_ts       | timestamp(6) without time zone |           |          |
```

third_party/ibis/ibis_postgres/client.py
```diff
@@ -40,72 +45,116 @@ def _get_schema_using_query(self, query: str) -> sch.Schema:
...
+    )
+
+    ty = pg_array | decimal
+    return ty.parse(text)
+
+
+def _get_type(self, typestr: str) -> dt.DataType:
+    is_array = typestr.endswith(_BRACKETS)
+    # typ = _type_mapping.get(typestr.replace(_BRACKETS, ""))
+    # handle bracket length
```
sharangagarwal, line 109: @nj1973 
I have added this code piece to handle fixed length brackets being passed on with the typestr

Example: The typestr comes as "timestamp(6) without time zone" instead of "timestamp without time zone" when a length is specified, so while passing the key, I am removing the (6) part and and mapping to dt.timestamp

Also I checked this would also work for character( ... ) and others since its generic fix

nj1973: Great, I can confirm my test case works too.

nj1973 (approved):
LGTM

PR #726, opened 2023-02-16
fix: Protect non-date columns from astype(str) date workaround

### LINKED ISSUE #684: validate row: Incorrect end report for PostgreSQL target when ID column is decimal
Ran into this on an Oracle to PostgreSQL project.
Test Oracle table:
```
create table dvt_test.tab_dec_id (
  id number(2) not null
, col1 number(5));
ALTER TABLE dvt_test.tab_dec_id ADD PRIMARY KEY (id);
insert into dvt_test.tab_dec_id values (1,1);
commit;
```

Test PostgreSQL table:
```
create table dvt_test.tab_dec_id (
  id decimal(2) not null
, col1 decimal(5)
);
ALTER TABLE dvt_test.tab_dec_id ADD PRIMARY KEY (id);
insert into dvt_test.tab_dec_id values (1,1);
```

Validation command below:
```
data-validation validate row -sc ora_conn -tc pg_conn \
-tbls=dvt_test.tab_dec_id=dvt_test.tab_dec_id --primary-keys id --concat '*' \
--filter-status fail
```

And we get the following report (I've removed some white space to try and make it more readable):
```
╒═════════════╤════╤═════════════════════╤════════════╤═══════════╤════════════╤════════════════╤══════╕
│ validation_ │type│ source_table_name   │ source_col │source_val │ target_val │ pct_difference │status│
╞═════════════╪════╪═════════════════════╪════════════╪═══════════╪════════════╪════════════════╪══════╡
│ concat__all │Row │ dvt_test.tab_dec_id │ concat__all│        11 │        nan │            nan │fail  │
├─────────────┼────┼─────────────────────┼────────────┼───────────┼────────────┼────────────────┼──────┤
│ concat__all │Row │ nan                 │ nan        │       nan │         11 │            nan │fail  │
╘═════════════╧════╧═════════════════════╧════════════╧═══════════╧════════════╧════════════════╧══════╛
```

If I switch the PostgreSQL table to use an integer for the ID column:
```
drop table dvt_test.tab_dec_id;
create table dvt_test.tab_dec_id (
  id bigint not null
, col1 decimal(5)
);
ALTER TABLE dvt_test.tab_dec_id ADD PRIMARY KEY (id);
insert into dvt_test.tab_dec_id values (1,1);
```

The report then shows no failures:
```
╒═════════════╤════╤═════════════════════╤════════════╤═══════════╤════════════╤════════════════╤══════╕
│ validation_ │type│ source_table_name   │ source_col │source_val │ target_val │ pct_difference │status│
╞═════════════╪════╪═════════════════════╪════════════╪═══════════╪════════════╪════════════════╪══════╡
╘═════════════╧════╧═════════════════════╧════════════╧═══════════╧════════════╧════════════════╧══════╛
```

I've had a look at the objects being compared and the problem appears to be that the join value is `1` for the source table expression but `1.0` for the target and these fail to match. You can see from the actual comparison row value in the incorrect output that `source_val` and `target_val` do actually match each other.

nj1973:
There is a workaround related to joining source and target data frames on date based columns. The workaround turns join columns in both source and target data frames to strings before combining them to create a final report.

Unfortunately this workaround was also converting numeric columns to string which was causing problems with trailing zeros after the decimal place. A join on the numeric would ignore these, a join on strings of `1` and `1.0` was a mismatch.

This change modifies the existing workaround to only be active when the join column in the source schema is date based.

### DIFF
--- a/data_validation/data_validation.py
+++ b/data_validation/data_validation.py
@@ -18,6 +18,7 @@ import warnings
 from concurrent.futures import ThreadPoolExecutor
 
 import ibis.backends.pandas
+import ibis.expr.datatypes as dt
 import numpy
 import pandas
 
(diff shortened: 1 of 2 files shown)


nehanene15 (approved):
LGTM!

PR #746, opened 2023-02-22
docs: oracleuserpriv

slatawa:
docs: add instructions on privileges required by oracle user to connect to DVT

### DIFF
--- a/docs/connections.md
+++ b/docs/connections.md
@@ -155,6 +155,11 @@ Then `pip install cx_Oracle`.
 }
 ```
 
+### Oracle User permissions to run the validator tool:
+* GRANT CREATE SESSION TO DVT-USER (Create Session Privilege)
+* GRANT READ ON SCHEMA_NAME.TABLE_NAME to DVT-USER; (Oracle Read table privilege)
+
+
 ## MSSQL Server
 Please note the MSSQL Server package is not installed by default. You will need to follow [SQL Server](https://docs.microsoft.com/en-us/sql/connect/odbc/linux-mac/installing-the-microsoft-odbc-driver-for-sql-server) installation steps.
 Then `pip install pyodbc`.
--- a/docs/connections.md
+++ b/docs/connections.md
@@ -158,6 +158,7 @@ Then `pip install cx_Oracle`.
 ### Oracle User permissions to run the validator tool:
 * CREATE SESSION
 * READ or SELECT on any tables to be validated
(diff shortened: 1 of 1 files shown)


docs/connections.md
```diff
@@ -155,6 +155,11 @@ Then `pip install cx_Oracle`.
 }
 ```
 
+### Oracle User permissions to run the validator tool:
+* GRANT CREATE SESSION TO DVT-USER (Create Session Privilege)
```
nj1973: I wonder if it would be better to be consistent with the BigQuery section and simply list the privileges we require?
For example:

### Oracle user permissions required to run this validator tool:
* CREATE SESSION
* READ or SELECT on any tables to be validated

Just a suggestion, not an essential change (I just like consistency :-) )

docs/connections.md
```diff
@@ -155,6 +155,11 @@ Then `pip install cx_Oracle`.
...
+* GRANT READ ON SCHEMA_NAME.TABLE_NAME to DVT-USER; (Oracle Read table privilege)
+
+
```
nj1973, line 162: I was recently pointed to an issue logged in the Oracle alert log that is caused by SQL Alchemy when the user doesn't have read privileges on V$TRANSACTION. I wonder if we should cover this too. For example:

If you encounter "WARNING: too many parse errors" messages in your Oracle alert log relating to a query from your DVT database account then a workaround is to grant READ privilege on SYS.V_$TRANSACTION. See here for background on this: https://docs.sqlalchemy.org/en/14/dialects/oracle.html#transaction-isolation-level-autocommit


Or do you think this is overkill?

docs/connections.md
slatawa: @nj1973  - Incorporated.

docs/connections.md
slatawa: @nj1973  - Looking at the issue you shared ,I concur it should be a pre-requisite to have read on SYS.V_$TRANSACTION but don't see any reason to only give this privilege if the users sees the warning. from the doc 

"The cx_Oracle dialect attempts to call the [Connection.get_isolation_level()](https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Connection.get_isolation_level) method when the dialect makes its first connection to the database in order to acquire the “default”isolation level."

so it seems to me that we should get a read priv on SYS.V_$TRANSACTION instead of the alternative behaviour which would be getting a background warning and setting to default.

nj1973: OK, that makes sense.
Because v$transaction is an unusual privilege to require I think we should include a comment or link to explain why we expect it. Also stating that it is optional (because code continues to function fine without it).

slatawa: @nj1973 - have added the same to doc now.

nj1973 (approved):
LGTM

PR #747, opened 2023-02-22
fix: issue 740 teradata strftime function

### LINKED ISSUE #740: Teradata invalid strftime function - DATE colums failing
The Teradata [implementation of strftime](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/third_party/ibis/ibis_teradata/compiler.py#L632) compiles FORMAT_DATE() which is an invalid function. 

This needs to be changed to a CAST(date_column AS DATE FORMAT 'DD-MM-YYYY')

nj1973:
Add strftime implementation for Teradata and also add an integration test.

### DIFF
--- a/tests/system/data_sources/test_teradata.py
+++ b/tests/system/data_sources/test_teradata.py
@@ -30,7 +30,7 @@ conn = {
 }
 
 
-TERADATA_CONFIG = {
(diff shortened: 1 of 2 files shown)


nehanene15 (approved):
LGTM- thanks!

PR #749, opened 2023-02-27
fix: Protect column and row validation calculated column names from Oracle 30 character identifier limit

### LINKED ISSUE #724: validate row: ORA-00972: identifier is too long error on Oracle 11g
Prior to version 12.2 Oracle has an identifier length limit of 30 characters.

When running a row validation columns in the SQL projection are aliased with names reflecting the functions that have been applied. Example:

```
data-validation validate row -sc ora_conn -tc bq_conn -tbls=dvt_test.tab_dec_id=dvt_test.tab_dec_id \
--primary-keys id --concat 'col1'
...
sqlalchemy.exc.DatabaseError: (cx_Oracle.DatabaseError) ORA-00972: identifier is too long
[SQL: SELECT t0.concat__all, t0.id
FROM (SELECT t1.id AS id, t1.col1 AS col1, t1.cast__col1 AS cast__col1, t1.ifnull__cast__col1 AS ifnull__cast__col1, t1.rstrip__ifnull__cast__col1 AS rstrip__ifnull__cast__col1, t1.upper__rstrip__ifnull__cast__col1 AS upper__rstrip__ifnull__cast__col1, upper__rstrip__ifnull__cast__col1 AS concat__all
FROM (SELECT t2.id AS id, t2.col1 AS col1, t2.cast__col1 AS cast__col1, t2.ifnull__cast__col1 AS ifnull__cast__col1, t2.rstrip__ifnull__cast__col1 AS rstrip__ifnull__cast__col1, upper(t2.rstrip__ifnull__cast__col1) AS upper__rstrip__ifnull__cast__col1
FROM (SELECT t3.id AS id, t3.col1 AS col1, t3.cast__col1 AS cast__col1, t3.ifnull__cast__col1 AS ifnull__cast__col1, rtrim(t3.ifnull__cast__col1) AS rstrip__ifnull__cast__col1
FROM (SELECT t4.id AS id, t4.col1 AS col1, t4.cast__col1 AS cast__col1, coalesce(t4.cast__col1, :param_1) AS ifnull__cast__col1
FROM (SELECT t5.id AS id, t5.col1 AS col1, CAST(t5.col1 AS VARCHAR2(4000 CHAR)) AS cast__col1
FROM dvt_test.tab_dec_id t5) t4) t3) t2) t1) t0]
[parameters: {'param_1': 'DEFAULT_REPLACEMENT_STRING'}]
```

The identifier length has exceeded 30:
```
$ echo "upper__rstrip__ifnull__cast__col1" | awk '{print length($1)}'
33
```

Most other engines (including Oracle 12.2+) have a limit of 128 characters so we are less likely to run into this, although in theory they are all vulnerable.

nj1973:
This protects all engines not just Oracle 12.1 and below. It's just that Oracle 12.1 and below have a 30 character limit which we almost always cross for row validation. Other engines have a 128 character identifier limit and therefore are less likely to be breached.

This PR introduces the following change:

1. Identify the maximum identifier length for both source and target engines and take the lowest one as the ceiling for the validation.
2. When prefixing a column alias with an operation, such as sum__, ifnull__ etc, we compare the length of the resulting name with the maximum length.
3. If the new name breaches the maximum then a simplified alias is used. The simplified alias uses the column position in place of its name.
4. This has been done for both concat/hash row validation and aggregate column validation.
5. I've added unit tests for the name shortening logic.
6. Ideally we could do with adding integration tests for Oracle 11g, I think this can happen outside of this issue.

Example of column validation SQL from new logic:
```
SELECT count(:count_1) AS count
, max(t0.id) AS max__id
, max(t0.col_dec_long_123456789012345) AS max__dvt_calc_col_1
, max(t0.length__dvt_calc_col_4) AS max__length__dvt_calc_col_4
FROM (
  SELECT t1.id AS id
  , t1.col_dec_long_123456789012345 AS col_dec_long_123456789012345
  , t1.col_dec2_long_123456789012345 AS col_dec2_long_123456789012345
  , t1.col_date_long_1234567890 AS col_date_long_1234567890
  , t1.col_string_long_1234567890 AS col_string_long_1234567890
  , length(t1.col_string_long_1234567890) AS length__dvt_calc_col_4
  FROM dvt_test.tab_long_cols t1
) t0
```

Note that:

- `length__dvt_calc_col_4` is introduced at depth 1 because the length of `length__col_string_long_1234567890` would cross 30 characters
- The `max__` alias is then applied to the length column at depth 0: `max__length__dvt_calc_col_4`
- The `max__` alias is applied to the decimal column at depth 0 because the length of `max__col_dec_long_123456789012345` would cross 30 characters
- `max__id` is unaffected because it remains below 30 characters

The same logic would be seen in a row validation too but more functions are applied so we're more likely to see `dvt_calc_col` columns. It's worth pointing out that this is only visible in the SQL statement, the end report still uses proper column names.

### DIFF
--- a/data_validation/config_manager.py
+++ b/data_validation/config_manager.py
@@ -743,31 +743,13 @@ class ConfigManager(object):
 
         return calculated_config
 
-    @staticmethod
-    def _is_oracle_client(client):
-        try:
-            return isinstance(client, clients.OracleClient)
-        except TypeError:
-            # When no Oracle client installed clients.OracleClient is not a class
-            return False
-
     def _get_comparison_max_col_length(self) -> int:
-        def max_length(client):
-            if self._is_oracle_client(client):
-                # We can't reliably know which Version class client.version is stored in
-                # because it is out of our control. Therefore using string identification
-                # of Oracle <= 12.1 to avoid exceptions of this nature:
-                #  TypeError: '<' not supported between instances of 'Version' and 'Version'
-                if (
-                    str(client.version)[:2] in ["10", "11"]
-                    or str(client.version)[:4] == "12.1"
-                ):
-                    return 30
-            return 128
-
         if not self._comparison_max_col_length:
             self._comparison_max_col_length = min(
-                [max_length(self.source_client), max_length(self.target_client)]
+                [
+                    clients.get_max_column_length(self.source_client),
+                    clients.get_max_column_length(self.target_client),
+                ]
             )
         return self._comparison_max_col_length
 
@@ -776,7 +758,7 @@ class ConfigManager(object):
     ) -> str:
         if isinstance(column_type, dt.Timestamp):
             return "%Y-%m-%d %H:%M:%S"
-        if self._is_oracle_client(client):
+        if clients.is_oracle_client(client):
             # Oracle DATE is a DateTime
             return "%Y-%m-%d %H:%M:%S"
         return "%Y-%m-%d"
--- a/data_validation/config_manager.py
+++ b/data_validation/config_manager.py
@@ -527,23 +527,41 @@ class ConfigManager(object):
 
         return aggregate_config
 
+    def _prefix_calc_col_name(
+        self, column_name: str, prefix: str, column_number: int
+    ) -> str:
+        """Prefix a column name but protect final string from overflowing SQL engine identifier length limit."""
+        new_name = f"{prefix}__{column_name}"
+        if len(new_name) > self._get_comparison_max_col_length():
+            # Use an abstract name for the calculated column to avoid composing invalid SQL.
+            new_name = f"{prefix}__dvt_calc_col_{column_number}"
+        return new_name
+
     def build_and_append_pre_agg_calc_config(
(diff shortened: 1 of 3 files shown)


data_validation/config_manager.py
```diff
@@ -712,19 +743,40 @@ def build_config_calculated_fields(
 
         return calculated_config
 
+    @staticmethod
+    def _is_oracle_client(client):
```
dhercher: A max size per database could live in clients.py?  Something like
```
def get_max_column_length(client):
  if ..
  else:
      return 128
```

nj1973: Thanks, that's a good idea. I initially wanted to add a method to each client but I was nervous of making such a broad change so I added in `config_manager.py`. I'll move the logic to `client.py`.

nj1973: I've made that change.

nehanene15 (approved):
LGTM!

dhercher (approved):

PR #766, opened 2023-03-10
test: Core data types integration tests

### LINKED ISSUE #759: test: Add core data types test table and tests
Add a test table that covers off all typical data types for each SQL engine. Craft the table in such a way that column names and contents are the same across all engines, this opens up the opportunity to add cross engine integration tests in the future.

Don't put any challenging types/values into the table. Specific problem types or values should have their own test tables/tests added in the future.

Add integration tests comparing each engine's table back to itself. This proves we have support for all core data types. Once this is complete raise a second issue to add cross-engine tests.

Add tests for:

- validate schema
- validate column
- validate row

For any issues we discover while running the new tests above we should:

- Create an issue describing the problem and which test should be enabled to reproduce it.
- Disable the test on this feature branch so it can be enabled again as part of the related issue.

nj1973:
Added a test table that covers off all typical data types for each SQL engine and added tests for `validate schema`, `validate column`, `validate row` for each engine.

Notes:

- Crafted the table in such a way that column names and contents are the same across all engines, this opens up the opportunity to add cross engine integration tests in the future. BUT, for now, all new integration tests compare the engine back to itself.
- Don't put any challenging types/values into the table. Specific problem types or values should have their own test tables/tests added in the future.
- Add integration tests comparing each engine's table back to itself. This proves we have support for all core data types. Once this is complete raise a second issue to add cross-engine tests.
- Issues found while testing these changes have been raised in github and the issue numbers referenced where I've had to disable tests.
- As far as I can tell the integration test tables are manually created. Therefore I have manually created each table and put the DDL I used in `tests/resources/` where there was already a `bigquery_test_tables.sql` file. Spanner is the exception to this, Spanner tests create instances/tables automatically. I've modified the files below to include the new test table:
  - third_party/ibis/ibis_cloud_spanner/tests/ddl.sql
  - third_party/ibis/ibis_cloud_spanner/tests/dml.sql
- I've copied the Oracle ITEMS_PRICE table from the XE database which is a container database and not for storing application tables. It is now in XEPDB in the pso_data_validator schema along with the new table.
- The new integration tests are driven from `argparse` options, this is unlike other integration tests which are driven from hardcoded config. I've done this because I want to test that config generation can also interrogate data types correctly.
- In most cases I've put the new table in a pso_data_validator container (schema/database) except for the cases below:
  - Teradata: I've have put the new table in the udf database because I don't have privileges to create a new database.
  - Spanner: The new table is in the instance/database generated afresh for each run.
- Hive tests are really slow. There was originally 1 test, there are now 4. If I take any of the 4 and run it in isolation then it completes in around 20 seconds. If run all 4 tests back to back then we get an elapsed time in excess of 6 minutes. Not ideal but I believe related to infrastructure rather than code changes.
- Before these changes the MySQL integration tests were not running. CLoud SQL Proxy was not being used and the tests were silently failing in caught `exceptions.DataClientConnectionFailure` exceptions. I've changed so the tests do run, but then a couple of them failed for other reasons so I've raised issues and disabled them.

Upon merging these changes we should create issues to add cross-engine integration tests. I think the new cross-engine tests will uncover many more problems and generate more issues.


nj1973:
I propose we leave the Teradata table in the UDF database for now so we can proceed with review and (hopefully) merge. I can raise another issue to move the tables in future.

tests/system/data_sources/test_postgres.py
```diff
@@ -460,3 +463,87 @@ def test_schema_validation(cloud_sql):
...
+    assert len(df) == 0
+
+
+@mock.patch(
+    "data_validation.state_manager.StateManager.get_connection_config",
+    return_value=CONN,
+)
+def test_column_validation_core_types(mock_conn):
+    parser = cli_tools.configure_arg_parser()
+    # TODO Add col_datetime,col_tstz to --sum string below when issue-XXX is complete. Or change whole string to * if YYY is also complete.
```
nehanene15: Was an issue opened for these Postgres bugs?

tests/system/data_sources/test_mysql.py
```diff
@@ -13,20 +13,24 @@
...
 
 import os
+from unittest import mock
 
-from data_validation import data_validation, consts, exceptions
+from data_validation import __main__ as main
+from data_validation import cli_tools, data_validation, consts, exceptions
 
 
 # TODO: To use this code I would need to use the Cloud SQL Proxy.
```
nehanene15: nit: We can delete this TODO

nehanene15 (commented):
LGTM besides documenting the issue numbers for the Postgres test!

tests/system/data_sources/test_postgres.py
nj1973: No, I've made a mistake there, PostgreSQL tests can run with these set to `*`. Thanks for spotting this.
I've modified the code and tests re-running.

tests/system/data_sources/test_mysql.py
nj1973: I've removed the TODO.
I'd not seen "nit" shorthand before - nice :-)

nehanene15 (approved):

nehanene15:
Looks great! Thank you for working on this much needed issue!

Issue #743, opened 2023-02-19
Impala DVT connection does not allow use-ssl parameter that is allowed in impyla under the hood.


wt1309:
The use-ssl parameter is not included in the DVT connection parameters, however, the Impala connection I am working with today is kerberized and requires SSL flag to be true. I performed the following to get this to work. Plan to eventually submit a pull request once I get some time, but for now, here is the solution: 

1. Missing python library: sudo pip3 install kerberos
2. Used port 21050 instead of 21000 (21000 works for CLI impala shell)
3. Update DVT source code to accomodate new parameter use_ssl

- 	Add following lines of code after Line 35, 42,and 48 respectively:

	professional-services-data-validator/third_party/ibis/ibis_impala/api.py

	after line 35: use_ssl=False
	after line 42: use_ssl = (use_ssl, False)[use_ssl is None]
	after line 48: use_ssl=use_ssl,

- 	Add the following line of code after Line 131 in:

	professional-services-data-validator/data_validation/cli_tools.py 

	after line 131: ["use_ssl", "Use SSL when connecting to HiveServer2 (default is False)"],

4. cd to professional-services-data-validator folder
5. Then run command to build new code: sudo python3 -m pip install . 

Now this connection should work: 
data-validation connections add --connection-name <Conneciton_Name> Impala --host <hostname> --auth-mechanism GSSAPI --kerberos-service-name impala --port 21050 **--use-ssl True**

mloeberc:
closed by PR #790

PR #790, opened 2023-03-31
feat: add Impala connection optional parameters (#743)

mloeberc:
Closes issue #743 

Add optional parameters supported by Impala Connector, to give support in DVT:
Parameters added:
- use_ssl
- timeout
- ca_cert
- user
- password
- pool_size
- hdfs_client

Attaching result for execution of the help command for the connection:
Screenshot from 2023-03-31 00-15-13

### DIFF
--- a/third_party/ibis/ibis_impala/api.py
+++ b/third_party/ibis/ibis_impala/api.py
@@ -33,6 +33,13 @@ def impala_connect(
     database="default",
     auth_mechanism="PLAIN",
     kerberos_service_name="impala",
+    use_ssl=False,
+    timeout=45,
+    ca_cert=None,
+    user=None,
+    password=None,
+    pool_size=8,
+    hdfs_client=None
 ):
     auth_mechanism = (auth_mechanism, "PLAIN")[auth_mechanism is None]
     database = (database, "default")[database is None]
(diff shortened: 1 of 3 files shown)


third_party/ibis/ibis_impala/api.py
```diff
@@ -33,19 +33,37 @@ def impala_connect(
...
 ):
     auth_mechanism = (auth_mechanism, "PLAIN")[auth_mechanism is None]
     database = (database, "default")[database is None]
     port = (port, 10000)[port is None]
     kerberos_service_name = (kerberos_service_name, "impala")[
         kerberos_service_name is None
     ]
+    use_ssl = (use_ssl, False)[use_ssl is None]
+    timeout = (timeout, 45)[timeout is None]
+    pool_size = (pool_size, 8)[pool_size is None]
```
nehanene15, line 52: Do we need to add user, password, ca_cert, and the other variables here too? Or will it be picked up from the default parameters?

mloeberc: I took as a reference how the code was already handling "host" parameters. Since the default value is None then it is not added in this section; for the other variables the default values are boolean or Int, and that is why it validates if the value is None to assign the default value. If it requires a change just let me know.

nehanene15 (approved):
LGTM!

PR #796, opened 2023-04-06
fix: schema validation fixes for Oracle/SQL Server float64 and SQL Server datetimeoffset

### LINKED ISSUE #764: validate schema: SQL Server DATETIMEOFFSET Throws Exception
Test table:
```
CREATE TABLE dvt_test.tab_tstz
(   id              int NOT NULL PRIMARY KEY
,   col_tstz        datetimeoffset(3)
);

INSERT INTO dvt_test.tab_tstz VALUES
(1,cast('1970-01-01 00:00:01 -01:00' as datetimeoffset(3));
```

Test command:
```
data-validation validate schema \
-sc sqlserver_conn -tc sqlserver_conn \
-tbls=dvt_test.tab_tstz
```

Exception:
```
NotImplementedError: Could not find signature for dtype: <MSDialect_pyodbc, DATETIMEOFFSET>
```

When complete we need to enable the integration test for this problem. Search for "issue-764" in `tests/system/data_sources/test_sql_server.py` to action this. Enabling this test is blocked by issue https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/759.

### LINKED ISSUE #783: validate schema: Oracle and SQL Server float64 identified by DVT as float32
Test table:
```
CREATE TABLE dvt_test.doubles
(   id              NUMBER(5) NOT NULL PRIMARY KEY
,   col_double      BINARY_DOUBLE
);
```

Test command:
```
data-validation validate schema \
-sc=ora_conn -tc=pg_conn \
--filter-status=fail \
--allow-list="decimal(5,0):int32[non-nullable]" \
-tbls=dvt_test.doubles
```

The Oracle column is identified as `float32`
The PostgreSQL column is (correctly) identified as `float64`

SQL Server has the same problem.

Thoughts on tests:

- I wonder if we can add unit tests for this. Mocking up the types returned from each engine and asserting they are correctly identified. I haven't looked into this so don't know if it is feasible.
- We could consider modifying schema validation integration tests. At the moment they compare a source system back to itself and assert no failures were identified. This proves we can interact with each column but does not prove it has the correct data type. We could remove the fail filter, check all status values are pass and check that the data type for col_float64 is the correct value.
- Or we could wait until we add cross engine integration tests which would highlight this problem.

nj1973:
This PR fixes two issues:

- https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/783
- https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/764

I included issue 764 in the same changes because it was preventing me from enabling unit tests for the issue 783 changes.

The fixes are:

- Oracle BINARY_FLOAT is identified as Float64 (not Float32)
- SQL Server FLOAT is identified as Float64 (not Float32)
- SQL Server DATETIMEOFFSET is identified as Timestamp(UTC) (previously threw an exception)

I couldn't think of a way to add unit tests for these changes because it seemed to me like it needs a SQL engine connection to test. Therefore I've relied upon integration tests.

### DIFF
--- a/tests/system/data_sources/test_sql_server.py
+++ b/tests/system/data_sources/test_sql_server.py
@@ -211,8 +211,6 @@ def test_schema_validation_core_types(mock_conn):
     assert len(config_managers) == 1
     config_manager = config_managers[0]
     validator = data_validation.DataValidation(config_manager.config, verbose=False)
-    # TODO When issue-764 is complete remove the return statement below.
-    return
     df = validator.execute()
     # With filter on failures the data frame should be empty
     assert len(df) == 0
--- a/third_party/ibis/ibis_mssql/client.py
+++ b/third_party/ibis/ibis_mssql/client.py
@@ -29,6 +29,7 @@ from third_party.ibis.ibis_mssql.compiler import MSSQLDialect
 
 import pyodbc  # NOQA fail early if the driver is missing
 
+
 @dt.dtype.register(MSDialect_pyodbc, sa.dialects.mssql.UNIQUEIDENTIFIER)
 def sa_string(_, satype, nullable=True):
     return dt.String(nullable=nullable)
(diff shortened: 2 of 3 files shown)


nehanene15 (approved):
LGTM!

Issue #792, opened 2023-04-05
data-validation connections list output is limited.


mahtodeepak05:
Folks,

Currently _data-validation connections list_ is limited to only connection name.

```
r$ data-validation connections list
04/05/2023 12:00:18 PM-INFO: Connection Name: ora
04/05/2023 12:00:18 PM-INFO: Connection Name: pg
04/05/2023 12:00:18 PM-INFO: Connection Name: src_con
```

its understandable that we can't display all details but If we can include additional details like _"source_type"_, it would be really helpful to differentiate based on source type.

Proposed changes.
```
r$ data-validation connections list
04/05/2023 12:00:18 PM-INFO: Connection Name: ora, source_type : Oracle
04/05/2023 12:00:18 PM-INFO: Connection Name: pg, source_type : Postgres
04/05/2023 12:00:18 PM-INFO: Connection Name: src_con, source_type : Oracle
```

PR #798, opened 2023-04-06
 data-validation connections list output is limited. #792 

jlonge4:
Ran the nox script and passed blacken and lint, failed unit_small due to not having api_keys, however I believe this should fix this issue.


jlonge4:
CLA has been signed

helensilva14:
@jlonge4 closing this PR because we already have this change merged on develop (#803), but thanks a lot for your contribution! We look forward to see you more here :)

PR #800, opened 2023-04-11
feat: add db2 hash and concat support

### LINKED ISSUE #799: Incorrect query generation for cast integer to string for DB2 row validation.
There is incorrect query generation while doing row validation for DB. When casting to string we are casting to CLOB and query is: 
```
SELECT t6.guestname AS guestname, t6.content AS content, t6.entryid AS entryid, t6.guestname AS cast__guestname, t6.content AS cast__content, CAST(t6.entryid AS CLOB) AS cast__entryid 
FROM db2inst1.entries AS t6
```

But we are facing an error that: 
`A value with data type "SYSIBM.INTEGER" cannot be CAST to type "SYSIBM.CLOB".`

I think instead of casting to CLOB casting to VARCHAR(n) will be a solution. 

Some samples from DB2 terminal: 
image
image

ngdav:
Add db2 row hash support and db2 row concat support.
Fix issue where primary keys are being translated to CLOB which is not supported. PKs (and all other ibis.String types) now translate to VARCHAR(3000).

### DIFF
--- a/third_party/ibis/ibis_addon/operations.py
+++ b/third_party/ibis/ibis_addon/operations.py
@@ -47,6 +47,8 @@ from third_party.ibis.ibis_oracle.compiler import OracleExprTranslator
 from third_party.ibis.ibis_teradata.compiler import TeradataExprTranslator
 from third_party.ibis.ibis_mssql.compiler import MSSQLExprTranslator
 from ibis.backends.postgres.compiler import PostgreSQLExprTranslator
+from third_party.ibis.ibis_DB2.compiler import DB2ExprTranslator
+
 
 # from third_party.ibis.ibis_snowflake.compiler import SnowflakeExprTranslator
 # from third_party.ibis.ibis_oracle.compiler import OracleExprTranslator <<<<<< DB2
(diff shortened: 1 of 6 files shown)


third_party/ibis/ibis_addon/operations.py
```diff
@@ -213,6 +220,12 @@ def sa_format_hashbytes_oracle(translator, expr):
     hash_func = sa.func.standard_hash(compiled_arg, sa.sql.literal_column("'SHA256'"))
     return sa.func.lower(hash_func)
 
+def sa_format_hashbytes_db2(translator, expr):
+    arg, how = expr.op().args
+    compiled_arg = translator.translate(arg)
+    hashfunc = sa.func.hash(compiled_arg,sa.sql.literal_column("2"))
+    return sa.func.hex(hashfunc)
```
piyushsarraf: I think we have to convert the hash to lower case before returning `sa.func.lower(sa.func.hex(hashfunc))`. By default as I have checked it returns hash value in caps, so validation fails.

ngdav: Thanks for catching that, I've made the changes

nehanene15 (approved):
LGTM!

nehanene15:
We could also add random row support (or leave that for another PR) Looks like DB2 supports `RAND()` so it should just be one line add [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/data_validation/query_builder/random_row_builder.py#L54)

PR #802, opened 2023-04-13
feat: Option for simpler report output grid

arjitshu:
We can now use **--format=text/csv/json** without any extra columns in output.


Raniksingh (approved):
LGTM

PR #803, opened 2023-04-13
feat: added source_type in output  while listing connections list

SwatiT17:
/gcbrun

### DIFF
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -990,7 +990,7 @@ def list_connections():
     connections = mgr.list_connections()
     for conn_name in connections:
         source_type = mgr.get_connection_config(conn_name).get("source_type")
-        logging.info(f"Connection Name: {conn_name} : {source_type}")
+        logging.info(f"Connection Name: {conn_name} :{source_type}")
 
 
 def get_connection(connection_name):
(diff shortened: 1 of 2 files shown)


SwatiT17:
Fixes issue: #792 data-validation connections list output is limited

data_validation/cli_tools.py
```diff
@@ -988,9 +988,9 @@ def list_connections():
     """List all saved connections."""
     mgr = state_manager.StateManager()
     connections = mgr.list_connections()
-
     for conn_name in connections:
-        logging.info(f"Connection Name: {conn_name}")
+        source_type = mgr.get_connection_config(conn_name).get("source_type")
```
Raniksingh, line 992: Should we also add target_type along with source? @nehanene15 any thoughts?

mahtodeepak05: @Raniksingh 
AFAIU, connection listing implies databases whose connection details is added,  we can't confirm whether it is used as target when using DVT. source_type itself should suffice, it says to which source\database connection is pointing to.

nehanene15 (commented):
Looks good! Maybe we can update this unit test to reflect the new output - https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/tests/unit/test_cli_tools.py#L117

Raniksingh (approved):

Issue #776, opened 2023-03-17
validate row: MySQL ibis.expr.operations.IfNull Exception


nj1973:
Any `validate row` command on MySQL (both `--hash` and `--concat`) throws an exception:
```
...
  File "/home/user/professional-services-data-validator/env/lib/python3.10/site-packages/ibis/backends/base_sqlalchemy/compiler.py", line 1362, in get_result
    translated = self.translate(self.expr)
  File "/home/user/professional-services-data-validator/env/lib/python3.10/site-packages/ibis/backends/base_sqlalchemy/compiler.py", line 1403, in translate
    raise com.OperationNotDefinedError(
ibis.common.exceptions.OperationNotDefinedError: No translation rule for <class 'ibis.expr.operations.IfNull'>
```

When complete we need to enable the integration test for this problem. Search for "issue-776" in tests/system/data_sources/test_mysql.py to action this (there are two disabled tests). Enabling these test is blocked by issue https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/759.

nehanene15:
This has been resolved in #812 and uncommented from MySQL tests.

PR #812, opened 2023-04-20
fix: Mysql fix to support row hash validations, random row validation, and filter

kanhaPrayas:
This PR adds the fixes for MYSQL support with DVT for Row validations.


nehanene15 (approved):
LGTM, but can we also enable the integration test for this as per #776  Search for "issue-776" in tests/system/data_sources/test_mysql.py to action this (there are two disabled tests). 

There are also some unit tests not passing

PR #828, opened 2023-05-02
fix: Http server impala

renzokuken:
adding options to pass http proxy for impala clients

### DIFF
--- a/third_party/ibis/ibis_impala/api.py
+++ b/third_party/ibis/ibis_impala/api.py
@@ -39,7 +39,9 @@ def impala_connect(
     user=None,
     password=None,
     pool_size=8,
-    hdfs_client=None
+    hdfs_client=None,
+    use_http_transport=False,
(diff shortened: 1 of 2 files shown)


third_party/ibis/ibis_impala/api.py
```diff
@@ -63,7 +65,9 @@ def impala_connect(
         user=user,
         password=password,
         pool_size=pool_size,
-        hdfs_client=hdfs_client
+        hdfs_client=hdfs_client,
```
nehanene15, line 68: I don't know if this will work since the [connect() method ](https://github.com/ibis-project/ibis/blob/1.4.0/ibis/backends/impala/__init__.py#L41) doesn't support the new flags. We'll probably have to write our own connect() signature and copy the original connect logic.

nehanene15:
Closing since we opened a new PR for the fix: #829

Issue #827, opened 2023-05-02
MSSQL schema validation errors with data type and find-tables


nehanene15:
- [x] SQL Server to Postgres find-tables is not showing all matching tables with schema 'dbo'
- [x] SQL Server MONEY data type is not supported and excluding the column still parses though the whole schema

Supporting the MONEY data type will involve adding the mapping [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/third_party/ibis/ibis_mssql/compiler.py#L492) as `{dt.Int64: mssql.MONEY}` and testing.

nehanene15:
I would also add the following for the SQL Server [clients.py](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/third_party/ibis/ibis_mssql/client.py#L33):
```
@dt.dtype.register(MSDialect_pyodbc, sa.dialects.mssql.MONEY)
def sa_money(_, satype, nullable=True):
    return dt.Int64(nullable=nullable)
```

helensilva14:
>  SQL Server to Postgres find-tables is not showing all matching tables with schema 'dbo'

By checking the code and doing tests, we confirmed there is no bug with the current implementation of `find-tables`, the strategy to get results is to decrease the value for the `score-cutoff` param since we currently have 1 as default value and it seeks for identical matches. By using numbers such as 0.7, 0.65 etc you can get more matches.

For reference, we make use of this [jaro_similarity method](https://jamesturk.github.io/jellyfish/functions/#jaro-similarity) for the string comparison.

PR #846, opened 2023-05-11
docs: Update our documentation about `find-tables` command and the `score-cutoff` parameter

helensilva14:
Improves our docs about the `find-tables` command related to the issue #827


Raniksingh (approved):
LGTM

PR #856, opened 2023-05-12
feat: Numpy for random row generation for faster execution

### LINKED ISSUE #844: Random Row Validation - Not performant on large MySQL tables
Random Row validation uses ORDER BY RAND() on MySQL tables which is not performant on large tables. 
Recommendation to calculated MIN() and MAX() and generate Random IDs for monotonically increasing PKs

nehanene15:
Closes #844 

Only works for monotonically increasing numeric PKs.


nehanene15:
Closing this PR without merging since the random row generation did not speed up random row generation significantly.

Issue #858, opened 2023-05-15
Add support to missing SQL Server data types available on current Ibis implementation


helensilva14:
Check on this link https://github.com/ibis-project/ibis/blob/master/ibis/backends/mssql/datatypes.py#L47 which ones to add

helensilva14:
These data types are now supported after the Ibis upgrade (https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/894)!

PR #859, opened 2023-05-16
feat: Add support for SQL Server's IMAGE, BINARY, VARBINARY, NCHAR, NTEXT, NVARCHAR data types (#858)

helensilva14:
Related to #858: adding some of the missing data types such as IMAGE, BINARY, VARBINARY, NCHAR, NTEXT, NVARCHAR

### DIFF
--- a/third_party/ibis/ibis_mssql/compiler.py
+++ b/third_party/ibis/ibis_mssql/compiler.py
@@ -489,6 +489,7 @@ class MSSQLExprTranslator(alch.AlchemyExprTranslator):
     _registry = _operation_registry
     _rewrites = alch.AlchemyExprTranslator._rewrites.copy()
     _type_map = alch.AlchemyExprTranslator._type_map.copy()
+    # only used to map SQLAlchemy types back to SQL Server types, it does not add support to new data types, for that go to client.py
     _type_map.update(
         {
             dt.Boolean: mssql.BIT,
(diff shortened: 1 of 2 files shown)


third_party/ibis/ibis_mssql/compiler.py
```diff
@@ -499,6 +499,14 @@ class MSSQLExprTranslator(alch.AlchemyExprTranslator):
             dt.Double: mssql.REAL,
             dt.String: mssql.VARCHAR,
             dt.Int64: mssql.MONEY,
+            # Unicode character strings
+            dt.String: mssql.NCHAR,
```
nj1973: I don't think this is right. Because this is an `update()` only one of the `dt.String` key value pairs will persist, the others will be overwritten. Did Ibis do this too?

helensilva14: Ahhh I see what you mean, in Ibis code the mapping is reversed SQL Server Type to Ibis Data type: https://github.com/ibis-project/ibis/blob/master/ibis/backends/mssql/datatypes.py#L72. I just adapted to our implementation.

nj1973: We still have two entries for `dt.String`:

Which one survives will be open to chance:
```
            dt.String: mssql.VARCHAR,    <<<<<
            dt.Int64: mssql.MONEY,
            # Unicode character strings
            dt.String: mssql.NCHAR,      <<<<<
```

The Ibis code you linked to above only defines a key-value pair for `dt.String` once:
```
_MSSQL_TYPE_MAP = {
    dt.Boolean: mssql.BIT,
    dt.Int8: mssql.TINYINT,
    dt.Int16: mssql.SMALLINT,
    dt.Int32: mssql.INTEGER,
    dt.Int64: mssql.BIGINT,
    dt.Float16: mssql.FLOAT,
    dt.Float32: mssql.FLOAT,
    dt.Float64: mssql.REAL,
    dt.String: mssql.NVARCHAR,
}
```

It looks to me like those are used for mapping from SQLAlchemy types back to SQL Server types. Are you sure you need top change this bit?

nehanene15: Yeah, you probably don't need to change this as long as you registered the data types in third_party/ibis/ibis_mssql/client.py

nj1973 (approved):
LGTM

Issue #719, opened 2023-02-10
MySql - Spanner Comp-Field Validation failing for FLOAT datatype


madhu-jain:
Comp field comparison between SQL server and Big query for FLOAT datatype (column name : adjusted) failing.

Source data (SQL Server)

sr_no       adjusted                
----------- ------------------------
 2217126       7.3018260000000001
 2622064     4.8977870000000001
 3339058     43.23809
 3960098     60.973129999999998
 4585861      8.3525019999999994
 10158240     2.598392

Target data (Bigquery)

sr_no | adjusted |  
2217126 | 7.301826 |  
2622064 | 4.897787 |  
3339058 | 43.23809 |  
3960098 | 60.97313 |  
4585861 | 8.352502 |  
10158240 | 2.598392

Command used : 

data-validation validate row \
-sc MSSQL_CONN \
-tc BQ_CONN \
-tbls tempdb.dbo.dvt_prices=paras-sandbox-364713.dlp.dvt_prices \
--primary-keys sr_no \
-comp-fields symbol,adjusted \
-rr \
-rbs 5 \
--bq-result-handler paras-sandbox-364713.data_validator.results

Output : 

<img width="983" alt="Screenshot 2023-02-11 at 12 33 52 AM" src="https://user-images.githubusercontent.com/124554976/218176118-440b72ed-5455-4677-b463-d9a753cd1b11.png">

nehanene15:
@piyushsarraf Does #771 fix this issue as well?

nehanene15:
This will be resolved with the Ibis upgrade - I wouldn't work on this until then.
cc @piyushsarraf

piyushsarraf:
The issue persists with MySql and Spanner comp field validation with float datatype. 
Source MySql has Float(10, 2) and Target Spanner has float64 datatype

In custom-query we are checking a single row which has 4.2 as value in source and target. But the validation fails. 
![image](https://github.com/GoogleCloudPlatform/professional-services-data-validator/assets/47020544/ac1982a6-9851-432a-b844-6755346f91c8)

The issue is there is a value mismatch when we are casting Float(10,2) to float64 in pandas dataframe. 
[Source df](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/8c76c1bba214c1848d7cdcb401e5e28d3153a0a9/data_validation/data_validation.py#L306) value: _values:array([[4.19999981, 6.        ]])
[...]

PR #860, opened 2023-05-16
feat: [ON HOLD] Add support to other missing SQL Server data types - check list on commit 

### LINKED ISSUE #858: Add support to missing SQL Server data types available on current Ibis implementation

helensilva14:
Closes #858 -> Added types: SMALLINT, TIMESTAMP, SMALLMONEY,

In progress: DECIMAL, NUMERIC,  DATE, DATETIME2, DATETIME, DATETIMEOFFSEST, SMALLDATETIME, TIME, CHAR, TEXT

### DIFF
--- a/third_party/ibis/ibis_mssql/client.py
+++ b/third_party/ibis/ibis_mssql/client.py
@@ -55,6 +55,7 @@ def sa_money(_, satype, nullable=True):
     return dt.Int64(nullable=nullable)
 
 
+<<<<<<< HEAD
 @dt.dtype.register(MSDialect_pyodbc, sa.dialects.mssql.BINARY)
 def sa_binary(_, satype, nullable=True):
     return dt.Binary(nullable=nullable)
@@ -65,11 +66,14 @@ def sa_varbinary(_, satype, nullable=True):
     return dt.Binary(nullable=nullable)
 
 
+=======
+>>>>>>> feat: Add support for SQL Server's IMAGE data type (#858)
 @dt.dtype.register(MSDialect_pyodbc, sa.dialects.mssql.IMAGE)
 def sa_image(_, satype, nullable=True):
     return dt.Binary(nullable=nullable)
 
 
+<<<<<<< HEAD
 @dt.dtype.register(MSDialect_pyodbc, sa.dialects.mssql.NCHAR)
 def sa_nchar(_, satype, nullable=True):
     return dt.String(nullable=nullable)
@@ -85,6 +89,8 @@ def sa_nvarchar(_, satype, nullable=True):
     return dt.String(nullable=nullable)
 
 
+=======
+>>>>>>> feat: Add support for SQL Server's IMAGE data type (#858)
 class MSSQLTable(alch.AlchemyTable):
     pass
 
--- a/third_party/ibis/ibis_mssql/client.py
+++ b/third_party/ibis/ibis_mssql/client.py
@@ -80,6 +80,16 @@ def sa_image(_, satype, nullable=True):
     return dt.Binary(nullable=nullable)
 
 
+@dt.dtype.register(MSDialect_pyodbc, sa.dialects.mssql.CHAR)
+def sa_char(_, satype, nullable=True):
+    return dt.String(nullable=nullable)
+
+
+@dt.dtype.register(MSDialect_pyodbc, sa.dialects.mssql.TEXT)
+def sa_text(_, satype, nullable=True):
+    return dt.String(nullable=nullable)
+
+
 @dt.dtype.register(MSDialect_pyodbc, sa.dialects.mssql.NCHAR)
 def sa_nchar(_, satype, nullable=True):
     return dt.String(nullable=nullable)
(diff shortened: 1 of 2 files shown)


third_party/ibis/ibis_mssql/client.py
```diff
@@ -55,6 +60,26 @@ def sa_money(_, satype, nullable=True):
     return dt.Int64(nullable=nullable)
 
 
+@dt.dtype.register(MSDialect_pyodbc, sa.dialects.mssql.SMALLMONEY)
```
nj1973, line 63: I would question whether `int32` is the correct type to use.
Both `smallmoney` and `money` can hold fractional values so they would most likely be stored as a decimal/numeric in Cloud SQL or BigQuery.
I suspect `smallmoney` should become `Decimal(10,4)` and `money` should become `Decimal(19,4)`.
Or did I miss something?

third_party/ibis/ibis_mssql/compiler.py
```diff
@@ -489,18 +489,47 @@ class MSSQLExprTranslator(alch.AlchemyExprTranslator):
...
             dt.Int8: mssql.TINYINT,
+            dt.Int16: mssql.SMALLINT,
             dt.Int32: mssql.INTEGER,
             dt.Int64: mssql.BIGINT,
-            dt.Float: mssql.REAL,
-            dt.Double: mssql.REAL,
-            dt.String: mssql.VARCHAR,
+            dt.Float16: mssql.FLOAT,
+            dt.Float32: mssql.FLOAT,
+            dt.Float64: mssql.REAL,
```
nj1973: Have we got Float and REAL the wrong way around here?

According to SQL Server docs: https://learn.microsoft.com/en-us/sql/t-sql/data-types/float-and-real-transact-sql?view=sql-server-ver16

"real is float(24)"

Which, I think, means real holds 32 bit values and float can go up to 64 bit.

third_party/ibis/ibis_mssql/compiler.py
```diff
@@ -489,18 +489,47 @@ class MSSQLExprTranslator(alch.AlchemyExprTranslator):
...
             dt.Int64: mssql.MONEY,
```
nj1973: As mentioned earlier, I think the money types should not be integers.

nj1973 (commented):
Perhaps it is a job for a follow on issue but a few of the new types fail row validation, for example:
```
ibis.common.exceptions.IbisTypeError: Given argument with datatype uuid is not subtype of string nor implicitly castable to it
```

third_party/ibis/ibis_mssql/client.py
helensilva14: Hi Neil! For the implementation for `MONEY` and `SMALLMONEY` as well as the other data types I followed the mapping available in the current Ibis implementation here: https://github.com/ibis-project/ibis/blob/master/ibis/backends/mssql/datatypes.py#L55.

I do agree with you that `DECIMAL` would make more sense, though. Not sure how we proceed about it, any ideas?

third_party/ibis/ibis_mssql/compiler.py
helensilva14: Same as my other answer about the reference for my implementation (https://github.com/ibis-project/ibis/blob/master/ibis/backends/mssql/datatypes.py#L96) and actually this part really confused me, specially having a duplicate mapping for FLOAT.  

So you suggest this?

```suggestion
            dt.Float64: mssql.FLOAT,
            dt.Float32: mssql.REAL,
```

third_party/ibis/ibis_mssql/client.py
nj1973: My word, I wonder why they've done that.
I still think the Ibis code looks wrong but perhaps I'm missing something. If I were migrating SQL Server to BigQuery or PostgreSQL and had MONEY columns then I would define them as NUMERIC (in BigQuery) or DECIMAL (in PostgreSQL), not INT64.
Having said that, perhaps it's worth getting a second opinion, I do see the sense in doing the same as Ibis.

third_party/ibis/ibis_mssql/compiler.py
nj1973: In the code right above the code you linked to Ibis maps like this:
```
    # Approximate numerics
    'FLOAT': dt.Float64,
    'REAL': dt.Float32,
```

But then they do the opposite right below (as you pointed out). Again, it looks wrong to me but I might be missing something. Perhaps it is safest to do as you've done until we run into an issue. We could easily test this using row validation on a big 64 bit value in a FLOAT column (I'm not suggesting you do this now but we could come back to it - or raise an issue to remind us to test it).

Perhaps the best way forward is to add a comment to our code saying we're not convinced it is right but have gone with it based on Ibis code?

nj1973: I wonder if this could be related: https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/719

third_party/ibis/ibis_mssql/client.py
ghost: In Database Migration Service, for SQL Server to postgres, this is the mapping:

DECIMAL(p,s) Fixed point number. Decimal and numeric are synonyms and can be used interchangeably.  −10^38+1 to 10^38-1. 5 to 17 bytes 
NUMERIC(p,s) 
up to 131072 digits before the decimal point; up to 16383 digits after the decimal point. 5 to 17 bytes

SMALLMONEY
4 bytes currency amount | MONEY | 8 bytes currency amount

MONEY | 8 bytes currency amount | numeric(19,4) | 8 bytes. Scale matches that of money in SQL Server. Scale of Money is 2 in Postgres which is 4 in SQL Server so numeric gives the correct mapping of data typ

helensilva14:
These data types are now supported after the Ibis upgrade (#894)!

PR #870, opened 2023-05-23
fix: Add missing operations for SQL Server - ExtractEpochSeconds, ExtractDayOfYear, ExtractWeekOfYear

### LINKED ISSUE #869: ibis.common.exceptions.OperationNotDefinedError: No translation rule for <class 'ibis.expr.operations.ExtractEpochSeconds'>
This function is missing a mapping/translation rule, we might check other missing ones as well and refer to this code: https://github.com/ibis-project/ibis/blob/master/ibis/backends/mssql/registry.py#L177

helensilva14:
Closes #869 

/gcbrun

### DIFF
--- a/third_party/ibis/ibis_mssql/compiler.py
+++ b/third_party/ibis/ibis_mssql/compiler.py
@@ -371,9 +371,28 @@ def _day_of_week_name(t, expr):
     (sa_arg,) = map(t.translate, expr.op().args)
     return sa.func.trim(sa.func.format(sa_arg, 'dddd'))
 
+
 def _string_join(t, expr):
     sep, elements = expr.op().args
     return sa.func.concat(*map(t.translate, elements))
+
+
+def _timestamp_truncate(t, op):
+    arg = t.translate(op.arg)
+    unit = op.unit.short
+    if unit not in _truncate_precisions:
+        raise com.UnsupportedOperationError(f'Unsupported truncate unit {op.unit!r}')
+
+    return sa.func.datetrunc(sa.text(_truncate_precisions[unit]), arg)
+
+
+def _timestamp_from_unix(x, unit='s'):
+    if unit == 's':
+        return sa.func.dateadd(sa.text('s'), x, '1970-01-01 00:00:00')
+    if unit == 'ms':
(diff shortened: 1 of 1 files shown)


third_party/ibis/ibis_mssql/compiler.py
```diff
@@ -418,11 +437,25 @@ def _string_join(t, expr):
...
+                sa.func.datediff(sa.text('s'), '1970-01-01 00:00:00', x), sa.BIGINT
+            ),
+            1,
+        ),
+        ops.TimestampFromUNIX: lambda t, op: _timestamp_from_unix(
+            t.translate(op.arg), op.unit.short
+        ),
+        ops.TimestampTruncate: _timestamp_truncate,
+        ops.DateTruncate: _timestamp_truncate,
+        ops.Hash: unary(sa.func.checksum),
```
nehanene15: We shouldn't add hash here since it's already added in ibis_addon/operations.py

third_party/ibis/ibis_mssql/compiler.py
```diff
@@ -371,10 +371,29 @@ def _day_of_week_name(t, expr):
     (sa_arg,) = map(t.translate, expr.op().args)
     return sa.func.trim(sa.func.format(sa_arg, 'dddd'))
 
+
 def _string_join(t, expr):
     sep, elements = expr.op().args
     return sa.func.concat(*map(t.translate, elements))
 
+
+def _timestamp_truncate(t, op):
```
nehanene15: Has this been tested? I'm not sure if this will work since the older versions of Ibis use different signatures i.e. function(t, expr) instead of function(t,ops).

I think we can stick to supporting ExtractEpochSeconds in this PR since the other methods aren't used in DVT

nehanene15 (changes requested):

third_party/ibis/ibis_mssql/compiler.py
helensilva14: Agreed, I didn't know how to test all of them but the ExtractEpochSeconds is working! Removed the others

nehanene15 (approved):
LGTM

PR #888, opened 2023-06-16
feat: Support allow list decimals having a range for precision and scale. Also add --allow-list-file.

### LINKED ISSUE #887: validate schema: Enhancements to --allow-list option
**Enhancement 1**
Oracle does not have native integral data types like int, bigint, smallint. All numbers are decimal in nature.
This means validation of a migration from Oracle to PostgreSQL is generating a lot of --allow-list entries. Example below from a simple 3 table validation:

```
--allow-list="decimal(38,0):int32[non-nullable],decimal(38,0):int32,decimal(38,0):int32[non-nullable],decimal(3,0):int32,decimal(3,0):int16[non-nullable],decimal(6,0):int32[non-nullable]"
```

I suspect we'll have similar challenges for Oracle to BigQuery migrations.

It would be a nice option to be able to provide decimal precision as a range, for example:

`decimal(1-9,0):int32`: Matches any decimal precision between 1 and 8, which are all valid for int32.
`decimal(1-18,0):int64`: Matches any decimal precision between 1 and 18, which are all valid for int64.

**Enhancement 2**

Another customer request was to support a config file for the allowlist, e.g. `--allow-list orace_pg.yaml` so they can reuse mappings between certain systems and it will prevent long configs on the command line.

nj1973:
This PR add the following functionality:

- `--allow-list` supports integer ranges in decimal specifications
- `--allow-list-file` enables the user to provide a YAML file of `--allow-list` pairs

**`--allow-list` Example**

```
data-validation validate schema -sc=ora_local -tc=pg_local -tbls=dvt_test.dvt_types \
--allow-list="decimal(1-4,0):int16,decimal(1-9,0):int32,decimal(1-18,0):int64"
```
Imagine trying to cover all of that on the command line!

**`--allow-list-file` Examples**
```
data-validation validate schema -sc=ora_local -tc=pg_local -tbls=dvt_test.dvt_types \
-alf=samples/allow_list/oracle_to_bigquery.yaml

data-validation validate schema -sc=ora_local -tc=pg_local -tbls=dvt_test.dvt_types \
--allow-list-file=gs://dvt-bucket/allow-list-files/oracle_to_bigquery.yaml
```

These options can be used at the same time which enables the user to have base config and provide extra mappings when required:
```
data-validation validate schema -sc=ora_local -tc=pg_local -tbls=dvt_test.dvt_types \
-alf=samples/allow_list/oracle_to_bigquery.yaml --allow-list="decimal(4,0):int16"
```

I've added two example allow list files with safe mappings:
```
$ ls -l samples/allow_list
-rw-r----- 1 526 Jun 16 15:36 oracle_to_bigquery.yaml
-rw-r----- 1 372 Jun 16 16:26 oracle_to_postgres.yaml
```


data_validation/schema_validation.py
```diff
@@ -216,40 +228,99 @@ def schema_validation_matching(
...
+            expanded_pairs.extend(
+                [(_[0], _[1]) for _ in itertools.product(dt1_list, dt2_list)]
+            )
+        return expanded_pairs
+
+    def convert_pairs_to_dict(expanded_pairs: list) -> dict:
+        return_pairs = {}
+        for dt1, dt2 in expanded_pairs:
+            if dt1 in return_pairs:
+                return_pairs[dt1].append(dt2)
```
svetakvsundhar, line 321: shouldn't this be `dt[1]=dt[2]`?

svetakvsundhar: if not, can you provide a description of what this function is doing?

data_validation/schema_validation.py
```diff
@@ -216,40 +228,99 @@ def schema_validation_matching(
...
+        return_list = [
+            f"{type_name}({p},{s})" for p, s in itertools.product(p_list, s_list)
+        ]
+    else:
+        return_list = [f"{type_name}({_})" for _ in p_list]
+    return return_list
+
+
+def parse_allow_list(st: str) -> dict:
+    """Convert allow-list data type pairs into a dictionary like {key[value1, value2, etc], }"""
```
svetakvsundhar, line 299: what is the value in the dictionary here?

tests/unit/test_schema_validation.py
```diff
@@ -15,9 +15,17 @@
 import json
 import pytest
 import random
+import re
 from datetime import datetime, timedelta
```
svetakvsundhar, line 19: Can we make use of Python Unittest (https://docs.python.org/3/library/unittest.html) here?

svetakvsundhar (changes requested):

data_validation/schema_validation.py
nj1973: It is how it was in the version of the code before I modified it:
```
# Converting allow list from string to map, to ease validation.
def parse_allow_list(st):
    output = {}
    stack = []
    key = None
    for i in range(len(st)):
        if st[i] == ":":
            key = "".join(stack)
            if key not in output:
                output[key] = []
            stack = []
            continue
        if st[i] == "," and not st[i + 1].isdigit():
            value = "".join(stack)
            output[key].append(value)
            stack = []
            i += 1
            continue
        stack.append(st[i])
    value = "".join(stack)
    output[key] = value
    stack = []
    return output
```

The value is that we can reference the list of valid target types for a source type directly without search for it. You can see this in action in `schema_validation_matching()`.

data_validation/schema_validation.py
nj1973: No, it should not be `dt[1]=dt[2]` because there is no list named `dt`.

I'll add this docstr to the function to help explain:
```
"""Take the list data type tuples and convert them into a dictionary keyed on source data type.
        For example:
            [('decimal(2,0)', 'int64'), ('decimal(2,0)', 'int32')]
        becomes:
            {'decimal(2,0)': ['int64', 'int32']}
        """
```

tests/unit/test_schema_validation.py
nj1973: We use `pytest` in DVT for unit testing. We've only used `unittest` for mocking objects.

data_validation/schema_validation.py
nj1973: Did that satisfy the query?
Or do you think my code should use a list called `dt` instead of two variables `dt1` and `dt2`?

svetakvsundhar: Thanks, docstring helps

svetakvsundhar (approved):

svetakvsundhar:
LGTM, thank you!

PR #889, opened 2023-06-17
feat!: Partition based on non-numeric and multiple keys

sundar-mudupalli-work:
This PR includes code, testing and documentation changes to partition tables based on non-numeric and multiple keys. The partitioned tables can be used to perform `validate row` and possibly `grouped-columns` validations without running into `MemoryError` issues.

One system test was added to the postgres test set to partition a small database with multiple numeric and string keys. The results are checked against the expected value. One unit test was deleted that was written for partitioning a numeric key in pandas - the updated approach with ntile does not work on pandas tables. 

This code works correctly with Oracle and Postgres databases - has not been tested with other databases.

BREAKING CHANGE:  generate-table-partitions no longer supports a partition-key parameter, since the primary key(s) are used to partition the table.


svetakvsundhar (commented):
Thanks. I'm not completely familiar with how this tool works, so I will defer to someone with more knowledge of the tool for a review.

data_validation/query_builder/partition_row_builder.py
```diff
@@ -20,7 +20,7 @@
 class PartitionRowBuilder(object):
     def __init__(
         self,
-        partition_key: str,
+        primary_keys: [str],
```
renzokuken, line 23: @sundar-mudupalli-work, what is the rationale for changing the partition key argument to primary key? I'm not opposed just worried that primary key may be referenced in other paths of the tool

tests/unit/test_clients.py
```diff
@@ -85,7 +85,7 @@ def test_get_oracle_data_client():
         clients.get_data_client(ORACLE_CONN_CONFIG)
 
 
-def test_get_pandas_data_client(fs):
+def test_get_pandas_data_client():
```
renzokuken, line 88: is this just cleanup or was the test failing?

renzokuken (commented):
Just a couple of questions in-line. Otherwise LGTM

data_validation/query_builder/partition_row_builder.py
sundar-mudupalli-work: We use the primary-key(s) to match corresponding rows between the source and target. When we partition the table, we only compare one partition in the source with the equivalent (corresponding) partition in the target. So ensuring the same set of primary keys fall into the same partition in both the source and target is necessary and sufficient. If the partition-key uniquely identifies the row, then the partition key can be used instead of the primary key. I am going to update the document on this point - the primary key is only used for matching rows and does not need to be a `primary` key in the table schema. When we say primary key, it can be any column (or set of columns) that uniquely identifies the row.

tests/unit/test_clients.py
sundar-mudupalli-work: I saw an error one time saying fs is not a valid fixture. I can no longer reproduce it. Therefore consider it a cleanup.

data_validation/query_builder/partition_row_builder.py
sundar-mudupalli-work: change that to sequence of columns - the order is important.

renzokuken (approved):
Approved peinding primary key definition is updated. LGTM

nehanene15:
Looks great Sundar! One comment - can we also document that we only support Oracle and Postgres for this feature? Also, did you want to refer to the internal partition PRD you created in the main README?

PR #903, opened 2023-07-13
feat: Adding RawSQL function for Redshift

piyushsarraf:
* This PR adds RawSql function for RedShift, now we will be able to execute a dry run query and add filters to validation query


Raniksingh (approved):

m-zain:
Hi @Raniksingh @nehanene15 thank you for this PR, it has resolved customer' issue for Redshift filter feature for random row validation, however filter filter feature is not working for hash_all validation. 
After the PR filter feature has started to work for random_row validation but still failing for hash_all validation. 

Can this be fixed please or it is not supposed to work for hash_all?

nehanene15:
@m-zain Can you provide an example command and the error message please?

Issue #754, opened 2023-03-02
validate column: sum, min, max false success for numeric values with precision > 16


nj1973:
Test BigQuery tables:
```
create table `dvt_test.tab_long_dec1` (
  id int64 not null
, col_16 numeric(16)
, col_17 numeric(17)
, col_18 numeric(18));
insert into dvt_test.tab_long_dec1 values
(1,NUMERIC '1234567890123456',NUMERIC '12345678901234567',NUMERIC '123456789012345678');
create table dvt_test.tab_long_dec2 (
  id int64 not null
, col_16 numeric(16)
, col_17 numeric(17)
, col_18 numeric(18));
insert into dvt_test.tab_long_dec2 values
(1,NUMERIC '1234567890123456'+1,NUMERIC '12345678901234567'+1,NUMERIC '123456789012345678'+1);
```
Note that tab_long_dec2 data differs from tab_long_dec1. The value in col_16, col_27 and col_18 has been incremented by 1.
Therefore sum and max comparison of these columns should fail.

**col_16**
```
data-validation validate column -sc bq -tc bq \
-tbls=dvt_test.tab_long_dec1=dvt_test.tab_long_dec2 \
-sum=col_16 -max=col_16
```
Correct behaviour, all columns have validation_status="fail"

**col_17**
```
data-validation validate column -sc bq -tc bq \
-tbls=dvt_test.tab_long_dec1=dvt_test.tab_long_dec2 \
-sum=col_17 -max=col_17
```
Incorrect behaviour, `sum__col_17` and `max__col_17` have validation_status="success". This is wrong.

**col_18**
```
data-validation validate column -sc bq -tc bq \
-tbls=dvt_test.tab_long_dec1=dvt_test.tab_long_dec2 \
-sum=col_18 -max=col_18
```
Incorrect behaviour, `sum__col_18` and `max__col_18` have validation_status="success". This is wrong.

I suspect this relates to code in `data_validation/combiner.py` that casts all values to `float64`. 64 bit float is a lossy data type for large values.
I saw the same issue when testing PostgreSQL too, I believe the problem is not engine specific.

nj1973:
We should also add an integration test for this.

nj1973:
Closed by fix for issue 900, duplicate issues.

PR #918, opened 2023-07-27
fix: validate column sum/min/max issue for decimals with precision beyond int64/float64 

### LINKED ISSUE #900: validate column: Fix for issue 763 has side effect which causes false positive validations
https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/763 was raised for an exception thrown when Oracle NUMBER columns generate a value greater than int64.

This fix, linked here https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/809/files, had side effects which need to be resolved. The changes cast all decimal columns to float 64 before summing them. This has the negative effect that NUMBER values with a precision > float 64 will be rounded and the original value lost. This produces false positive results when there is a genuine mismatch.

PostgreSQL example of wrong results this generates, notice how a 30 digit value lost all values after 16th digit:

```
# select cast(cast(123456789012345678901234567890 as double precision) as decimal(38));
            numeric
--------------------------------
 123456789012346000000000000000
```

We need to reverse the original fix and find a new fix without side effects.

nj1973:
This PR adds a change to cast decimal aggregations to string if the source column precision > 18. Prior to this change the values were cast to float64 which would lose precision and give false success statuses. Because the change casts to string *after* any aggregations there should be no side effects.

I have tested this for `column` and `custom-query` validations and have added a regression test to `test_oracle.py`.


nehanene15 (approved):
LGTM, this is great. Didn't realize we already had support to CAST within the aggregate field config.

nehanene15:
I believe this closes Issue #754 as well

PR #919, opened 2023-07-27
feat: Enhance validate schema to support time zoned timestamp columns

### LINKED ISSUE #706: validate schema: TIMESTAMP WITH TIME ZONE fails for Oracle/PostgreSQL validation
Oracle test table:
```
create table dvt_test.tab_tstz (
id number(2) not null, col_tstz timestamp(0) with time zone
);
ALTER TABLE dvt_test.tab_tstz ADD PRIMARY KEY (id);
insert into dvt_test.tab_tstz values 
(1,to_timestamp_tz('1973-10-31 23:22:21 +00:00','YYYY-MM-DD HH24:MI:SS TZH:TZM'));
insert into dvt_test.tab_tstz values 
(2,to_timestamp_tz('1973-10-31 23:22:21 +02:00','YYYY-MM-DD HH24:MI:SS TZH:TZM'));
commit;
```

PostgreSQL test table:
```
create table dvt_test.tab_tstz (
id decimal(2) not null, col_tstz timestamp(0) with time zone
);
ALTER TABLE dvt_test.tab_tstz ADD PRIMARY KEY (id);
insert into dvt_test.tab_tstz values (1,timestamp'1973-10-31 23:22:21 +00:00');
insert into dvt_test.tab_tstz values (2,timestamp'1973-10-31 23:22:21 +02:00');
```

DVT command:
```
data-validation validate schema \
-sc ora_conn -tc pg_conn \
-tbls dvt_test.tab_tstz=dvt_test.tab_tstz
```

Output (relevant columns only):
```
╤══════════════════╤══════════════════╤═════════════════╤═════════════════╤═══════╕
│source_column_name│target_column_name│ source_agg_value│ target_agg_value│status │
╪══════════════════╪══════════════════╪═════════════════╪═════════════════╪═══════╡
│id                │id                │ decimal(2, 0)   │ decimal(2, 0)   │success│
┼──────────────────┼──────────────────┼─────────────────┼─────────────────┼───────┤
│col_tstz          │col_tstz          │ timestamp       │ timestamp('UTC')│ fail  │
╧══════════════════╧══════════════════╧═════════════════╧═════════════════╧═══════╛
```

Try to use `--allow-list` workaround but validation output is the same as above:
```
data-validation validate schema \
-sc ora_conn -tc pg_conn \
-tbls dvt_test.tab_tstz=dvt_test.tab_tstz \
--allow-list 'timestamp:timestamp('UTC')'
```

nj1973:
This check ensures all engines (where relevant) pick up the time zone for time zoned timestamp columns.
Also enabled integration tests for col_tstz columns.
There are still problems with column and row validation, this PR is only for schema validation.

### DIFF
--- a/third_party/ibis/ibis_oracle/datatypes.py
+++ b/third_party/ibis/ibis_oracle/datatypes.py
@@ -40,7 +40,7 @@ class _FieldDescription(TypedDict):
     internal_size: Optional[int]
     precision: Optional[int]
     scale: Optional[int]
-    null_ok: Optional[int] 
+    null_ok: Optional[int]
 
 
 def _get_type(col: _FieldDescription) -> dt.DataType:
@@ -50,7 +50,7 @@ def _get_type(col: _FieldDescription) -> dt.DataType:
         raise NotImplementedError(
             f"Oracle type {typename} is not supported"
         )
-    
+
     if typename == cx_Oracle.DB_TYPE_NUMBER:
         if col[4] == 0 and col[5] == -127:
             # This will occur if type is NUMBER with no precision/scale
@@ -140,7 +140,10 @@ def sa_oracle_NCHAR(_, satype, nullable=True):
 
 @dt.dtype.register(OracleDialect_cx_oracle, sa.dialects.oracle.TIMESTAMP)
 def sa_oracle_TIMESTAMP(_, satype, nullable=True):
-    return dt.Timestamp(nullable=nullable)
+    if satype.timezone:
+        return dt.Timestamp(timezone="UTC", nullable=nullable)
+    else:
+        return dt.Timestamp(nullable=nullable)
 
 @dt.dtype.register(OracleDialect_cx_oracle, sa.dialects.oracle.BLOB)
 def sa_oracle_BLOB(_, satype, nullable=True):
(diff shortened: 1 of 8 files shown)


third_party/ibis/ibis_oracle/datatypes.py
```diff
@@ -140,7 +140,10 @@ def sa_oracle_NCHAR(_, satype, nullable=True):
 
 @dt.dtype.register(OracleDialect_cx_oracle, sa.dialects.oracle.TIMESTAMP)
 def sa_oracle_TIMESTAMP(_, satype, nullable=True):
-    return dt.Timestamp(nullable=nullable)
+    if satype.timezone:
+        return dt.Timestamp(timezone="UTC", nullable=nullable)
```
nehanene15, line 144: Couldn't the timezone be different from 'UTC' here?

nj1973: The time zone is attached to an individual piece of data rather than to the column so whatever we put there is wrong. The other engines that already worked with time zones before this work (e.g. PostgreSQL and SQL Server) also use a 'UTC' literal as a default.

My understanding is that DVT will attempt to convert all data to UTC before comparing it (row and column comparisons) so this is not particularly wrong.

Having said that, we are also displaying that literal across the board in schema validation to signify that a timestamp is time zone aware. We could use a different literal such as `(TZ)` instead of the time zone?

nehanene15: Ah, I see this makes sense. I think 'UTC' works in this case since we do convert to UTC before comparing.

nehanene15 (approved):

Issue #907, opened 2023-07-18
Support Snowflake as supported DB type


nehanene15:
Support Snowflake as DB for column and row validations. The Ibis upgrade to 5.1.0 will enable usage of Snowflake as a connection type, although we need to test out validations and connectivity.

Issue #916, opened 2023-07-26
Snowflake: Support TIMESTAMP_TZ data type


nehanene15:
Snowflake TIMESTAMP with time zone causes the following error when running column/row validations:

```
sqlalchemy.exc.ProgrammingError: (snowflake.connector.errors.ProgrammingError) 001003 (42000): 01ade410-0504-af59-0067-8487000b0aae: SQL compilation error:
syntax error line 2 at position 541 unexpected 'AT'.
syntax error line 2 at position 544 unexpected 'TIME'.
[SQL: WITH t0 AS 
(SELECT t2."ID" AS "ID", t2."COL_INT8" AS "COL_INT8", t2."COL_INT16" AS "COL_INT16", t2."COL_INT32" AS "COL_INT32", t2."COL_INT64" AS "COL_INT64", t2."COL_DEC_20" AS "COL_DEC_20", t2."COL_DEC_38" AS "COL_DEC_38", t2."COL_DEC_10_2" AS "COL_DEC_10_2", t2."COL_FLOAT32" AS "COL_FLOAT32", t2."COL_FLOAT64" AS "COL_FLOAT64", t2."COL_VARCHAR_30" AS "COL_VARCHAR_30", t2."COL_CHAR_2" AS "COL_CHAR_2", t2."COL_STRING" AS "COL_STRING", t2."COL_DATE" AS "COL_DATE", t2."COL_DATETIME" AS "COL_DATETIME", t2."COL_TSTZ" AS "COL_TSTZ", CAST(t2."COL_TSTZ" AT TIME ZONE %(COL_TSTZ_1)s AS TIMESTAMP_NTZ) AS "cast_timestamp__COL_TSTZ" 
FROM pso_data_validator.public."DVT_CORE_TYPES" AS t2)
 SELECT count(*) AS "count", sum(t1."epoch_seconds__cast_timestamp__COL_TSTZ") AS "sum__epoch_seconds__cast_timestamp__COL_TSTZ" 
[...]

PR #921, opened 2023-07-27
feat: Snowflake support

nehanene15:
Closes Issue #907 

Support for all validation types for Snowflake. 
The TIMESTAMP WITH TIME ZONE data type is not yet supported, but tracked by #916 

This also adds support for case insensitive column names for row hash validation. For example, a user can pass in `-hash COL_A` on a table with column 'col_a' and the validation won't throw an error. This was added since Snowflake defaults to uppercase column names.


renzokuken (commented):
It looks like this field is split out in several areas of the stack to accommodate the lack of TZ support for snowflake. What is default behavior here? Does the value where reference originally was get split out by default?

renzokuken (approved):

PR #930, opened 2023-08-07
fix: Issues with validate column for time zoned timestamps

### LINKED ISSUE #917: validate column: TIMESTAMP WITH TIME ZONE fails for Oracle/BigQuery column validation
This is a spin off of issue https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/706 which dealt with schema validation.

For this issue we should focus on column validation and then review whether row validation can be included in this issue or needs to be an issue of its own.

Example using Oracle/BigQuery but all engines need to be checked as many will have the same problems. We can test using the core data types integration test table (https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/tests/resources/oracle_test_tables.sql), the earliest time in the COL_TSTZ column in this table is `1970-01-01 00:00:01 -01:00`

Test command:
```
data-validation validate column -sc oracle -tc bq -tbls pso_data_validator.dvt_core_types --min=col_tstz

╒═══════════════════╤═══════════════════╤═══════════════════════════════════╤══════════════════════╤═══════════════════════════╤═══════════════════════════╤══════════════════╤═════════════════════╤══════════════════════════════════════╕
│ validation_name   │ validation_type   │ source_table_name                 │ source_column_name   │ source_agg_value          │ target_agg_value          │   pct_difference │ validation_status   │ run_id                               │
╞═══════════════════╪═══════════════════╪═══════════════════════════════════╪══════════════════════╪═══════════════════════════╪═══════════════════════════╪══════════════════╪═════════════════════╪══════════════════════════════════════╡
│ min__col_tstz     │ Column            │ pso_data_validator.dvt_core_types │ col_tstz             │ 1970-01-01 00:00:01+00:00 │ 1969-12-31 23:00:01+00:00 │          -360000 │ fail                │ e6a48a16-bb09-4974-ac84-045a84c390d8 │
├───────────────────┼───────────────────┼───────────────────────────────────┼──────────────────────┼───────────────────────────┼───────────────────────────┼──────────────────┼─────────────────────┼──────────────────────────────────────┤
│ count             │ Column            │ pso_data_validator.dvt_core_types │                      │ 3                         │ 3                         │                0 │ success             │ e6a48a16-bb09-4974-ac84-045a84c390d8 │
╘═══════════════════╧═══════════════════╧═══════════════════════════════════╧══════════════════════╧═══════════════════════════╧═══════════════════════════╧══════════════════╧═════════════════════╧══════════════════════════════════════╛
```

Notice how the BigQuery value has been adjusted to UTC but the Oracle value has not, we've lost the time zone.

nj1973:
The change fixes Oracle TIMESTAMP WITH TIME ZONE validation and adds column validation tests for:

- Oracle
- PostgreSQL
- SQL Server
- Teradata
- MySQL

### DIFF
--- a/tests/system/data_sources/test_postgres.py
+++ b/tests/system/data_sources/test_postgres.py
@@ -616,7 +616,6 @@ def test_column_validation_core_types():
 )
 def test_column_validation_core_types_to_bigquery():
     parser = cli_tools.configure_arg_parser()
-    # TODO Change --min/max strings below to include col_tstz when issue-917 is complete.
     # We've excluded col_float32 because BigQuery does not have an exact same type and float32/64 are lossy and cannot be compared.
     # TODO Change --sum and --max options to include col_char_2 when issue-842 is complete.
     args = parser.parse_args(
@@ -628,8 +627,8 @@ def test_column_validation_core_types_to_bigquery():
             "-tbls=pso_data_validator.dvt_core_types",
             "--filter-status=fail",
             "--sum=col_int8,col_int16,col_int32,col_int64,col_dec_20,col_dec_38,col_dec_10_2,col_float64,col_varchar_30,col_string,col_date,col_datetime",
-            "--min=col_int8,col_int16,col_int32,col_int64,col_dec_20,col_dec_38,col_dec_10_2,col_float64,col_varchar_30,col_string,col_date,col_datetime",
-            "--max=col_int8,col_int16,col_int32,col_int64,col_dec_20,col_dec_38,col_dec_10_2,col_float64,col_varchar_30,col_string,col_date,col_datetime",
+            "--min=col_int8,col_int16,col_int32,col_int64,col_dec_20,col_dec_38,col_dec_10_2,col_float64,col_varchar_30,col_string,col_date,col_datetime,col_tstz",
+            "--max=col_int8,col_int16,col_int32,col_int64,col_dec_20,col_dec_38,col_dec_10_2,col_float64,col_varchar_30,col_string,col_date,col_datetime,col_tstz",
         ]
     )
     config_managers = main.build_config_managers_from_args(args)
--- a/tests/system/data_sources/test_postgres.py
+++ b/tests/system/data_sources/test_postgres.py
@@ -626,7 +626,7 @@ def test_column_validation_core_types_to_bigquery():
             "-tc=bq-conn",
             "-tbls=pso_data_validator.dvt_core_types",
             "--filter-status=fail",
-            "--sum=col_int8,col_int16,col_int32,col_int64,col_dec_20,col_dec_38,col_dec_10_2,col_float64,col_varchar_30,col_string,col_date,col_datetime",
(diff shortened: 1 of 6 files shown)


tests/system/data_sources/test_sql_server.py
```diff
@@ -356,8 +355,8 @@ def test_column_validation_core_types_to_bigquery():
             "-tbls=pso_data_validator.dvt_core_types",
             "--filter-status=fail",
             "--sum=col_int8,col_int16,col_int32,col_int64,col_float64,col_date,col_datetime,col_dec_10_2,col_dec_20,col_dec_38,col_varchar_30,col_char_2,col_string",
```
nehanene15, line 357: Should we add col_tstz to the --sum flag as well?

tests/system/data_sources/test_postgres.py
```diff
@@ -628,8 +627,8 @@ def test_column_validation_core_types_to_bigquery():
             "-tbls=pso_data_validator.dvt_core_types",
             "--filter-status=fail",
             "--sum=col_int8,col_int16,col_int32,col_int64,col_dec_20,col_dec_38,col_dec_10_2,col_float64,col_varchar_30,col_string,col_date,col_datetime",
```
nehanene15: Should we include col_tstz in sum here as well?

nj1973: Yes, I've added it and the test passes.

tests/system/data_sources/test_sql_server.py
nj1973: Hmm, this is a tricky one.
I added the column and the test fails but it's not clear what the best course of action is.
If I compare SQL Server to BigQuery then we get an epoch seconds mismatch:
- SQL Server: 259206
- BigQuery: 280806

However if I compare SQL Server to itself then the value is 280806, which matches BigQuery.

The problem stems from these lines in `config_manager.py`:

```
        elif column_type == "timestamp" or column_type == "!timestamp":
            if (
                self.source_client.name == "bigquery"
                or self.target_client.name == "bigquery"
            ):
                calc_func = "cast"
                cast_type = "timestamp"
```

They are saying that if the source or target is BigQuery then cast both source and target to timestamp before doing the epoch seconds expression. The cast to timestamp is cropping the time zone in SQL Server.

At first I thought I needed to change the SQL Server cast to first convert to UTC but I also think it is wrong that a BigQuery requirement changes what we do on the other engine. If BigQuery needs a pre-cast to TIMESTAMP then it probably shouldn't be catered for here. Although I haven't looked into how hard it is to do that instead.

This needs more investigation.

nehanene15: Yeah agreed - seeing something very similar in Teradata. Since we cast to timestamp, we lose the timezone and cause a mismatch when validating col_tstz for TD to BQ. 

TD to TD also produces the correct 280806 value.

nehanene15 (approved):
LGTM!

tests/system/data_sources/test_sql_server.py
nj1973: I've created https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/938 for this problem.

PR #941, opened 2023-08-10
feat: support timestamp aggregation for Oracle and TD

### LINKED ISSUE #762: validate column: Sum on Oracle and Teradata TIMESTAMP Throws Exception
It's fair to day that a SUM on a date based column doesn't make a lot of sense but we support it on other engines and it will often get triggered through columns being provided using `*`.

Test table:
```
CREATE TABLE dvt_test.tab_ts
(   id              NUMBER(8) NOT NULL PRIMARY KEY
,   col_datetime    TIMESTAMP(3)
,   col_tstz        TIMESTAMP(3) WITH TIME ZONE
);

INSERT INTO dvt_test.tab_ts VALUES
(1,TIMESTAMP'1970-01-01 00:00:01'
,to_timestamp_tz('1970-01-01 00:00:01 -01:00','YYYY-MM-DD HH24:MI:SS TZH:TZM'));
COMMIT;
```

Test command:
```
data-validation validate column \
-sc ora_conn -tc ora_conn \
-tbls=dvt_test.tab_ts --sum='*' --min='*' --max='*'
```

Exception:
```
sqlalchemy.exc.DatabaseError: (cx_Oracle.DatabaseError) ORA-00907: missing right parenthesis
[SQL: SELECT count(:count_1) AS count, sum(t0.id) AS sum__id, sum(t0.epoch_seconds__col_datetime) AS sum__epoch_seconds__col_datetime, sum(t0.epoch_seconds__col_tstz) AS sum__epoch_seconds__col_tstz, min(t0.id) AS min__id, min(t0.col_datetime) AS min__col_datetime, min(t0.col_tstz) AS min__col_tstz, max(t0.id) AS max__id, max(t0.col_datetime) AS max__col_datetime, max(t0.col_tstz) AS max__col_tstz
FROM (SELECT t1.id AS id, t1.col_datetime AS col_datetime, t1.col_tstz AS col_tstz, CAST(EXTRACT(epoch FROM t1.col_datetime) AS NUMBER(19)) AS epoch_seconds__col_datetime, CAST(EXTRACT(epoch FROM t1.col_tstz) AS NUMBER(19)) AS epoch_seconds__col_tstz
FROM dvt_test.tab_ts t1) t0]
[parameters: {'count_1': '*'}]
```

The problem SQL expressions are:
```
CAST(EXTRACT(epoch FROM t1.col_datetime) AS NUMBER(19)) AS epoch_seconds__col_datetime,
CAST(EXTRACT(epoch FROM t1.col_tstz) AS NUMBER(19)) AS epoch_seconds__col_tstz
```

`EXTRACT(epoch …)` is invalid syntax on Oracle.

When complete we need to enable the integration test for this problem. Search for "issue-762" in `tests/system/data_sources/test_oracle.py` and `tests/system/data_sources/test_teradata.py` to action this. Enabling this test is blocked by https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/759.

nehanene15:
Closes #762 

This removes timestamp from being included in wildcard aggregations by default.

Adds new flag `--wildcard-include-timestamp` to include timestamp columns in wildcard aggregations i.e. `--sum='*'`. This will calculate the SUM(UNIX_SECONDS(ts_column))

When manually specifying `--sum ts_column`, the aggregation will run over SUM(UNIX_SECONDS(ts_column)).

### DIFF
--- a/third_party/ibis/ibis_oracle/registry.py
+++ b/third_party/ibis/ibis_oracle/registry.py
@@ -391,7 +391,7 @@ def _random(t, op):
 
 def _extract_epoch(t, op):
     sub = operator.sub(sa.cast(t.translate(op.arg), sa.DATE), sa.sql.literal_column("DATE '1970-01-01'")).self_group()
-    mult = sa.cast(operator.mul(sub, sa.sql.literal_column("86400")), sa.NUMERIC)
+    mult = sa.cast(operator.mul(sub, sa.sql.literal_column("86400")), sa.dialects.oracle.NUMBER)
     return mult
 
 
--- a/third_party/ibis/ibis_oracle/registry.py
+++ b/third_party/ibis/ibis_oracle/registry.py
@@ -389,6 +389,11 @@ def _day_of_week_name(t, op):
 def _random(t, op):
     return sa.sql.literal_column("DBMS_RANDOM.VALUE")
 
+def _extract_epoch(t, op):
+    sub = operator.sub(sa.cast(t.translate(op.arg), sa.DATE), sa.sql.literal_column("DATE '1970-01-01'")).self_group()
+    mult = sa.cast(operator.mul(sub, sa.sql.literal_column("86400")), sa.NUMERIC)
+    return mult
+
 
 operation_registry.update(
     {
@@ -442,12 +447,7 @@ operation_registry.update(
         ops.ExtractDay: _extract('day'),
         ops.ExtractDayOfYear: _extract('doy'),
         ops.ExtractQuarter: _extract('quarter'),
-        ops.ExtractEpochSeconds: fixed_arity(
-            lambda x: sa.cast(
-                sa.func.datediff(sa.text('s'), '1970-01-01 00:00:00', x), sa.BIGINT
-            ),
-            1,
-        ),
+        ops.ExtractEpochSeconds: _extract_epoch,
         ops.ExtractHour: _extract('hour'),
         ops.ExtractMinute: _extract('minute'),
         ops.ExtractSecond: _second,
--- a/third_party/ibis/ibis_teradata/datatypes.py
+++ b/third_party/ibis/ibis_teradata/datatypes.py
@@ -170,13 +170,6 @@ class TeradataTypeTranslator(object):
             )
 
         return "TIMESTAMP WITH TIME ZONE"
-    
-    @classmethod
-    def to_ibis_from_TZ(cls, col_data, return_ibis_type=True):
-        if return_ibis_type:
-            return dt.timestamp(timezone="UTC")
-
-        return "TIMESTAMP WITH TIME ZONE"
 
 
 ibis_type_to_teradata_type = Dispatcher("ibis_type_to_teradata_type")
--- a/third_party/ibis/ibis_teradata/datatypes.py
+++ b/third_party/ibis/ibis_teradata/datatypes.py
@@ -159,7 +159,14 @@ class TeradataTypeTranslator(object):
         if return_ibis_type:
             return dt.timestamp(timezone="UTC")
 
-        return "TIMESTAMP"
+        return "TIMESTAMP WITH TIME ZONE"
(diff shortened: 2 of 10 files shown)


third_party/ibis/ibis_oracle/registry.py
```diff
@@ -389,6 +389,11 @@ def _day_of_week_name(t, op):
 def _random(t, op):
     return sa.sql.literal_column("DBMS_RANDOM.VALUE")
 
+def _extract_epoch(t, op):
+    sub = operator.sub(sa.cast(t.translate(op.arg), sa.DATE), sa.sql.literal_column("DATE '1970-01-01'")).self_group()
+    mult = sa.cast(operator.mul(sub, sa.sql.literal_column("86400")), sa.NUMERIC)
```
nj1973: You might feel that this doesn't matter but I'll flag it anyway.

Using `sa.NUMERIC` here gives a final SQL expression of:

```
CAST((CAST(t0.cast_timestamp__col_tstz AS DATE) - DATE '1970-01-01') * 86400 AS NUMERIC)
```

`NUMERIC` works because it is an ANSI data type name. Oracle developers would expect to see `NUMBER` in their expressions.

```
mult = sa.cast(operator.mul(sub, sa.sql.literal_column("86400")), sa.dialects.oracle.NUMBER)
```

Giving:

```
CAST((CAST(t0.cast_timestamp__col_tstz AS DATE) - DATE '1970-01-01') * 86400 AS NUMBER)
```

Having typed that out I feel it's not a big deal. But I'll leave it here anyway for you to read and ignore :-)

third_party/ibis/ibis_teradata/datatypes.py
```diff
@@ -169,7 +169,14 @@ def to_ibis_from_SZ(cls, col_data, return_ibis_type=True):
                 timezone="UTC", nullable=cls._col_data_nullable(col_data)
             )
 
-        return "TIMESTAMP"
+        return "TIMESTAMP WITH TIME ZONE"
+    
+    @classmethod
+    def to_ibis_from_TZ(cls, col_data, return_ibis_type=True):
```
nj1973: Teradata type `TZ` is `TIME WITH TIME ZONE`, not `TIMESTAMP`. Is there a reason for this new function? 

The core data types test table doesn't have a time column because not all engines support it. I've never tested DVT with one, I wonder if we need a new test table. Not for this issue but for a future issue.

third_party/ibis/ibis_teradata/registry.py
```diff
@@ -226,6 +226,24 @@ def _string_join(translator, op):
     return "||".join(map(translator.translate, op.arg))
 
 
+def _extract_epoch(translator, op):
+    arg = translator.translate(op.arg)
+    if op.arg.output_dtype.timezone:
```
nj1973: Minor thought but this might be more concise as:
```
utc_expression = "AT TIME ZONE 'GMT'" if op.arg.output_dtype.timezone else ""
return (
    f"(CAST({arg} AS DATE {utc_expression}) - DATE '1970-01-01') * 86400 + "
    f"(EXTRACT(HOUR FROM {arg} {utc_expression}) * 3600) + "
    f"(EXTRACT(MINUTE FROM {arg} {utc_expression}) * 60) + "
    f"(EXTRACT(SECOND FROM {arg} {utc_expression}))"
)
```
Minor enough to ignore though.

nj1973 (approved):
Left two very minor nit-picky comments. I'll approve anyway because LGTM and leave the comments as your decision whether to act on or ignore.

third_party/ibis/ibis_teradata/datatypes.py
nehanene15: Ah, good catch. I'll remove this - I thought it was another abbreviation for `TIMESTAMP WITH TIME ZONE`

nj1973 (approved):
LGTM

Issue #923, opened 2023-07-31
`generate-table-partitions` generates invalid SQL in filter clause when a timestamp field is part of the primary-key


florisvink:
The following command:

```
data-validation generate-table-partitions -sc rs -tc bq -tbls source.table=target.table --primary-keys _hash_key,_hash_diff,_creation_timestamp,source_last_update_date,delivery_area_id --partition-num 164 --filter-status fail --bq-result-handler pso_data_validator.results -cdir ~/
```

generates 164 partitioned configs as expected but the none of those can be run using `data-validation configs run -c ~/config_X.yaml` throwing the follow:

```
07/31/2023 09:37:36 AM-ERROR: Error (psycopg2.errors.SyntaxError) syntax error at or near "10"
LINE 3: ...2f114e622b' AND (_creation_timestamp > 2023-05-16 10:48:06.8...
```

It seems the `filter: <sql_where_clause>` part of the YALM contains timestamp values without the required quotes (see `_creation_timestamp` below).

```
  filters:
  - source: (_hash_key > '04b0a057a5ceccd4fff06984b5ed8e0ba1e90591' OR _hash_key =
      '04b0a057a5ceccd4fff06984b5ed8e0ba1e90591' AND (_hash_diff > 'cba27aebaf720f51987e0d3087213bd50fc3a70c'
      OR _hash_diff = 'cba27aebaf720f51987e0d3087213bd50fc3a70c' AND (_creation_timestamp
      > 2023-04-18 10:39:12.808967 OR _creation_timestamp = 2023-04-18 10:39:12.808967 ...
```

helensilva14:
Hi @florisvink, thanks for reporting this! Could you please help us understand how much this is affecting you at the moment? We did a triage for this problem and we will tackle it as soon as we can, since we just merged some updates for this feature on #922.

sundar-mudupalli-work:
The problem is in two static methods ```_less_than_value``` and. ```_geq_value``` in  https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/data_validation/partition_builder.py. They both have the line
```
value0 = "'" + values[0] + "'" if isinstance(values[0], str) else str(values[0])
```
[...]

PR #945, opened 2023-08-17
fix: Change in filter tag creation in generate-table-partitions

piyushsarraf:
* Fixes issue #923.
* Changed the quotes to `" "` for string datatype, to handle quotes in string value.


sundar-mudupalli-work:
Piyush,

Thank you for working on this. I did some investigation and here are my findings (based on postgres)

1. Postgres has `time`, `date`, `timestamp` and `interval` date/time types and the `time` and `timestamp` type can be with or without timezone.
2. The `interval` type is not supported by ibis postgres extension - any table with a column with this type will throw an error. I will file a bug on this.
3. The `time` type shows up in pandas as an object. So I am not sure we can support this type as a primary key column
4. The `date` and `timestamp` type are both stored as datetime objects in pandas. While I used isinstance, I am now learning there is a better way to check the type. The recommended way for datetime type is to use [pandas.api.types.is_datetime64_any_dtype](https://pandas.pydata.org/docs/reference/api/pandas.api.types.is_datetime64_any_dtype.html) and for strings is [pandas.api.types.is_string_dtype](https://pandas.pydata.org/docs/reference/api/pandas.api.types.is_string_dtype.html)
5. So if you have a `date` object in pandas, you have to cast the key as timestamp, so like `"cast(" + keys[0] + "as timestamp)"`
6. While you are making this change, can you please check how things will work if a column is boolean ?
Thanks.

Sundar Mudupalli

sundar-mudupalli-work:
Hi,

This fix only support timestamp fields, but not date fields. We need a fix that supports both - as the customer has already reported the issue with the date field as well. Please see my comments on how to fix the date field.

Thanks.

Sundar Mudupalli

sundar-mudupalli-work:
Looking at the email trail, the customer reported 3 issues: 1. timestamp (this is bug you fixed), 2. Date (this is the bug they are editing the yaml to be cast(<keyname> as timestamp) and 3. escaping the quote character in the string. The third one is hard to fix and requires a redesign I think. Can you fix the second one as indicated in the notes to the PR. Thanks.

florisvink:
Many thanks for working on this! 
I also find that FLOAT/NUMERIC fields can cause issues as being part of a primary key since they can be `null` and they appear in the filters as `nan`

Example:
```
  filters:
  - source: (opportunity_id > '132252' OR opportunity_id = '132252' AND (stagename
      > 'In Progress' OR stagename = 'In Progress' AND (stage_change_datetime > '2023-01-02
      11:09:20' OR stage_change_datetime = '2023-01-02 11:09:20' AND (chain_name >
      '(Not available)' OR chain_name = '(Not available)' AND (amount >= nan)))))
      AND (opportunity_id < '134269' OR opportunity_id = '134269' AND (stagename <
      'Onboarding' OR stagename = 'Onboarding' AND (stage_change_datetime < '2020-06-29
      15:04:37' OR stage_change_datetime = '2020-06-29 15:04:37' AND (chain_name <
      '241 Pizza' OR chain_name = '241 Pizza' AND (amount < nan)))))
    target: (opportunity_id > '132252' OR opportunity_id = '132252' AND (stagename
      > 'In Progress' OR stagename = 'In Progress' AND (stage_change_datetime > '2023-01-02
      11:09:20' OR stage_change_datetime = '2023-01-02 11:09:20' AND (chain_name >
      '(Not available)' OR chain_name = '(Not available)' AND (amount >= nan)))))
      AND (opportunity_id < '134269' OR opportunity_id = '134269' AND (stagename <
      'Onboarding' OR stagename = 'Onboarding' AND (stage_change_datetime < '2020-06-29
      15:04:37' OR stage_change_datetime = '2020-06-29 15:04:37' AND (chain_name <
      '241 Pizza' OR chain_name = '241 Pizza' AND (amount < nan)))))
    type: custom
```
(see `(amount < nan)`)

data_validation/partition_builder.py
```diff
@@ -80,7 +81,13 @@ def _less_than_value(keys, values) -> str:
...
             String containing the expression - e.g. (birth_month < 5 OR (birth_month = 5 AND (birth_day <= 2)))
         """
-        value0 = "'" + values[0] + "'" if isinstance(values[0], str) else str(values[0])
+        if isinstance(values[0], str):
+            value0 = '"' + values[0] + '"'
+        elif isinstance(values[0], pd.Timestamp):
+            value0 = "'" + str(values[0]) + "'"
+        else:
+            value0 = str(values[0])
+
```
florisvink: double-quotes (`"`) seem to cause problems in RedShift: 

```
< SELECT count(*) FROM some_table WHERE some_field > 'some_value'
> OK

< SELECT count(*) FROM some_table WHERE some_field > "some_value"
> psycopg2.errors.UndefinedColumn: column "some_value" does not exist in some_table

```

florisvink (changes requested):

sundar-mudupalli-work:
> Many thanks for working on this! I also find that FLOAT/NUMERIC fields can cause issues as being part of a primary key since they can be `null` and they appear in the filters as `nan`
> 
> Example:
> 
> ```
>   filters:
>   - source: (opportunity_id > '132252' OR opportunity_id = '132252' AND (stagename
>       > 'In Progress' OR stagename = 'In Progress' AND (stage_change_datetime > '2023-01-02
>       11:09:20' OR stage_change_datetime = '2023-01-02 11:09:20' AND (chain_name >
>       '(Not available)' OR chain_name = '(Not available)' AND (amount >= nan)))))
>       AND (opportunity_id < '134269' OR opportunity_id = '134269' AND (stagename <
>       'Onboarding' OR stagename = 'Onboarding' AND (stage_change_datetime < '2020-06-29
>       15:04:37' OR stage_change_datetime = '2020-06-29 15:04:37' AND (chain_name <
>       '241 Pizza' OR chain_name = '241 Pizza' AND (amount < nan)))))
>     target: (opportunity_id > '132252' OR opportunity_id = '132252' AND (stagename
>       > 'In Progress' OR stagename = 'In Progress' AND (stage_change_datetime > '2023-01-02
>       11:09:20' OR stage_change_datetime = '2023-01-02 11:09:20' AND (chain_name >
>       '(Not available)' OR chain_name = '(Not available)' AND (amount >= nan)))))
>       AND (opportunity_id < '134269' OR opportunity_id = '134269' AND (stagename <
>       'Onboarding' OR stagename = 'Onboarding' AND (stage_change_datetime < '2020-06-29
>       15:04:37' OR stage_change_datetime = '2020-06-29 15:04:37' AND (chain_name <
>       '241 Pizza' OR chain_name = '241 Pizza' AND (amount < nan)))))
>     type: custom
> ```
> 
> (see `(amount < nan)`)

Floris,

It is strange that the database is allowing you to store null values in a primary key. Which database is this?

Per Oracle (and likely other databases) - `A primary key defines the set of columns that uniquely identifies rows in a table. When you create a primary key constraint, none of the columns included in the primary key can have NULL constraints; that is, they must not permit NULL values.`. Please see [Oracle docs](https://docs.oracle.com/javadb/10.10.1.2/ref/rrefsqlj13590.html#:~:text=A%20primary%20key%20defines%20the,must%20not%20permit%20NULL%20values.)

Sundar Mudupalli

sundar-mudupalli-work (changes requested):
Lines 87 -90 read
``` 
      elif isinstance(values[0], pd.Timestamp):
            if primary_key != "cast(" + keys[0] + " as timestamp)":
                primary_key = "cast(" + keys[0] + " as timestamp)"
            value0 = "'" + str(values[0]) + "'"
```
Should instead be
```
      elif isinstance(values[0], pd.Timestamp):
            primary_key = "cast(" + keys[0] + " as timestamp)"
            value0 = "'" + str(values[0]) + "'"
```
It is possible i am misunderstanding your code - so help me here. Similar comment on lines 122-125. 

The customer is also saying that the use of `"` does not work correctly in RedShift. So the following line (121) may create a problem.
```
          value0 = '"' + values[0] + '"'
```
Perhaps not use `"` and use `'` instead ?

Also you will need to build a test to demonstrate that primary date and timestamp keys work correctly - I suggest you create a function very similar to `test_postgres_generate_table_partitions` for each of the 8 [data sources](https://github.com/GoogleCloudPlatform/professional-services-data-validator/tree/develop/tests/system/data_sources).  I left the SQL code to generate the table for the postgres test in [postgres_test_tables.sql](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/tests/resources/postgresql_test_tables.sql). You can do something similar for all 8 tests.


Thank you.

Sundar Mudupalli

florisvink:
> It is strange that the database is allowing you to store null values in a primary key. Which database is this?

Hi @sundar-mudupalli-work. In this case the null value originates from a column that is part of a composite (combined) primary key. Apart from that: In BigQuery (and I think in most analytics databases) a unique primary key is not a necessity (other than relations databases where indeed primary keys have a unique constraint)

nehanene15:
@sundar-mudupalli-work @piyushsarraf I think rather than adding integration tests, this would be better suited for unit tests for the function PartitionBuilder. _less_than_value(). It won't be scalable to create a new integration test + table for every unique primary key data type.

@florisvink As for the 'nan' issue  if the PK were to be 'nan' would you want the entire condition `AND (amount < nan)` to be removed? In that case we can return an empty string "" from the _less_than_value function? Feel free to comment or add your suggestions to this PR.

florisvink:
> As for the 'nan' issue if the PK were to be 'nan' would you want the entire condition `AND (amount < nan)` to be removed? In that case we can return an empty string "" from the _less_than_value function? Feel free to comment or add your suggestions to this PR.

Yeah I think that makes sense. It could to unexpected results (partitions with 'too' many rows..) but at least it's better than creating corrupt SQL. 

Please not it's not only 'greater than' statement. I also see filters like: `(userid < nan OR userid = nan AND (restaurantid
 < '1204085'))`

sundar-mudupalli-work:
@florisvink  - let us move the Null / Nan discussion to [Issue 951](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/951) opened specifically for that issue - I have responded there.

Please keep other formatting issues - Date, Timestamps, String etc - here - we are working on resolving them ASAP.

Thank you for your patience.

Sundar Mudupalli

florisvink:
Apart from being able to apply a filter to partition generation it would also be helpful to be able to typecast fields used in the primary key before the partitions gets generated. As far as I know these settings are only editable in the YAML, but not as CLI arguments running `generate-table-partitions`

I often see errors like `cannot cast "2023-02-15 00:00.00" literal to date` while validating generated partitions. These errors originate from the partition-filter in the YAML.

PR #953, opened 2023-08-22
feat: Add more mappings to the allowlist configuration file oracle_to_bigquery.yaml

helensilva14:
Mappings identified when performing new tests for Oracle to BQ validations.


nj1973:
I've tweaked the additions but this has highlighted a code issue we need to resolve. The decimal precision/scale expansion doesn't work for not-null columns since we added the `!` marker for not null.
I've also added a unit test to prove this isn't working.
I'll add a fix for it today.

nj1973 (approved):
LGTM

helensilva14:
Thanks a lot Neil!

PR #957, opened 2023-08-24
fix: Add exception handling for invalid value to cast a comparison field 

### LINKED ISSUE #948: weird error message on wrong cast type (`datetime`) in config
`datetime` is not a valid value for `cast:` in the YAML config. Normally you can a clear error like:

```
...
validations:
- calculated_fields: []
  comparison_fields:
  - cast: some_invalid_cast_type
    field_alias: dd
    source_column: dd
    target_column: dd
...
```
..results in the error: `PM-ERROR: Error expected one of '!', 'array', 'bigdecimal', 'bignumeric', 'binary', 'bool', 'boolean', 'bytes', 'char', 'date', 'decimal', 'double', 'float', 'float16', 'float32', 'float64', 'geography', 'geometry', 'halffloat', 'inet', 'int', 'int16', 'int32', 'int64', 'int8', 'interval', 'json', 'linestring', 'macaddr', 'map', 'multilinestring', 'multipoint', 'multipolygon', 'null', 'point', 'polygon', 'set', 'str', 'string', 'struct', 'time', 'timestamp', 'uint16', 'uint32', 'uint64', 'uint8', 'uuid', 'varchar'`

However if you accidentally put `datetime` in the cast you field like:
```
...
validations:
- calculated_fields: []
  comparison_fields:
  - cast: datetime
    field_alias: dd
    source_column: dd
    target_column: dd
...
```
you get the weird error: ` Error expected 'EOF'`.

helensilva14:
Closes #948 

/gcbrun

### DIFF
--- a/data_validation/query_builder/query_builder.py
+++ b/data_validation/query_builder/query_builder.py
@@ -203,8 +203,8 @@ class ComparisonField(object):
         if self.cast:
             try:
                 comparison_field = comparison_field.force_cast(self.cast)
-            except:
-                raise SyntaxError("FOI AQUI ÒOOOOOO")
+            except Exception as e:
+                raise SyntaxError(f"'{self.cast}' is an invalid data type to cast a comparison field ({e})")
         comparison_field = comparison_field.name(alias)
 
         return comparison_field
--- a/data_validation/query_builder/query_builder.py
+++ b/data_validation/query_builder/query_builder.py
@@ -201,7 +201,10 @@ class ComparisonField(object):
         comparison_field = ibis_table[self.field_name]
         alias = self.alias or self.field_name
         if self.cast:
-            comparison_field = comparison_field.force_cast(self.cast)
+            try:
+                comparison_field = comparison_field.force_cast(self.cast)
+            except:
+                raise SyntaxError("FOI AQUI ÒOOOOOO")
         comparison_field = comparison_field.name(alias)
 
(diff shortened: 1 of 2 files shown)


data_validation/query_builder/query_builder.py
```diff
@@ -201,7 +201,10 @@ def compile(self, ibis_table):
         comparison_field = ibis_table[self.field_name]
         alias = self.alias or self.field_name
         if self.cast:
-            comparison_field = comparison_field.force_cast(self.cast)
+            try:
+                comparison_field = comparison_field.force_cast(self.cast)
+            except Exception as e:
```
nehanene15: We probably should catch the exception here since the 'force_cast()' method could throw a wide number of exceptions and we're assuming it's an invalid data type cast.

Instead let's add a try/except [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/third_party/ibis/ibis_addon/api.py#L46)

```
try:
        op = ops.Cast(self, to=target_type)
    except parsy.ParseError:
        raise SyntaxError(f"Invalid datatype provided in ComparisonField cast to {target_type}")
```

This will directly catch the ParseError that Ibis throws when a data type is invalid.

nehanene15 (changes requested):

nehanene15 (approved):
LGTM once tests pass!

PR #959, opened 2023-08-28
fix: Add missing SnowflakeDialect mapping for BINARY data type

### LINKED ISSUE #958: Snowflake and BINARY datatype
Trying to run data validation to Snowflake. Target table has fields with datatype BINARY and the following error occurs:
`Error Could not find signature for dtype: <SnowflakeDialect, BINARY> occurred while running config file example.yaml. Skipping it for now.`

Or without yaml-file:
`NotImplementedError: Could not find signature for dtype: <SnowflakeDialect, BINARY>`

helensilva14:
Closes #958 

/gcbrun

### DIFF
--- a/third_party/ibis/ibis_snowflake/datatypes.py
+++ b/third_party/ibis/ibis_snowflake/datatypes.py
@@ -18,7 +18,7 @@ import sqlalchemy as sa
 from ibis.backends.snowflake import Backend as SnowflakeBackend
 from ibis.backends.snowflake.datatypes import parse
 from snowflake.connector.constants import FIELD_ID_TO_NAME
-from snowflake.sqlalchemy import NUMBER
+from snowflake.sqlalchemy import NUMBER, BINARY
 from snowflake.sqlalchemy.snowdialect import SnowflakeDialect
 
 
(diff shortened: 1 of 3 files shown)


third_party/ibis/ibis_snowflake/datatypes.py
```diff
@@ -30,6 +30,10 @@ def sa_sf_numeric(_, satype, nullable=True):
         nullable=nullable,
     )
 
+@dt.dtype.register(SnowflakeDialect, BINARY)
```
nehanene15, line 33: This looks great! Should we also add support for other data types that may not be added i.e DATE, TEXT, BOOLEAN, etc? 
Unless they're already supported

helensilva14: Good point! I created the below table to make sure all these types listed [here on Ibis](https://github.com/ibis-project/ibis/blob/5.1.0/ibis/backends/snowflake/datatypes.py#L22) worked and and they're all supported!

```
create or replace table test_specific_data_types (
    --col_fixed fixed,
    col_real real,
    col_text text,
    col_variant variant,
    col_object object,
    col_array array,
    col_binary binary,
    col_time time,
    col_boolean boolean
);
```
image

helensilva14: Added a unit test for this scenario!

nehanene15 (approved):
LGTM

PR #960, opened 2023-08-28
feat: Support BQ decimal precision and scale for schema validation

### LINKED ISSUE #882: Schema Validation : Bigquery Numeric
BigQuery NUMERIC columns are being read by schema validation as decimal(38, 9). 

Example BigQuery column:

   col_1      NUMERIC(4,3)

Shows in DVT schema validation as **decimal(38, 9)**.

Since numeric precision in Ibis is fixed at 38 & scale 9, schema validation for Numeric columns with other sources will be a FAIL.

This issue is similar to #839

### LINKED ISSUE #883: Schema validation: BigQuery Decimal Precision
BigQuery Decimal columns are being read by schema validation as decimal(38, 9) i.e. fixed to precision 38, scale 9

Example BigQuery column:

,   col_dec_38      Decimal(4,3)

Shows in DVT schema validation as decimal(38, 9) 

This results in schema validation fail due to precision mismatch in source & target

https://github.com/GoogleCloudPlatform/professional-services-data-validator/assets/129146107/2a19dd1b-db03-4864-849a-602d6bbbe4eb)

This issue is similar to #839

nehanene15:
- Limit Teradata query to one row for custom query schema generation (Avoids full table scan)
- Support non-defualt BigQuery precision/scale (Closes #883 and Closes #882 )

### DIFF
--- a/third_party/ibis/ibis_teradata/__init__.py
+++ b/third_party/ibis/ibis_teradata/__init__.py
@@ -141,7 +141,7 @@ class Backend(BaseSQLBackend):
         return sch.Schema(schema)
 
     def _get_schema_using_query(self, query):
-        cur = self.raw_sql(query)
+        cur = self.raw_sql(f"SELECT TOP 1 * FROM ({query}) AS t0")
         # resets the state of the cursor and closes operation
         cur.fetchall()
         ibis_fields = self._adapt_types(cur.description)
--- a/tests/system/data_sources/test_hive.py
+++ b/tests/system/data_sources/test_hive.py
@@ -166,10 +166,6 @@ def test_schema_validation_core_types_to_bigquery():
             (
                 # All Hive integrals go to BigQuery INT64.
                 "--allow-list=int8:int64,int16:int64,int32:int64,"
-                # Hive decimals that map to BigQuery NUMERIC.
-                "decimal(20,0):decimal(38,9),decimal(10,2):decimal(38,9),"
-                # Hive decimals that map to BigQuery BIGNUMERIC.
-                "decimal(38,0):decimal(76,38),"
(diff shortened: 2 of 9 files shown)


third_party/ibis/ibis_teradata/__init__.py
```diff
@@ -141,7 +141,7 @@ def get_schema(self, table_name: str, database: str = None) -> sch.Schema:
         return sch.Schema(schema)
 
     def _get_schema_using_query(self, query):
-        cur = self.raw_sql(query)
+        cur = self.raw_sql(f"SELECT TOP 1 * FROM ({query}) AS t0")
```
nj1973, line 144: We might also have this problem for other engines, e.g: Hive:
```
def _get_schema_using_query(self, query):
    # Removing LIMIT 0 around query since it returns no results in Hive
    cur = self.raw_sql(query)
    cur.fetchall()
    ibis_fields = self._adapt_types(cur.description)
    cur.release()

    return sch.Schema(ibis_fields)
```

We should be able to use `fetchone` in place of `fetchall` for this. Or embed in a subquery.  I didn't check all engines so Hive may not be the only one.

Happy for this to be spun into another issue if you need to move quickly on this one.

nj1973 (approved):
LGTM aside from comment about Hive (and maybe other engines) having the full scan issue fixed here for Teradata.

third_party/ibis/ibis_teradata/__init__.py
nehanene15: Good catch! I'll update this for Hive and add a `LIMIT 1` similar to the Ibis impala connection. Spanner only materializes one row, and all other DBs are handled by Ibis.

Issue #880, opened 2023-06-09
Support column 'except' list for aggregations and hash


nehanene15:
Currently DVT only supports specifying columns to _include_ for aggregations or hashes i.e
`data-validation validate column ... -sum 'col_a,col_b,col_c,col_d'`

At times it is easier to provide an exclusion list instead to sum all columns _except_ col_e for example. 
DVT should support a flag that signals the column list is an exclusion list rather than an inclusion list (default).

For example,
`data-validation validate column ... -sum col_e --exclude-columns`

`--exclude-columns or -ec` can be a [store_false argument](https://docs.python.org/3/library/argparse.html#action). If the flag is present, we take all columns except the ones specified.

sudheersingampalli:
Hi @nehanene15 , I would like to work on this feature..

nehanene15:
@sudheersingampalli Is this still something you would like to work on? If not, I will reassign

sundar-mudupalli-work:
Hi,

I reviewed @mloeberc's code changes for this and this feature request requires more clarification

Two issues I see:
1. what if I specify `--sum '*' --exclude-columns` - this should raise ValueError
2. what if I specify `--sum num_bikes_available --exclude-columns --avg num_bikes_available,num_docks_disabled` - does the `--exclude-columns` apply to the `sum`, `avg` or both ? OR should we disallow --exclude-columns when there is more than one aggregation ?

On the second one, I don't have a strong opinion.

Thanks.

mloeberc:
Sundar, I'll wait for someone to answer to the first issue, and test what is the actual behavior.
As for the second issue, we discussed this in one meeting and the conclusion was when the --exclude-columns is present it will apply to all the aggregations mentioned in the command, since providing a list to include an another one to exclude would add complexity to the instructions given in a command, so it would be an apply to all flag.

sundar-mudupalli-work:
Hi,

Ok, I thought about what I said above  - if I specify `--sum '*' --exclude-columns` - this should raise ValueError.

I am reversing myself on the statement that `--sum '*' --exclude-columns` should not raise ValueError - instead `--exclude-columns` must be ignored for this aggregation. Here is why I think so:
1.  The `validate column` command always includes a `--count '*'` as an aggregation. Therefore, if we raised ValueError, every use of `--exclude-columns` would raise a ValueError.
[...]

PR #961, opened 2023-08-30
feat: Adding Exclude columns flag for aggregations in column validations

mloeberc:
Closes issue #880 

* Added parameter --exclude-columns or -ec to column validation to get a flag that identifies the column list as an exclusion, and not an inclusion.
* Updated Read Me to document the new parameter.
* Modified logic in Config Manager to process the column list accordingly to exclusion or inclusion commands.
* Added unit test to test exclusion configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,7 @@ data-validation (--verbose or -v) (--log-level or -ll) validate column
   [--max COLUMNS]       Comma separated list of columns for max or * for all numeric
   [--avg COLUMNS]       Comma separated list of columns for avg or * for all numeric
   [--std COLUMNS]       Comma separated list of columns for stddev_samp or * for all numeric
+  [--exclude-columns or -ec]   Flag to indicate the list of columns provided should be excluded and not included.
   [--bq-result-handler or -bqrh PROJECT_ID.DATASET.TABLE]
                         BigQuery destination for validation results. Defaults to stdout.
                         See: *Validation Reports* section
@@ -310,8 +311,8 @@ data-validation (--verbose or -v) (--log-level or -ll) validate custom-query col
   [--min COLUMNS]       Comma separated list of columns for min or * for all numeric
   [--max COLUMNS]       Comma separated list of columns for max or * for all numeric
   [--avg COLUMNS]       Comma separated list of columns for avg or * for all numeric
-  [--exclude-columns or -ec]   Flag to indicate the list of columns provided should be excluded and not included.
   [--std COLUMNS]       Comma separated list of columns for stddev_samp or * for all numeric
+  [--exclude-columns or -ec]   Flag to indicate the list of columns provided should be excluded and not included.
   [--bq-result-handler or -bqrh PROJECT_ID.DATASET.TABLE]
                         BigQuery destination for validation results. Defaults to stdout.
                         See: *Validation Reports* section
--- a/README.md
+++ b/README.md
@@ -311,6 +311,7 @@ data-validation (--verbose or -v) (--log-level or -ll) validate custom-query col
   [--min COLUMNS]       Comma separated list of columns for min or * for all numeric
   [--max COLUMNS]       Comma separated list of columns for max or * for all numeric
   [--avg COLUMNS]       Comma separated list of columns for avg or * for all numeric
+  [--exclude-columns or -ec]   Flag to indicate the list of columns provided should be excluded and not included.
   [--bq-result-handler or -bqrh PROJECT_ID.DATASET.TABLE]
                         BigQuery destination for validation results. Defaults to stdout.
                         See: *Validation Reports* section
--- a/data_validation/__main__.py
+++ b/data_validation/__main__.py
@@ -83,19 +83,18 @@ def get_aggregate_config(args, config_manager: ConfigManager):
         supported_data_types.extend(["timestamp", "!timestamp", "date", "!date"])
 
     cast_to_bigint = True if args.cast_to_bigint else False
-    exclude_cols = True if args.exclude_columns else False
 
     if args.count:
         col_args = None if args.count == "*" else cli_tools.get_arg_list(args.count)
         aggregate_configs += config_manager.build_config_column_aggregates(
-            "count", col_args, exclude_cols, None, cast_to_bigint=cast_to_bigint
+            "count", col_args, args.exclude_columns, None, cast_to_bigint=cast_to_bigint
         )
     if args.sum:
         col_args = None if args.sum == "*" else cli_tools.get_arg_list(args.sum)
         aggregate_configs += config_manager.build_config_column_aggregates(
             "sum",
             col_args,
-            exclude_cols,
+            args.exclude_columns,
             supported_data_types,
             cast_to_bigint=cast_to_bigint,
         )
@@ -104,7 +103,7 @@ def get_aggregate_config(args, config_manager: ConfigManager):
         aggregate_configs += config_manager.build_config_column_aggregates(
             "avg",
             col_args,
-            exclude_cols,
+            args.exclude_columns,
             supported_data_types,
             cast_to_bigint=cast_to_bigint,
         )
@@ -113,7 +112,7 @@ def get_aggregate_config(args, config_manager: ConfigManager):
         aggregate_configs += config_manager.build_config_column_aggregates(
             "min",
             col_args,
-            exclude_cols,
+            args.exclude_columns,
             supported_data_types,
             cast_to_bigint=cast_to_bigint,
         )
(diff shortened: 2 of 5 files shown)


data_validation/config_manager.py
```diff
@@ -677,7 +677,17 @@ def decimal_too_big_for_64bit(
         casefold_target_columns = {x.casefold(): str(x) for x in target_table.columns}
 
         if arg_value:
-            arg_value = [x.casefold() for x in arg_value]
+            if exclude_cols:
```
nj1973, line 680: I suspect this is in the wrong place. I think, but happy to be proved wrong, that `arg_value` is the column CSV passed in by the user, if the user specifies `'*'` then I think `arg_value` is `None`.

If I am right then I think your change only excludes column from a user provided list and not from `'*'`?

data_validation/__main__.py
sundar-mudupalli-work, line 1: It looks like `--exclude-columns` has to effect on --std columns. Is this something you missed ?

data_validation/config_manager.py
sundar-mudupalli-work, line 1: I think lines 717 to 719 can be re-written using list comprehension. You have done that  on line 715.

sundar-mudupalli-work:
Maria,

Thank you for working through this code change and getting it to submission. Generally, you have limited the code changes to only the files that are necessary, so that is awesome. As someone who has made code changes, it is not easy to understand someone else's code and make the changes. Here are my comments
1. The feature specification is unclear - I have raised that comment in the [issue](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/880) and looking for feedback from the team.
2. Has lint been run on your code? Please take a look at go/pso-data-validation-onboarding - slide 11 and follow the instructions there.
3. I made comments on a couple of your files, they should be straightforward to fix.

Thank you.

Sundar Mudupalli

data_validation/config_manager.py
mloeberc: Fixed

data_validation/__main__.py
mloeberc: You're right, already made the fix and updated the PR.

mloeberc:
@sundar-mudupalli-work I already worked and fixed on comments 2 and 3. As for the comments in the issue, I also answered, that, so we would be missing the confirmation for the "select * " scenario.

tests/unit/test_config_manager.py
sundar-mudupalli-work, line 1: Can you add the following assertion after line 315 just like you have in line 309
`assert len(aggregate_configs) == 1`

sundar-mudupalli-work: Otherwise looks good to me - like the fact that you have included an exception test as well. Thank you.

data_validation/config_manager.py
sundar-mudupalli-work, line 1: Lines 716 to 721 could be shortened as:
`arg_value = [
                    column
                    for column in casefold_source_columns
                    if column not in cols_excluded
                ]`
In theory, you could also remove the intermediate variables `cols_excluded` and `casefold_source_columns` - in my opinion that may make the code less readable.

data_validation/__main__.py
sundar-mudupalli-work, line 1: 1. `validate schema` defines a `--exclusion-columns` while here we are defining `--exclude-columns`. While I like the new wording, it would be nice to be consistent
2. I am suggesting some code cleanup in this file, `consts.py` and in `cli_tools.py` that does not change the functionality, makes the code more readable and less error prone. The problem we had with std earlier may not have occured with more compact code.

I will let Neha decide if and which ones she wants to be implemented.

Lines 85-146 can be shortened as
```
for agg_type in consts.AGGREGATION_TYPES:
    if args.getattr(agg_type) :
        col_args = None if args.getattr(agg_type) == '*' else cli_tools.get_arg_list(args.getattr(agg_type))
        aggregate_configs += config_manager.build_config_column_aggregates(agg_type, col_args, args.exclude_columns, None, args.cast_to_bigint)
```
provided we define the following in `consts.py`
```
AGGREGATION_HELP_PREFIX =  "Comma separated list of columns for "
AGGREGATION_HELP_SUFFIX =  " 'col_a,col_b' or * for all columns"
AGGREGATION_ARGS = [ ("count", "--count", "-count", "count"),
                    ("sum", "--sum", "-sum", "sum"),
                    ("avg", "--avg", "-avg", "avg"),
                    ("min", "--min", "-min", "min"),
                    ("max", "--max", "-max", "max"),
                    ("bit_xor", "--bit_xor", "-bit_xor", "hashing a concatenate"),
                    ("std", "--std", "-std", "standard deviation")]
AGGREGATION_TYPES=[AGG[0] for AGG in AGGREGATION_ARGS]
```
Then lines 520 to 554 and 770 to 805 in `cli_tools.py` can be replaced with a function that does: 
```
for agg_arg in consts.AGGREGATION_ARGS:
    optional_arguments.add_argument(agg_arg[1], agg_arg[2], help = consts.AGGREGATION_HELP_PREFIX + agg_arg[3] + consts.AGGREGATION_HELP_SUFFIX)
```
In theory, we can do similar cleanup for the flags and other arguments, that can be an exercise for another day. I have not tested the changes completely, so if there are some mistakes, apologies.

README.md
sundar-mudupalli-work, line 1: Maria, the help text in the code says standard deviation, whereas in the documentation below it says stddev_samp. It would be nice to be consistent.

nehanene15: `-std` is the flag name, but STDDEV_SAMP() is the actual function that will be called when the flag is present. If this is confusing, we should fix it in a separate docs PR.

data_validation/__main__.py
nehanene15: 1. Yeah - common flag names would be better... I think we should rename the schema validation to `--exclude-columns` since it's shorter + more intuitive.
2. I think this could be useful, but concatenating the string in runtime for each help message may not be the most optimized... don't think this is essential to do in this PR. Especially because 'stddev_samp' is the SQL function name so that was intended.

data_validation/config_manager.py
nehanene15: +1

data_validation/__main__.py
mloeberc: 1. I saw the difference between --exclusion-columns and --exclude-columns, but I left the difference because exclusion columns is a list of columns to exclude when validating the schema; while --exclude-columns (the new parameter associated to this PR) is a flag, not a list of columns, that indicates that the column lit in the aggregation has to be excluded and not included. That is why I didn't go for consistency, but for a differentiation between the two. I will hold on changing this for the moment.

mloeberc: As for the other changes, that involve reducing code in the files of main, cli_tools and consts, it might help to make code readable and avoid repeating the if statements, but it is code that was written before the change involving this issue, probably would be better to do it in a code refactor PR where tests are performed to make sure each operation still works; and would leave the scope of this PR in the addition of the exclude-columns flag.

Just to take into account, for the first recommendation in __main__.py for lines 85-146, is sending parameter supported_data_types to config_manager as "None", when that value only applies for "count" aggregation. I would recommend something like:

``` 
for agg_type in consts.AGGREGATION_TYPES:
        if args.getattr(agg_type) :
            supported_data_types = None if agg_type == "count" else supported_data_types
            col_args = None if args.getattr(agg_type) == '*' else cli_tools.get_arg_list(args.getattr(agg_type))
            aggregate_configs += config_manager.build_config_column_aggregates(agg_type, col_args, args.exclude_columns, supported_data_types, args.cast_to_bigint)
```

But as mentioned, I'm not sure this kind of code refactoring should be in the scope of this PR. @nehanene15, any thoughts?

sundar-mudupalli-work: Maria/Neha

1.I understand the distinction between exclusion columns and exclude columns, as you point out - they sound similar, but are very different.
2.This is code refactoring. I am OK with not making any changes and approve the PR as is.
3. If you want to do the `__main__.py` refactoring - I just cut and pasted the repeated code in the file today. Don't understand why you are changing the code (saying None only applies for count aggregation) while refactoring. Was that an oversight or is there a bug in the code that requires the change?
4. If you do the `__main__.py` refactoring you have to refactor `consts.py` as I mentioned. Otherwise your first test case will fail because the const `consts.AGGREGATION_TYPES` is not defined.

Cheers.

Sundar Mudupalli

data_validation/config_manager.py
```diff
@@ -711,7 +711,17 @@ def require_pre_agg_calc_field(
...
```
nehanene15, line 715: This may be more concise:

```
        if arg_value:
            arg_value = [x.casefold() for x in arg_value]
            if exclude_cols:
                included_cols = [
                    column
                    for column in casefold_source_columns
                    if column not in arg_value
                ]
                arg_value = included_cols

```

nehanene15 (commented):
Looks good besides the code optimization comment!

data_validation/config_manager.py
nehanene15: The above comment wasn't implemented here - you can get rid of the 'else' block on line 721 if you declare `arg_value` before the `if exclude_cols` condition

nehanene15 (approved):
LGTM!

PR #965, opened 2023-08-31
fix: Fix warning and precision detection when target precision higher than source

### LINKED ISSUE #949: Review warning messages for precision mismatch in schema validation
It seems that this warning below is longer needed, we need to review it and remove or update it.

```
data-validation validate schema -sc oracle21 -tc bq -tbls pso_data_validator.dvt_core_types --allow-list-file=professional-services-data-validator/samples/allow_list/oracle_to_bigquery.yaml
08/21/2023 01:43:24 PM-WARNING: Source and target data type has precision mismatch: decimal(20,0) - decimal(38, 9)
08/21/2023 01:43:24 PM-WARNING: Source and target data type has precision mismatch: decimal(38,0) - decimal(76, 38)
08/21/2023 01:43:24 PM-WARNING: Source and target data type has precision mismatch: decimal(10,2) - decimal(38, 9)
```

Implemented on: https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/data_validation/schema_validation.py#L189

nj1973:
When validating schema for decimal columns the logic for when to warn the user and when to fail was incorrect. I have corrected the logic to match the original intention from prior issue (defined below):

1. If source type = target type exactly then success
2. If source/target match an allow list pair then success
3. If not an exact match and not an allow list pair then
3.1. If base type is same and target precision lower then fail
3.2 If base type is same and target precision higher then success with warning
3.3 Else fail

Also, now the check is in the right place I discovered that the precision extraction code didn't work and precision was always `-1`. So I fixed that too.

Plus added a unit test.

### DIFF
--- a/data_validation/schema_validation.py
+++ b/data_validation/schema_validation.py
@@ -187,7 +187,7 @@ def schema_validation_matching(
                 string_val(source_field_type), string_val(target_field_type)
             )
             if higher_precision:
-                # If the target precision is higher then tha validation is acceptable but worth a warning.
+                # If the target precision is higher then the validation is acceptable but worth a warning.
                 logging.warning(
                     "Source and target data type has precision mismatch: %s - %s",
                     string_val(source_field_type),
--- a/data_validation/schema_validation.py
+++ b/data_validation/schema_validation.py
@@ -154,7 +154,6 @@ def schema_validation_matching(
             )
             continue
 
-        # Target field exists
         target_field_type = target_fields_casefold[source_field_name]
         if source_field_type == target_field_type:
             # Target data type matches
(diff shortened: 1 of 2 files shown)


data_validation/schema_validation.py
```diff
@@ -172,48 +171,49 @@ def schema_validation_matching(
...
+                ]
+            )
+        else:
+            # Target data type mismatch
             (higher_precision, lower_precision,) = parse_n_validate_datatypes(
                 string_val(source_field_type), string_val(target_field_type)
             )
-            if lower_precision:
+            if higher_precision:
+                # If the target precision is higher then tha validation is acceptable but worth a warning.
```
nehanene15: nit: typo 'tha'

nehanene15 (approved):
LGTM! Great job

PR #976, opened 2023-09-07
feat: Support for null decimal columns and TD access locks

### LINKED ISSUE #969: Need to add an option to provide lock types (e.g: access lock) to be provided in queries to Teradata
Need to add an option to provide lock types (e.g: access lock) to be provided in queries to Teradata. Right now if we are running any dvt command while query execution its taking default locking set in DB. Issue is if let say read lock is there in Teradata, the dvt will block any other incoming critical sessions from read and write.

### LINKED ISSUE #970: NullColumn error during column validation
We are getting following error while running column validation :

`09/04/2023 05:07:04 AM-ERROR: Error unsupported operand type(s) for -: 'NullColumn' and 'NullColumn' occurred while running config file f86910a6-1335-4b9d-841b-1d7b7b195fa4.yaml. Skipping it for now.`

nehanene15:
Closes #970 
Closes #969

This will require users to specify `--use-no-lock-tables True` to use an access lock instead of a read lock on TD.
If enabled, will append the prefix "LOCKING ROW FOR ACCESS" before query execution.

### DIFF
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -65,6 +65,7 @@ CONNECTION_SOURCE_FIELDS = {
         ["user_name", "User used to connect"],
         ["password", "Password for supplied user"],
         ["logmech", "(Optional) Log on mechanism"],
+        ["use_no_lock_tables", "Use an access lock for queries (defaults to False)"],
     ],
     "Oracle": [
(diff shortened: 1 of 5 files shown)


Raniksingh (approved):
LGTM. Verified by uesr.

PR #977, opened 2023-09-08
fix: yaml validation files in gcs

### LINKED ISSUE #972: YAML validation directories supported in GCS, but not validation files in GCS
Hi,

When we have YAML directories in GCS, they are supported - so the command `data-validation configs run -cdir gs://mudu-plex-pri-myb/db6rows/public.iowa_27rows` works when you have yaml files in that directory.

However specifying `data-validation configs run -c gs://mudu-plex-pri-myb/db6rows/public.iowa_27rows/0001.yaml` throws the following error
```
  File "/home/mudupalli/professional-services-data-validator/env/lib/python3.9/site-packages/data_validation/state_manager.py", line 190, in _read_file
    with open(file_path, "r") as f:
FileNotFoundError: [Errno 2] No such file or directory: './gs://mudu-plex-pri-myb/db6rows/public.iowa_27rows/0001.yaml'
```

This is a bug.

Thanks.

Sundar Mudupalli

sundar-mudupalli-work:
This change fixes [972](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/972). When using `data-validation configs run -c gs://mudu-plex-pri-myb/db6rows/public.iowa_27rows/0001.yaml` was throwing an error.

This fix does not have a test cases. We don't have any GCS based test cases in unit tests. Do the unit tests have access to a GCS bucket ? I tested the fix manually and it works. It would be nice for this to be tested in the test suite

Sundar Mudupalli

### DIFF
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -1062,14 +1062,16 @@ def get_validation(validation_name, config_dir=None):
         mgr = state_manager.StateManager(file_system_root_path=config_dir)
         return mgr.get_validation_config(validation_name, config_dir)
     else:
-       if validation_name.startswith("gs://"):
-           obj_depth = len(validation_name.split("/"))
-           gcs_prefix = '/'.join(validation_name.split('/')[:obj_depth-1])
-           mgr = state_manager.StateManager(file_system_root_path=gcs_prefix)
-           return mgr.get_validation_config(validation_name.split('/')[obj_depth-1],gcs_prefix)
-       else: 
-           mgr = state_manager.StateManager()
-           return mgr.get_validation_config(validation_name)
+        if validation_name.startswith("gs://"):
+            obj_depth = len(validation_name.split("/"))
+            gcs_prefix = "/".join(validation_name.split("/")[: obj_depth - 1])
+            mgr = state_manager.StateManager(file_system_root_path=gcs_prefix)
+            return mgr.get_validation_config(
+                validation_name.split("/")[obj_depth - 1], gcs_prefix
+            )
+        else:
+            mgr = state_manager.StateManager()
+            return mgr.get_validation_config(validation_name)
 
 
 def list_validations():
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -1062,8 +1062,14 @@ def get_validation(validation_name, config_dir=None):
         mgr = state_manager.StateManager(file_system_root_path=config_dir)
         return mgr.get_validation_config(validation_name, config_dir)
     else:
-        mgr = state_manager.StateManager()
-        return mgr.get_validation_config(validation_name)
+       if validation_name.startswith("gs://"):
+           obj_depth = len(validation_name.split("/"))
+           gcs_prefix = '/'.join(validation_name.split('/')[:obj_depth-1])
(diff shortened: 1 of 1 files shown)


data_validation/cli_tools.py
```diff
@@ -1062,8 +1062,16 @@ def get_validation(validation_name, config_dir=None):
         mgr = state_manager.StateManager(file_system_root_path=config_dir)
         return mgr.get_validation_config(validation_name, config_dir)
     else:
-        mgr = state_manager.StateManager()
-        return mgr.get_validation_config(validation_name)
+        if validation_name.startswith("gs://"):
```
nehanene15, line 1065: I think this logic should go within the StateManager get_validation_config() function. In the function, we can check for `if validation_name.startswith("gs://")` similar to [this function](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/47154b4139bf22358359c53fe25bbce44745589f/data_validation/state_manager.py#L147).

nehanene15 (changes requested):

sundar-mudupalli-work:
Neha,

Good question. I did a bit of research and understand the code better.

Summary: I think the fix I proposed is in the right place and happy to hear another point of view. Here is my reasoning:

1. The `StateManager` object upon instantiation decides whether it is working with the native filesystem or GCS. This value is in the object attribute `file_system` which is either `FileSystem.GCS` or `FileSystem.LOCAL`. If `StateManager` object is instantiated without any parameters, the `file_system` attribute is `FileSystem.LOCAL` unless the environment variable `PSO_DV_CONFIG_HOME` points to a GCS directory.
2. If we want to read from a file in GCS, we must (highly recommended?) instantiate the `StateManager` object with the GCS prefix - which is what this code does.
3. We can instantiate a `StateManager` object with no parameters and then move this code to `get_validation_config`. The last line in `get_validation_config` is `_read_file` which depends on `file_system` attribute being set correctly. Because we instantiated `StateManager` with no parameters (and environment variable `PSO_DV_CONFIG_HOME` is not set) we will have to reset the `file_system` attribute in `get_validation_config` before calling `_read_file`. Not a great idea in my opinion.

I looked at `list_validations_in_dir` function. The test `startswith("gs://")` appears unnecessary, it could have used the `self.file_system` attribute. This function is called from `__main__` and the `StateManager` is instantiated with the right parameter. This function is identical to the function `list_validations` right above it!!. These could have been collapsed into one function with `config_dir` as an optional argument with a default value of `None`.

It is a good question, hopefully you find it helpful.

Sundar Mudupalli

data_validation/cli_tools.py
```diff
@@ -1062,8 +1062,16 @@ def get_validation(validation_name, config_dir=None):
...
+            obj_depth = len(validation_name.split("/"))
+            gcs_prefix = "/".join(validation_name.split("/")[: obj_depth - 1])
```
nehanene15, line 1067: This might be easier than getting the depth - 1: `"/".join(file.split("/")[:-1])`

nehanene15:
@sundar-mudupalli-work That makes sense. Seems like we need to clean up the StateManager functions a bit.
Left a small comment on optimizing the gcs_prefix path, but besides that LGTM

nehanene15 (approved):

PR #980, opened 2023-09-11
fix: Add not-null string to accepted date types in append_pre_agg_calc_field()

### LINKED ISSUE #979: validate column: Exception from min/max on not null string column
In `data_validation/config_manager.py` function `append_pre_agg_calc_field()` we have this check:

```
...
        if column_type == "string":
            calc_func = "length"

        elif column_type == "timestamp" or column_type == "!timestamp":
...
```

A not null string will pass through this check and throw an exception:
```
ValueError: Unsupported column type: !string
```

We should change this to an `in` and include `"!string"`.

nj1973:
We were getting an exception from column validation when attempting a min or max validation on a non-null string column:

```
ValueError: Unsupported column type: !string
```

I've added a test for this to `test_config_manager.py`. I needed to add a new mock class `MockIbisColumn` so I could control the returned data type for a specific column.

I've also changed `data_validation/config_manager.py` to treat `!string` in the same way as `string`.

### DIFF
--- a/data_validation/config_manager.py
+++ b/data_validation/config_manager.py
@@ -595,10 +595,10 @@ class ConfigManager(object):
     ) -> dict:
         """Append calculated field for length(string) or epoch_seconds(timestamp) for preprocessing before column validation aggregation."""
         depth, cast_type = 0, None
-        if column_type == "string":
+        if column_type in ["string", "!string"]:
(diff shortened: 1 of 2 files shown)


nehanene15 (approved):
LGTM!

PR #994, opened 2023-09-19
fix: custom query row validation failing when SQL contains upper cased columns

### LINKED ISSUE #992: custom query row validation has inconsistent column case to table validation
This example came up on an Oracle to PostgreSQL validation. Default identifier case in Oracle is uppercase. Default identifier case in PostgreSQL is lowercase. 

Oracle test table:
```
create table dvt_test.tab_case (id number(6), col_data varchar2(20));
insert into dvt_test.tab_case values (1,'Hello');
commit;
```

PostgreSQL test table:
```
create table dvt_test.tab_case (id numeric(6), col_data varchar(20));
insert into dvt_test.tab_case values (1,'Hello');
```

**Table level row validation**

```
 -sc ora_local -tc pg_local \
 -tbls=dvt_test.tab_case \
 --primary-keys=id \
 --comparison-fields=col_data
```

This works fine and I can see from the debugger that the `table` objects in `query_builder.compile()` are:
```
AlchemyTable: tab_case
  id       decimal(6, 0)
  col_data string

AlchemyTable: tab_case
  id       decimal(6, 0)
  col_data string
```

Notice the upper case Oracle columns have been converted to lower case.

**Custom query row validation**

```
data-validation validate custom-query row \
 -sc ora_local -tc pg_local \
 --source-query='select * from dvt_test.tab_case' \
 --target-query="select * from dvt_test.tab_case" \
 --primary-keys=id \
 --comparison-fields=col_data
...
    raise com.IbisTypeError(
ibis.common.exceptions.IbisTypeError: Column 'col_data' is not found in table. Existing columns: 'ID', 'COL_DATA'.
```

I can see from the debugger that the `table` object in `query_builder.pt.compile()` is:
```
SQLQueryResult
  query: 'select * from dvt_test.tab_case'
  schema:
    ID       decimal(6, 0)
    COL_DATA string
```

The columns remain upper case.

If I change `--comparison-fields` to upper case we then fail on the PostgreSQL side.
```
data-validation validate custom-query row \
 -sc ora_local -tc pg_local \
 --source-query='select * from dvt_test.tab_case' \
 --target-query="select * from dvt_test.tab_case" \
 --primary-keys=id \
 --comparison-fields=COL_DATA
...
    raise com.IbisTypeError(
ibis.common.exceptions.IbisTypeError: Column 'COL_DATA' is not found in table. Existing columns: 'id', 'col_data'.
```

The workaround is to force the Oracle query to result in a lower case column.

I think we are inconsistent somewhere in the code. For table validations column names are lower cased, for custom query validations they are not.

nj1973:
We found this issue on Oracle where column default to upper case but it could easily be reproduced on other engines by writing SQL with upper cased columns, the columns in the actual SQL engine were irrelevant, it was just the SQL we pass in.

The problem was only seen for custom query validation using `--comparison-fields`. When using `--hash` the calculated fields resulted in lower case end column names.

I added tests to protect from regressions for Oracle, SQL Server, Teradata vs BigQuery. The fix was global so should wortk on all engines but I was reluctant to add too many tests seeing as integration tests are already pretty slow.

The fix was very minor, in `data_validation/clients.py` I convert the Ibis schema to have lower case columns:
```
-    return client.sql(query)
+    iq = client.sql(query)
+    # Normalise all columns in the query to lower case.
+    # https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/992
+    iq = iq.relabel(dict(zip(iq.columns, [_.lower() for _ in iq.columns])))
+    return iq
```

### DIFF
--- a/data_validation/clients.py
+++ b/data_validation/clients.py
@@ -140,7 +140,11 @@ def get_ibis_table(client, schema_name, table_name, database_name=None):
 
 def get_ibis_query(client, query):
     """Return Ibis Table from query expression for Supplied Client."""
-    return client.sql(query)
+    iq = client.sql(query)
+    # Normalise all columns in the query to lower case.
+    # https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/992
+    iq = iq.relabel(dict(zip(iq.columns, [_.lower() for _ in iq.columns])))
+    return iq
 
 
 def get_ibis_table_schema(client, schema_name, table_name):
--- a/tests/system/data_sources/test_oracle.py
+++ b/tests/system/data_sources/test_oracle.py
@@ -321,20 +321,80 @@ def test_row_validation_core_types_to_bigquery():
     "data_validation.state_manager.StateManager.get_connection_config",
     new=mock_get_connection_config,
 )
-def test_custom_query_validation_core_types():
-    """Oracle to Oracle dvt_core_types custom-query validation"""
+def test_custom_query_column_validation_core_types_to_bigquery():
+    """Oracle to BigQuery dvt_core_types custom-query column validation"""
     parser = cli_tools.configure_arg_parser()
     args = parser.parse_args(
         [
             "validate",
             "custom-query",
             "column",
-            "-sc=mock-conn",
-            "-tc=mock-conn",
+            "-sc=ora-conn",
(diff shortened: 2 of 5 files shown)


nehanene15 (approved):
LGTM!

PR #997, opened 2023-09-20
docs: Add a new sample code for row hash validation of Oracle BLOB

### LINKED ISSUE #996: Document sample code for row hash validation of Oracle BLOB
I can't see an easy way to build hash validation of Oracle BLOB into DVT but we have had a customer ask about it and I've figured out a way with custom query validation.

I think the best place to document this is in the repo and not just by sending the solution to the customer. Hence this issue.

nj1973:
I can't see an easy way to build hash validation of Oracle BLOB into DVT but we have had a customer ask about it and I've figured out a way forward with custom query validation.

I think the best place to document this is in the sample code area of the repo.

Please see if you think this is appropriate and makes sense.

### DIFF
--- a/samples/oracle/README.md
+++ b/samples/oracle/README.md
@@ -83,7 +83,7 @@ SELECT id,encode(sha256(col_blob),'hex') AS col_blob FROM dvt_test.tab_blob;
 DVT command:
 ```
 data-validation validate custom-query row \
- -sc ora_local -tc pg_local \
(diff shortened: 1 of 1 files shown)


nehanene15 (approved):
LGTM! If you want, we could link to this sample in the [Oracle connection guide](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/docs/connections.md#oracle) as well

PR #999, opened 2023-09-21
fix: Ensure max keyword is included in casts to VARCHAR

renzokuken:
In theory a cast to string on SQL Server should automatically include the `MAX` keyword, see this comment in SQL Alchemy code in `sqlalchemy/dialects/mssql/base.py`:
```
MAX on VARCHAR / NVARCHAR
-------------------------

SQL Server supports the special string "MAX" within the
:class:`_types.VARCHAR` and :class:`_types.NVARCHAR` datatypes,
to indicate "maximum length possible".   The dialect currently handles this as
a length of "None" in the base type, rather than supplying a
dialect-specific version of these types, so that a base type
specified such as ``VARCHAR(None)`` can assume "unlengthed" behavior on
more than one backend without using dialect-specific types.
Therefore it is strange that the change is not having the desired effect.
```

In `sqlalchemy/dialects/mssql/base.py` class `MSTypeCompiler` there is a `visit_VARCHAR` method which uses "max" if a type has no length. However, our casts are passing through `sqlalchemy/sql/compiler.py` class `GenericTypeCompiler.visit_VARCHAR` which does not support "max". I cannot figure out why, perhaps Ibis is getting in the way somewhere.

The best way forward I could find was to add a SQL Server specific cast override which ensures we include the MAX keyword. But seeing as I cannot reproduce the issue myself I've asked the original reporter to test it.

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -21,6 +21,7 @@ set -eo pipefail
 export PYTHONUNBUFFERED=1
 
 # Install nox
+python3.8 -m pip cache purge
 python3.8 -m pip install --upgrade --quiet nox pip
 python3.8 -m nox --version
 
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -21,7 +21,6 @@ set -eo pipefail
 export PYTHONUNBUFFERED=1
 
 # Install nox
-python3.8 -m pip cache purge
 python3.8 -m pip install --upgrade --quiet nox pip
 python3.8 -m nox --version
 
--- a/noxfile.py
+++ b/noxfile.py
@@ -41,7 +41,7 @@ def _setup_session_requirements(session, extra_packages=[]):
     """Install requirements for nox tests."""
 
     session.install("--upgrade", "pip", "pytest", "pytest-cov", "wheel")
-    session.install("--no-cache-dir", "-e", ".")
+    session.install("-e", ".")
 
     if extra_packages:
         session.install(*extra_packages)
--- a/noxfile.py
+++ b/noxfile.py
@@ -41,7 +41,7 @@ def _setup_session_requirements(session, extra_packages=[]):
     """Install requirements for nox tests."""
 
     session.install("--upgrade", "pip", "pytest", "pytest-cov", "wheel")
-    session.install("-e", ".")
+    session.install("--no-cache-dir", "-e", ".")
 
     if extra_packages:
         session.install(*extra_packages)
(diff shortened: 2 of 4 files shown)


nj1973:
I checked out this branch and ran my test case again and I don't think the current changes are doing the trick.
```
$ git status
On branch issue990-sqlserver-casting
Your branch is up to date with 'origin/issue990-sqlserver-casting'.
```

Validation of a not-null nvarchar column (notice cast__shortdescription2 includes a CAST to VARCHAR):
```
$ data-validation -v validate row -sc sqlserver -tc sqlserver -tbls dbo.tblIssue990 --primary-keys=id --hash='ID,ShortDescription'

09/26/2023 01:19:39 PM-INFO: -- ** Source Query ** --
09/26/2023 01:19:39 PM-INFO: WITH t0 AS
(SELECT t6."ID" AS "ID", t6."ShortDescription" AS "ShortDescription", t6."ShortDescription2" AS "ShortDescription2", CAST(t6."ID" AS VARCHAR) AS cast__id, CAST(t6."ShortDescription" AS VARCHAR) AS cast__shortdescription
FROM dbo."tblIssue990" AS t6),
t1 AS
...
```

Validation of a nullable nvarchar column (notice cast__shortdescription2 does not include a CAST):
```
$ data-validation -v validate row -sc sqlserver -tc sqlserver -tbls dbo.tblIssue990 --primary-keys=id --hash='ID,ShortDescription2'
09/26/2023 01:20:44 PM-INFO: {'data_client': <third_party.ibis.ibis_mssql.Backend object at 0x7879fd933ca0>, 'schema_name': 'dbo', 'table_name': 'tblIssue990', 'source_query': None}
09/26/2023 01:20:44 PM-INFO: -- ** Source Query ** --
09/26/2023 01:20:44 PM-INFO: WITH t0 AS
(SELECT t6."ID" AS "ID", t6."ShortDescription" AS "ShortDescription", t6."ShortDescription2" AS "ShortDescription2", CAST(t6."ID" AS VARCHAR) AS cast__id, t6."ShortDescription2" AS cast__shortdescription2
FROM dbo."tblIssue990" AS t6),
t1 AS
...
```

You should be able to use the commands above from a Cloud Workstation to reproduce yourself.

nj1973:
Closing this PR because I'm no longer sure what the original issue was. I'll create a new PR once I know more about what we're fixing.

Issue #899, opened 2023-07-11
test: Add an integration test for Oracle to PostgreSQL validations


nj1973:
Current cross engine integration tests are focused on only basic data types and have BigQuery as the target (BigQuery has a smaller number of data type options than most engines). Because of this we are not exercising plenty of data types.

Add an integration test for Oracle to PostgreSQL validations using any Oracle specific data types that are not covered by the `dvt_core_types` tests. Both Oracle and PostgreSQL validations are quick so this should not cause any problems with test elapsed times.

Raise issues for any problems we uncover.

nj1973:
Test table DDL:

```
CREATE TABLE dvt_ora2pg_types
(   id              NUMBER(8) NOT NULL PRIMARY KEY
,   col_num_4       NUMBER(4)
,   col_num_9       NUMBER(9)
,   col_num_18      NUMBER(18)
,   col_num_38      NUMBER(38)
,   col_num         NUMBER
,   col_num_10_2    NUMBER(10,2)
--,   col_num_6_m2    NUMBER(6,-2)
--,   col_num_3_5     NUMBER(3,5)
,   col_num_float   FLOAT
,   col_float32     BINARY_FLOAT
,   col_float64     BINARY_DOUBLE
,   col_varchar_30  VARCHAR2(30)
,   col_char_2      CHAR(2)
,   col_nvarchar_30 NVARCHAR2(30)
,   col_nchar_2     NCHAR(2)
,   col_date        DATE
,   col_ts          TIMESTAMP(6)
,   col_tstz        TIMESTAMP(6) WITH TIME ZONE
--,   col_tsltz       TIMESTAMP(6) WITH LOCAL TIME ZONE
--,   col_raw         RAW(16)
,   col_clob        CLOB
,   col_nclob       NCLOB
);
```

Currently cannot add certain columns for following reasons:

- col_raw: https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/873
- col_tsltz: TIMESTAMP WITH LOCAL TIME ZONE unsupported
- col_num_6_m2: https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/898
- col_num_3_5: https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/898

nj1973:
This new test has highlighted an issue.

- The col_num_18 column is an 18 digit number stored in a bigint.
[...]

Issue #773, opened 2023-03-14
validate row: Oracle RAW Columns Cause Exception


nj1973:
Oracle RAW columns hold binary data and are sometimes used to store UUID/GUID data. This is not a common pattern but certainly something I've seen in the past.

Below are two test tables, one with a GUID column and another with a GUID column as a primary key.
```
create table dvt_test.tab_guid (
  id number(2), col_guid raw(16)
);
ALTER TABLE dvt_test.tab_guid ADD PRIMARY KEY (id);
insert into dvt_test.tab_guid values (1,sys_guid());
insert into dvt_test.tab_guid values (2,sys_guid());
commit;

create table dvt_test.tab_guid_pk (
  id raw(16), col_dec number(5)
);
ALTER TABLE dvt_test.tab_guid_pk ADD PRIMARY KEY (id);
insert into dvt_test.tab_guid_pk values (sys_guid(),1);
insert into dvt_test.tab_guid_pk values (sys_guid(),2);
commit;
```

tab_guid command:
```
data-validation validate row -sc ora_local -tc ora_local \
-tbls=dvt_test.tab_guid --primary-keys id --concat '*' \
--filter-status fail
...
sqlalchemy.exc.DatabaseError: (cx_Oracle.DatabaseError) ORA-00904: "ENCODE": invalid identifier
```

tab_guid_pk command:
```
data-validation validate row -sc ora_local -tc ora_local \
-tbls=dvt_test.tab_guid_pk --primary-keys id --concat '*' \
--filter-status fail
...
  File "pandas/_libs/lib.pyx", line 2859, in pandas._libs.lib.map_infer
  File "/usr/local/google/home/neiljohnson/github/professional-services-data-validator/.venv/lib/python3.10/site-packages/ibis/backends/pandas/execution/strings.py", line 286, in replacer
    return pattern.sub(replacement, x)
TypeError: cannot use a string pattern on a bytes-like object
```

I haven't tested PostgreSQL but that has a dedicated UUID data type. When working on this we should do the equivalent PostgreSQL tests as above.

nj1973:
Still fails after Ibis upgrade

nj1973:
This is also true of Oracle BLOB columns.

nj1973:
I've retested the test commands today and no exceptions are throws.

PR #1005, opened 2023-09-27
fix: Prevent Oracle blob throwing exceptions during column validation

### LINKED ISSUE #991: Oracle BLOB throws exceptions for column and row validation
This has been requested by a large bank so is something we should try to prioritise.

Test Oracle table:
```
create table dvt_test.tab_blob (id number(6), col_blob blob, col_raw raw(64));
insert into dvt_test.tab_blob values (1,empty_blob(),NULL);
insert into dvt_test.tab_blob values (2,utl_raw.cast_to_raw('some binary data'),utl_raw.cast_to_raw('some binary data'));
insert into dvt_test.tab_blob values (3,NULL,NULL);
commit;
```

Test PostgreSQL table:
```
create table dvt_test.tab_blob (id decimal(6), col_blob bytea, col_raw bytea);
insert into dvt_test.tab_blob values (1,null,null);
insert into dvt_test.tab_blob values (2,'some binary data'::bytea,'some binary data'::bytea);
insert into dvt_test.tab_blob values (3,null,null);
```

**Column validation**

Counts:
```
data-validation -v validate column -sc=ora_local -tc=pg_local -tbls=dvt_test.tab_blob --count=col_blob,col_raw

sqlalchemy.exc.DatabaseError: (cx_Oracle.DatabaseError) ORA-00932: inconsistent datatypes: expected - got BLOB
[SQL: SELECT count(*) AS count, count(t0.col_blob) AS count__col_blob
FROM dvt_test.tab_blob t0]
```

Sum/min/max:
```
data-validation -v validate column -sc=ora_local -tc=pg_local -tbls=dvt_test.tab_blob --min=col_blob,col_raw

# No error but the column is skipped

data-validation -v validate column -sc=ora_local -tc=pg_local -tbls=dvt_test.tab_blob --sum=col_blob,col_raw

# No error but the column is skipped
```

**Row validation**

Oracle to Oracle validation:
```
data-validation -v validate row -sc=ora_local -tc=ora_local -tbls=dvt_test.tab_blob \
--primary-keys=id --hash=col_blob

(SELECT t6.id AS id, t6.col_blob AS col_blob, encode(t6.col_blob, :encode_1) AS cast__col_blob
FROM dvt_test.tab_blob AS t6),
t1 AS
...
cx_Oracle.DatabaseError: ORA-00904: "ENCODE": invalid identifier
```

PostgreSQL to PostgreSQL is fine:
```
data-validation -v validate row -sc=pg_local -tc=pg_local -tbls=dvt_test.tab_blob \
--primary-keys=id --hash=col_blob

# success
```

Once #899 is complete we should add a BLOB column to the new test table.

nj1973:
Changes in this PR:

- Add a new `BinaryLength` expression which is used in the same way we use `StringLength` to support min/max/sum aggregations
- Use `BinaryLength` when a user explicitly asks for a binary column to be validated
- Use `BinaryLength` when count validating a binary column if either the source or target system are Oracle

In most engines `COUNT(binary_column)` is fine but in Oracle it throws an exception if the binary column is of type BLOB. It does work fine for RAW though which is also binary. This makes detection tricky. I can't see a way to know the original Oracle type in `config_manager.py` so I decided we would need to use `COUNT(LENGTH(binary_column))` for both BLOB and RAW columns.

This PR does not fix an exception in row validation for BLOB/RAW columns, that is covered by https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/773.

This PR does not add an integration test for BLOB, that will be covered by https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/899.

### DIFF
--- a/data_validation/config_manager.py
+++ b/data_validation/config_manager.py
@@ -721,6 +721,8 @@ class ConfigManager(object):
                     "!timestamp",
                     "date",
                     "!date",
+                    "binary",
+                    "!binary",
                 ]
 
         allowlist_columns = arg_value or casefold_source_columns
--- a/data_validation/config_manager.py
+++ b/data_validation/config_manager.py
@@ -593,10 +593,13 @@ class ConfigManager(object):
     def append_pre_agg_calc_field(
         self, source_column, target_column, agg_type, column_type, column_position
     ) -> dict:
-        """Append calculated field for length(string) or epoch_seconds(timestamp) for preprocessing before column validation aggregation."""
+        """Append calculated field for length(string | binary) or epoch_seconds(timestamp) for preprocessing before column validation aggregation."""
         depth, cast_type = 0, None
         if column_type in ["string", "!string"]:
             calc_func = "length"
+
+        elif column_type in ["binary", "!binary"]:
+            calc_func = "byte_length"
 
         elif column_type in ["timestamp", "!timestamp", "date", "!date"]:
             if (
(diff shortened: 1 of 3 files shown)


nehanene15 (approved):
Looks great!

PR #1012, opened 2023-10-11
test: Remove old MySQL integration tests before standardization for all systems 

### LINKED ISSUE #777: test: MySQL Integration Test test_schema_validation Fails
This test has not been executing due to missing Cloud SQL Proxy in integration tests.
After changing test setup to use  Cloud SQL Proxy the test then executes and fails as below:
```
E       TypeError: reduce() of empty iterable with no initial value

data_validation/combiner.py:231: TypeError
============================================================ short test summary info =============================================================
FAILED tests/system/data_sources/test_mysql.py::test_schema_validation - TypeError: reduce() of empty iterable with no initial value
```

The test has been disabled in `tests/system/data_sources/test_mysql.py`. When this issue is complete we need to enable the integration test, search for "issue-777" in `tests/system/data_sources/test_mysql.py` to action this. Enabling these test is blocked by issue https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/759.

helensilva14:
Closes #777 

Initially I wanted to check if this test was still failing, that's when we got the #1018.

Since it's solved now, test finally executed and it was falling. But after reviewing our current tests and the standardization we made, I realized `test_schema_validation` is actually an old test that does not make sense anymore, since now we have `test_schema_validation_core_types` and `test_schema_validation_not_null_vs_nullable`. Same thing for  `test_mysql_row` that I'm also deleting on this PR, we have other methods now to test the feature.


nj1973:
Regarding the failing tests, I tried re-running tests for a branch I know built previously and it also fails with the pip error. So I think we have a system wide problem.

helensilva14:
@nj1973 thanks again for working on the build error pip issue, very appreciated!!!

I updated this PR's description, could you please take a look? Complementing the description, I was wondering if we need to add `test_schema_validation_core_types_to_bigquery` for MySQL like we have for Oracle and Postgres? I don't think so because the allow-list feature addresses Oracle and Postgres specifically, but just wanted to confirm with you, what do you think? Thanks in advance!

nj1973:
I agree, I think the newer tests make the original `test_schema_validation` and `test_mysql_row` tests redundant.
I think we probably should have a `test_schema_validation_core_types_to_bigquery` for MySQL, I must have missed it when I added those tests.
If it is easy to add then I would add it. If it causes problems then perhaps save it for a new issue. I'll approve anyway because what you've done LGTM.

nj1973 (approved):
Approving on the basis that current changes look good. I added a comment to optionally add `test_schema_validation_core_types_to_bigquery` tests.

helensilva14:
/gcbrun

Added `test_schema_validation_core_types_to_bigquery` for MySQL using this doc as reference: https://cloud.google.com/data-fusion/docs/reference/replication-data-types#mysql. Checking if it passes!

PR #1024, opened 2023-10-30
test: integration test for Oracle to PostgreSQL

### LINKED ISSUE #899: test: Add an integration test for Oracle to PostgreSQL validations

nj1973:
Add an integration test for Oracle to PostgreSQL validations using any Oracle specific data types that are not covered by the dvt_core_types tests. Both Oracle and PostgreSQL validations are quick so this should not cause any problems with test elapsed times.

I've added tests for schema, column and row validations.


nehanene15 (approved):
Looks great Neil!

PR #1034, opened 2023-11-07
fix: Adjust `find-tables` to properly get Oracle and Postgres schemas

### LINKED ISSUE #1017: find-tables not working
We are running an homogeneous migration: AWS RDS Postgres 12.14 to Google Cloud SQL AlloyDB.

After running migration with DMS, when checking with Data Validation Tool, as explained [here](https://cloud.google.com/database-migration/docs/oracle-to-postgresql/verify-migration#verifying-using-DVT), we can create connections, but running **find-tables** return an empty result.

Running with verbose mode gives no extra info.

UPDATE: Correction in my initial statement - This is a heterogeneous migration: **AWS RDS Oracle 12.14 to Google Cloud SQL AlloyDB**.

helensilva14:
Closes #1017

### DIFF
--- a/data_validation/__main__.py
+++ b/data_validation/__main__.py
@@ -368,6 +368,9 @@ def get_table_map(client, allowed_schemas=None):
     """Return dict with searchable keys for table matching."""
     table_map = {}
     table_objs = clients.get_all_tables(client, allowed_schemas=allowed_schemas)
+
+    print(f"TABLE MAP: {str(table_objs)}")
+
     for table_obj in table_objs:
         table_key = ".".join([t for t in table_obj if t])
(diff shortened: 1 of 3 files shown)


nj1973 (approved):
Changes LGTM, only comment is that we don't have any tests for find-tables to protect against regressions.

I'll leave it up to you, seeing as this is a customer issue you could merge this and add a new issue to create some tests at some in the future if you prefer.

helensilva14:
@nj1973 Thank you, totally agree! I tried to create a test for it to check how much effort it'd require and I couldn't make it work for now even with some searching. Regardless, I created #1043 and added my initial attempts there!

Issue #990, opened 2023-09-13
row validation: Long text are being cropped to 30 characters for hash validation on SQL Server


nj1973:
When running hash validations columns are cast to VARCHAR, then concatenated and finally a hash is produced.

We've had a customer issue reported for nvarchar(500) and nvarchar(2000) columns. The cast to varchar uses a length of 30 by default so is trimming the text. We should consider casting to `varchar(max)` instead (or another technique if that works out to be easier).

Varchar docs: https://learn.microsoft.com/en-us/sql/t-sql/data-types/char-and-varchar-transact-sql?view=sql-server-ver16
Cast docs: https://learn.microsoft.com/en-us/sql/t-sql/functions/cast-and-convert-transact-sql?view=sql-server-ver16

Examples from sqlcmd

Notice the `x` (at position 31) is cropped:
```
1> select concat(cast('123456789012345678901234567890x' as varchar),cast('123456789012345678901234567890x' as varchar));
2> go

------------------------------------------------------------
123456789012345678901234567890123456789012345678901234567890

```

Notice the `x` (at position 31) is included:
```
1> select concat(cast('123456789012345678901234567890x' as varchar(max)),cast('123456789012345678901234567890x' as varchar(max)));
2> go

----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
123456789012345678901234567890x123456789012345678901234567890x
```

nehanene15:
This will involve updating the following line to reference VARCHAR(max) instead: https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/third_party/ibis/ibis_mssql/datatypes.py#L34

nj1973:
I've attempted to reproduce this with these columns:
```
,   col_nvarchar      nvarchar(max)
,   col_nvarchar_500  nvarchar(500)
```

And I do not see any CAST to VARCHAR in a row validation SQL:
```
data-validation -v validate row -sc sqlserver -tc sqlserver -tbls dvt_test.dvt_mssql2pg_types \
--primary-keys=id --hash='id,col_nvarchar_500,col_nvarchar'
...
09/15/2023 10:03:11 AM-INFO: -- ** Source Query ** --
09/15/2023 10:03:11 AM-INFO: WITH t0 AS
[...]

PR #1037, opened 2023-11-08
fix: Cast treats nullable and non-nullable columns in the same way

### LINKED ISSUE #1036: row validation: Inconsistency between nullable and non-nullable columns
Inconsistent behaviour for nullable vs non-null columns, this is caused by this expression in `ibis_addon/api.py`:
```
    if op.to == self.type() and not op.to.is_timestamp():
         # noop case if passed type is the same
         return self
```
`self.type()` is an Ibis type expression that internally holds a value prefixed with `!` for not-null columns, e.g. `!string` for String columns.
`op.to` is always the nullable variant.
Because of this only nullable columns drop into the no-op branch and bypass a cast to string.

We could fix this with a simple workaround (aka hack) like this:
```
     """
+    def same_type(from_type, to_type) -> bool:
+        # The data type for Non-nullable columns if prefixed with "!", this is causing deviations
+        # between nullable and non-nullable columns. The second comparison below is catering for this.
+        return bool(
+            from_type == to_type
+            or str(from_type).lstrip("!") == str(to_type).lstrip("!")
+        )
+
     # validate
     op = ops.Cast(self, to=target_type)

-    if op.to == self.type() and not op.to.is_timestamp():
+    if same_type(op.to, self.type()) and not op.to.is_timestamp():
         # noop case if passed type is the same
```

nj1973:
This was spun off from issue https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/990.

Previously we were seeing a cast of `string` to `string` being bypassed because it is the same type but `!string` to `string` being retained in the final SQL.

With this commit both will bypass the cast.

### DIFF
--- a/third_party/ibis/ibis_addon/api.py
+++ b/third_party/ibis/ibis_addon/api.py
@@ -23,10 +23,19 @@ def cast(self, target_type: dt.DataType) -> Value:
     """Override ibis.expr.api's cast method.
     This allows for Timestamp-typed columns to be cast to Timestamp, since Ibis interprets some similar but non-equivalent types (eg. DateTime) to Timestamp (GitHub issue #451).
     """
+
+    def same_type(from_type, to_type) -> bool:
+        # The data type for Non-nullable columns if prefixed with "!", this is causing deviations
+        # between nullable and non-nullable columns. The second comparison below is catering for this.
+        return bool(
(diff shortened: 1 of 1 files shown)


nehanene15 (approved):
LGTM!

PR #1047, opened 2023-11-15
fix: --dry-run for SQLAlchemy clients with valid raw SQL

nehanene15:
Previously, the `--dry-run` flag, similar to the `--verbose` flag, didn't print the exact raw query w/client specific rendering for SQLAlchemy clients. This PR corrects that and adds tests for one non-SQLAlchemy backend (BigQuery) and one SQLAlchemy backend (MySQL).

This prevents rendering of `:param1` parameters and invalid quotes or backticks when printing a dry run.
This PR does not affect the current `-verbose` functionality since that flag should only be used for debugging purposes.

### DIFF
--- a/tests/system/data_sources/test_bigquery.py
+++ b/tests/system/data_sources/test_bigquery.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import json
 import os
 from unittest import mock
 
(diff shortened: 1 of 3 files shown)


tests/system/data_sources/test_bigquery.py
sundar-mudupalli-work, line 1: At line 1160, can you assert that "err" is "". If this test errors out, the error message will be printed, so we can find out the problem.

tests/system/data_sources/test_mysql.py
sundar-mudupalli-work, line 1: Same comment as above regarding err.

Overall looks good - we also need to test for all the databases ? Of course the test will run long and it will be worth it if it catches any errors.

sundar-mudupalli-work (changes requested):
Looks Good Neha - would consider adding tests for other databases as well.

tests/system/data_sources/test_mysql.py
nehanene15: Good catch for handling the err - I'll add that. I think testing for all databases may be redundant - the goal of the tests here is to ensure the params/backtick syntax is properly rendered for SQLAlchemy clients + non-SQLAlchemy clients. We know the query itself is valid for all databases since our other tests assert successful validation.

Therefore, by testing on BQ (non Alchemy) and MySQL (Alchemy) I feel we cover both use cases.

sundar-mudupalli-work (approved):

Issue #756, opened 2023-03-02
Issue with environment variable PSO_DV_CONFIG_HOME


madhu-jain:
Description : 

While using environment variable PSO_DV_CONFIG_HOME to fetch connection details and configuration yaml from GCS bucket, it expects 2 folders : connections and validations.

Given connection json would have sensitive information like password, it's not advisable to store connection json in GCS bucket instead create it at run time.
But when using variable PSO_DV_CONFIG_HOME, it seems it expect both connection json and config yaml reside in GCS bucket ie. both connections and validations are tightly coupled. 
Is it possible to decouple them?

nehanene15:
Closed by #1129

PR #1058, opened 2023-11-23
feat: Support for Kubernetes

sundar-mudupalli-work:
Hi,

This merge includes an upgrade to enable DVT to be orchestrated using Kubernetes jobs. See below
**feat**: _Horizontal scaling row validations via Kubernetes_ - For how this can be used see [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/tree/K8-indexed#scaling-dvt-to-run-10s-to-1000s-of-validations-concurrently). For design details, please [see here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/K8-indexed/docs/internal/kubernetes_jobs.md#scaling-data-validation-with-kubernetes-jobs)

Please review and set up time with me if you have any questions.

Thank you.

Sundar Mudupalli


docs/internal/kubernetes_jobs.md
```diff
@@ -0,0 +1,45 @@
...
+
+I am proposing a simple command line change to resolve this issue. Whenever a connection parameter is specified, allow the user to optionally specify a secret manager (provider, project-id). If a secret manager is specified, then DVT retrieves the connection information directly from the secret manager at the time of creating the connection. This is the recommended approach to handle secrets as opposed to mounting secrets as volumes. Cloud Run also has a limitation that multiple secrets [cannot be mounted with the same path](https://cloud.google.com/run/docs/configuring/services/secrets#disallowed_paths_and_limitations). Since DVT requires connections to two different databases with the connection info being mounted in the same directory, i.e. `$HOME/.config/google-pso-data-validator`, DVT cannot run within Cloud Run without this change. With this change, DVT can be run in a container in Cloud Run or Kubernetes fetching the connection information from the GCP Secret Manager.
```
nehanene15: Introducing the secret-manager-project/type for every validation command removes the de-coupling of connection configuration and validation configuration. I understand the issue here is storing connection configs locally, which would be inconvenient in Kubernetes. 
But why can't we use GCS connections + Secret Manager to get the correct config at runtime? We would only need to set the PSO_CONFIG_DIR env variable to enable GCS connections.

sundar-mudupalli-work: Neha,

I had not thought of that as an option initially. There are some challenges:

1. The directory structure under PSO_DV_CONFIG is two directories, one `connections` with connection information and another directory `validations` with yaml files. When we look for yaml files we only look for yaml files in the current directory - so in this case under `validations`. The way `generate-table-partitions` works, it creates yaml files under the _config-dir/<schema_name>.<table_name>_. Therefore PSO_DV_CONFIG cannot be directly used - we would need to copy files after running `generate-table-partitions`.
2. We have an [open security issue](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/756) with using GCS (or file system) for connection information.
3. Database passwords and connection information are secrets and the proposed implementation is simpler than the current implementation which requires multiple secrets per connection.

Hope that helps.

Sundar Mudupalli

nehanene15: Yeah - Issue #756 which you mentioned involves de-coupling `connections/` and `validations/`. I think we need to address this issue before we move forward with Kubernetes or other distributed validations. This addresses your first point as well - theoretically `generate-table-partitions -cdir my-table/validations/` should drop the YAML in the right directory.

I agree with your proposed change to make the entire connection JSON a secret rather than each field in the JSON - we should move forward with that. I just think we need to reconsider adding secret manager flags to each command and instead use GCS based connections, even if we need to address Issue 756 first.

README.md
```diff
@@ -411,12 +415,14 @@ For example, this flag can be used as follows:
 }
 ```
 
-### YAML Configuration Files
+### Running DVT with YAML Configuration Files
 
-You can customize the configuration for any given validation by providing use
-case specific CLI arguments or editing the YAML configuration file.
+Running DVT with YAML configuration files is the recommended approach if:
+* you want to customize the configuration for any given validation by providing use case specific CLI arguments or editing the YAML configuration file OR
```
nehanene15:
```suggestion
* you want to customize the configuration for any given validation OR
```

Suggest making this simpler since it may be confusing to mention "CLI arguments" in this section

README.md
```diff
@@ -411,12 +415,14 @@ For example, this flag can be used as follows:
...
+* you want to run DVT at scale (i.e. row validations across many partitions)
+Nearly every validation command can take the argument `-c <file-name>` (the `generate-table-partitions` command takes  `-cdir <directory-name>`) where one or more yaml files are produced. These yaml files can be modified (for more specific validations), see below
```
nehanene15:
```suggestion
Nearly every validation command can take the argument `-c <file-name>` (the `generate-table-partitions` command takes  `-cdir <directory-name>`) where one or more yaml files are produced. These yaml files can be modified for custom validations.
```

README.md
```diff
@@ -441,6 +447,24 @@ OR
...
+
+By default, each partition validation is retried upto 3 times if there is any error. In Kubernetes and Cloud Run, you can set the parallelism to the number you want. If the parallelism is set to too high a number, then the databases being queried can become overloaded, resulting in timeouts and errors. If you are validating 1000's of partitions in parallel, you may find that setting the parallelism too high (say 100) may result in timeouts and slow down the validation.
```
nehanene15:
```suggestion
By default, each partition validation is retried up to 3 times if there is an error. In Kubernetes and Cloud Run, you can set the parallelism to the number you want. Keep in mind that if you are validating 1000's of partitions in parallel, you may find that setting the parallelism too high (say 100) may result in timeouts and slow down the validation.   
```

README.md
```diff
@@ -441,6 +447,24 @@ OR
...
+
+By default, each partition validation is retried upto 3 times if there is any error. In Kubernetes and Cloud Run, you can set the parallelism to the number you want. If the parallelism is set to too high a number, then the databases being queried can become overloaded, resulting in timeouts and errors. If you are validating 1000's of partitions in parallel, you may find that setting the parallelism too high (say 100) may result in timeouts and slow down the validation.   
 
 View the complete YAML file for a Grouped Column validation on the
```
nehanene15: Lines 469 - 472 should be moved up before the 'Scaling DVT to run 10's to 1000's of validations concurrently' section to line 449.

README.md
```diff
@@ -441,6 +447,24 @@ OR
...
+
+DVT validations can be run concurrently (horizontal scaling) using GKE (Kubernetes Jobs) or Cloud Run Jobs, GCP's managed Kubernetes offering. In order to run DVT in a container, you have to build a docker image, [see instructions](https://github.com/GoogleCloudPlatform/professional-services-data-validator/tree/develop/samples/docker#readme). Set the `PSO_DV_CONFIG` environment variable to point to a GCS prefix where the connection configuration files are stored. In Cloud Run you can use the option `--set-env-vars` or `--update-env-vars` to pass [the environment variable](https://cloud.google.com/run/docs/configuring/services/environment-variables#setting). We recommend that you use the `bq-result-handler` to save your validation results. In order to validate partitions concurrently, run DVT in Kubernetes or Cloud Run as shown below:
+```
+data-validation (--verbose or -v) (--log-level or -ll) configs run
+  [--kube-completions or -kc]
```
nehanene15: Let's add a description to the right of `-kc` flag similar to how we do in `-cdir` i.e "Indicate that validations are running in parallel Kubernetes jobs"

Also, let's clarify: should this flag should only be used for Kubernetes or for Cloud Run as well?

data_validation/__main__.py
```diff
@@ -314,14 +314,51 @@ def build_config_managers_from_args(
...
+            )
+            config_file_path = (
+                f"{args.config_dir}{job_index:04d}.yaml"
+                if args.config_dir.endswith("/")
+                else f"{args.config_dir}/{job_index:04d}.yaml"
+            )
+            setattr(args, "config_dir", None)
+            setattr(args, "config_file", config_file_path)
+            config_managers = build_config_managers_from_yaml(args, config_file_path)
+        else:
```
nehanene15, line 347: Maybe we want to convert this into an elif block?
```
elif args.kube_completions:
   # log warning
else:
   # create state manager and config_managers
```

docs/internal/kubernetes_jobs.md
```diff
@@ -0,0 +1,17 @@
+# Scaling Data Validation with Kubernetes Jobs 
+
+## Nature of Data Validation
+Data Validation by nature is a batch process. We are presented with a set of arguments, the validation is performed, the results are provided and Data Validation completes. Data Validation can also take time (multiple secs, minutes) if a large amount of data that needs to be validated.
```
nehanene15:
```suggestion
Data Validation by nature is a batch process. We are presented with a set of arguments, the validation is performed, the results are provided and Data Validation completes. Data Validation can also take time if a large amount of data needs to be validated. 
```

README.md
```diff
@@ -441,6 +447,24 @@ OR
...
+  --config-dir or -cdir GCS_DIRECTORY 
+              where GCS_DIRECTORY = CONFIG_DIR/SOURCE_SCHEMA.SOURCE_TABLE, where CONFIG_DIR, SOURCE_SCHEMA and SOURCE_TABLE are
+              the arguments provided to generate-table-partitions to generate the partition yamls. GCS_DIRECTORY is the directory
+              where the partition files numbered 0000.yaml to <partno-1>.yaml are stored.
+```
+
+The Cloud Run and Kubernetes pods must run in a network with access to the database servers. Every Cloud Run job is associated with a [service account](https://cloud.google.com/run/docs/securing/service-identity). You need to ensure that this service account has access to Google Cloud Storage (to read connection configuration and yaml files) and BigQuery (to publish results). If you are using Kubernetes, you will need to use a service account with the same privileges as mentioned for Cloud Run. In Kubernetes, you will need to set up [workload identity](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity) so the DVT container can impersonate the service account.
```
nehanene15, line 470: Let's reference your docs/internal/kubernetes_jobs somewhere in this section.

docs/internal/kubernetes_jobs.md
```diff
@@ -0,0 +1,17 @@
+# Scaling Data Validation with Kubernetes Jobs 
+
+## Nature of Data Validation
+Data Validation by nature is a batch process. We are presented with a set of arguments, the validation is performed, the results are provided and Data Validation completes. Data Validation can also take time (multiple secs, minutes) if a large amount of data that needs to be validated. 
+
+Data Validation has the `generate-table-partitions` function that partitions a row validation into a specified number of smaller, equally sized validations. Using this feature, validation of two large tables can be split into a number of row validations of partitions of the same tables. See [partition table PRD](partition_table_prd.md) for details on partitioning. This process generates a sequence of yaml files which can be used to validate the individual partitions.
```
nehanene15:
```suggestion
Data Validation has the `generate-table-partitions` function that partitions a row validation into a specified number of smaller, equally sized validations. Using this feature, validation of one large table can be split into a number of row validations on partitions of the same table. See [partition table PRD](partition_table_prd.md) for details on partitioning. This process generates a sequence of yaml files which can be used to validate the individual partitions. 
```

docs/internal/kubernetes_jobs.md
```diff
@@ -0,0 +1,17 @@
...
...ts different types of workloads including a few batch workload types. The Job workload is a batch workload that retries execution until a specified number of them successfully complete. If a row validation has been split into `n` partitions, then we need to validate each partition and merge the results of the validation. Using Kubernetes Jobs we need to successfully run `n` completions of the job, as long as we guarantee that each completion is associated with a different partition. Kubernetes provides a type of job management called indexed completions that supports the Data Validation use case. A Kubernetes job can use multiple parallel worker processes. Each worker process has an index number that the control plane sets which identifies which part of of the overall task (i.e. which partition) to work on. The index is available in the environment variable `JOB_COMPLETION_INDEX` (in cloud run the environment variable is `CLOUD_RUN_TASK_INDEX`). An explanation of this is provided in [Introducing Indexed Jobs](https://kubernetes.io/blog/2021/04/19/introducing-indexed-jobs/#:~:text=Indexed%20%3A%20the%20Job%20is%20considered,and%20the%20JOB_COMPLETION_INDEX%20environment%20variable).
```
nehanene15:
```suggestion
Kubernetes supports different types of workloads including a few batch workload types. The Job workload is a batch workload that retries execution until a specified number of them successfully complete. If a row validation has been split into `n` partitions, then we need to validate each partition and merge the results of the validation. Using Kubernetes Jobs we need to successfully run `n` completions of the job, as long as we guarantee that each completion is associated with a different partition. Kubernetes provides a type of job management called indexed completions that supports the Data Validation use case. A Kubernetes job can use multiple parallel worker processes. Each worker process has an index number that the control plane sets which identifies which part of of the overall task (i.e. which partition) to work on. The index is available in the environment variable `JOB_COMPLETION_INDEX` (in Cloud Run the environment variable is `CLOUD_RUN_TASK_INDEX`). An explanation of this is provided in [Introducing Indexed Jobs](https://kubernetes.io/blog/2021/04/19/introducing-indexed-jobs/#:~:text=Indexed%20%3A%20the%20Job%20is%20considered,and%20the%20JOB_COMPLETION_INDEX%20environment%20variable).
```

nehanene15: We also may want to split this into multiple lines so it's easier to comment on.

docs/internal/kubernetes_jobs.md
```diff
@@ -0,0 +1,17 @@
...
+
+Indexed completion mode supports partitioned yaml files generated by `generate-table-partitions` in Data Validation, if each worker process ran only the yaml file corresponding to its index. I have an introduced an optional parameter `--kube-completions` or `-kc`. When this flag is used with `data-validation configs run` with a config directory and the container runs in indexed jobs mode, each container only processes the specific validation yaml file corresponding to its index. If the flag is used `data-validation configs run` with a config directory and DVT is not running in indexed jobs mode, a warning is issued. In all other instances, this flag is ignored.
```
nehanene15:
```suggestion
Indexed completion mode supports partitioned yaml files generated by the `generate-table-partitions` command if each worker process runs only the yaml file corresponding to its index. The`--kube-completions` or `-kc` flag when running configs from a directory indicates that the validation is running in indexed jobs mode on Kubernetes or Cloud Run. This way, each container only processes the specific validation yaml file corresponding to its index. 
```

docs/internal/kubernetes_jobs.md
```diff
@@ -0,0 +1,17 @@
...
+### IAM Permissions
```
nehanene15: Let's delete the IAM Permissions and Future Work sections since they are empty.

docs/internal/kubernetes_jobs.md
```diff
@@ -0,0 +1,17 @@
...
+### Passing database connection parameters
```
nehanene15, line 12: Is this section accurate? I know we plan on creating a separate PR that tackles Secret Manager support. Until then, users should be able to use GCS connections for Kubernetes/Cloud Run orchestration.

nehanene15 (commented):
Looks great overall, just left comments on documentation!

docs/internal/kubernetes_jobs.md
sundar-mudupalli-work: I modified this slightly differently. Please see text.

docs/internal/kubernetes_jobs.md
sundar-mudupalli-work: If I split into multiple lines, then each line becomes its own paragraph and there is a lot of white space in the document. Do we really want to do this for an internal doc?

docs/internal/kubernetes_jobs.md
sundar-mudupalli-work: Updated document and addded things to Future Work regarding using secret manage and other clieanup

docs/internal/kubernetes_jobs.md
sundar-mudupalli-work: Done.

README.md
sundar-mudupalli-work: Done

sundar-mudupalli-work:
@nehanene15 - pleaes take a look. I have updated the document(s).

README.md
```diff
@@ -446,6 +452,25 @@ View the complete YAML file for a Grouped Column validation on the
...
+
+DVT validations can be run concurrently (horizontal scaling) using GKE (Kubernetes Jobs) or Cloud Run Jobs. In order to run DVT in a container, you have to build a docker image, [see instructions](https://github.com/GoogleCloudPlatform/professional-services-data-validator/tree/develop/samples/docker#readme). Set the `PSO_DV_CONFIG_HOME` environment variable to point to a GCS prefix where the connection configuration files are stored. In Cloud Run you can use the option `--set-env-vars` or `--update-env-vars` to pass [the environment variable](https://cloud.google.com/run/docs/configuring/services/environment-variables#setting). We recommend that you use the `bq-result-handler` to save your validation results. In order to validate partitions concurrently, run DVT in Kubernetes or Cloud Run as shown below:
+```
+data-validation (--verbose or -v) (--log-level or -ll) configs run
+  [--kube-completions or -kc]
```
nehanene15:
```suggestion
  [--kube-completions or -kc] Specifies that validation is being run in Kubernetes or Cloud Run in indexed completion mode.
```

When documenting the command itself, we should keep it short. The paragraphs below should link to the kubernetes docs so the user knows how to set the completions in GKE.

README.md
nehanene15: We can mention in this section something like: "To read more, see docs/internal/kubernetes_jobs"

docs/internal/kubernetes_jobs.md
nehanene15: I think we should get rid of the 'Future Work' section personally since it doesn't belong in the public documentation.. these are internal roadmap items that we can track with issues.

If you want to keep it in, we should at least delete the line "This inconsistent behavior is challenging and should be fixed." since it doesn't belong in a product's public docs

sundar-mudupalli-work:
Done. PTAL

nehanene15 (approved):
There is still an unresolved comment - I would remove the 'Future Work' section or delete the line "This inconsistent behavior is challenging and should be fixed" since we shouldn't point out bugs in our public docs.

I'll approve it since the next PR to fix the GCS configs will delete this section in the docs anyways

PR #1078, opened 2024-01-16
fix: Better casts to string for binary floats/doubles

### LINKED ISSUE #841: validate row: Binary float and double values in scientific notation which does not match BigQuery
Example Oracle columns:

```
,   col_float32     BINARY_FLOAT
,   col_float64     BINARY_DOUBLE
```

Example BigQuery columns:

```
,   col_float32     FLOAT64
,   col_float64     FLOAT64
```

Concat comparison output:

```
╤════════════════════════════════════════╤════════════════════╤
│ source_agg_value                       │ target_agg_value   │
╪════════════════════════════════════════╪════════════════════╪
│ 1.23456102E+0051.23456781E+007         │ 123456.112345678.1 │
┼────────────────────────────────────────┼────────────────────┼
│ 1.23456203E+0051.2345678199999999E+007 │ 123456.212345678.2 │
┼────────────────────────────────────────┼────────────────────┼
```

Actual Oracle values:

```
COL_FLOAT32		  COL_FLOAT64
------------------------- -------------------------
123456.102		  12345678.1
123456.203		  12345678.199999999
```

This is also an issue on Hive, SQL Server and Teradata.

When this has been fixed we should edit the files below and search for "issue-841":

- test_oracle.py
- test_hive.py
- test_teradata.py
- test_sql_server.py

nj1973:
This PR is to work around problems validating floating point values where some engines return all digits and others return scientific notation, for example:
```
1.23456781E+007         │ 12345678.1
```

Changes in the PR:

- Use SQL Server FORMAT() instead of CAST() for real/float columns. In my tests that ensures we do not end up with scientific notation
- Use Teradata TO_CHAR('TM9') instead of CAST() for binary_float/double columns. In my tests that ensures we do not end up with scientific notation
- Use Oracle TO_CHAR('TM9') instead of CAST() for binary_float/double columns. In my tests that ensures we do not end up with scientific notation but we still get some unpredicatable values due to the lossy nature of binary floats. For this reason the column remain excluded from Oracle integration tests
- Reduces code duplication by importing the original Ibis `_cast()` function and calling that when not in one of our workarounds

Problems not fixed by this PR:

- Hive float and double remain problematic
- Oracle binary_float/double remain problematic

### DIFF
--- a/tests/system/data_sources/test_hive.py
+++ b/tests/system/data_sources/test_hive.py
@@ -312,7 +312,7 @@ def disabled_test_row_validation_core_types():
 )
 def test_row_validation_core_types_to_bigquery():
     parser = cli_tools.configure_arg_parser()
-    # TODO Change --hash string below to include col_float32,col_float64 when issue-841 is complete.
+    # col_float64 is excluded below because there is no way to control the format when casting to string.
     args = parser.parse_args(
         [
             "validate",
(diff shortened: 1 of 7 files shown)


nehanene15 (approved):
Looks great! I like the idea of importing sa_fixed_cast() to clean up the code

PR #1083, opened 2024-01-31
feat: Add --url to Oracle connections add options

### LINKED ISSUE #1003: connections add: Does not support Oracle wallet
It is possible to configure DVT to use an Oracle wallet for authentication, but the options supported by `data-validation connections add Oracle` do not cater for it.

To support a wallet I needed to use the `url` config option which I cannot supply via the command line.
```
$ cat ~/.config/google-pso-data-validator/ora_local_wallet.connection.json
{"source_type": "Oracle", "url": "oracle+cx_oracle://@system_xepdb"}
```

There are two potential ways to support Oracle wallet:

1. Enhance by documentation. Simply add a section to the Oracle connections documentation (https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/docs/connections.md#oracle) detailing that the user must manually create the json file.
2. Enhance DVT to support an option for the connection url. It appears we already support it for DB2.

I propose to attempt option 2 above and if it proves problematic then we can fall back on option 1.

nj1973:
This small change enables support via the CLI for Oracle Wallet based connections. The support was already there if we manually created the connection json file, this just exposes an option and add some guidance to the docs.

### DIFF
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -74,6 +74,7 @@ CONNECTION_SOURCE_FIELDS = {
         ["user", "User used to connect"],
         ["password", "Password for supplied user"],
         ["database", "Database to connect to"],
+        ["url", "Oracle SQLAlchemy connection URL"],
     ],
     "MSSQL": [
         ["host", "Desired SQL Server host (default localhost)"],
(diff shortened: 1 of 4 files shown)


nehanene15 (approved):
Looks great! Did you use a linter for markdown files? If so, maybe we can consider adding that to our lint tests.

nj1973:
> Looks great! Did you use a linter for markdown files? If so, maybe we can consider adding that to our lint tests.

No I didn't use a linter, I edited in VS Code and it looks like it trimmed trailing white space, hence the changes on lines I didn't actually edit.

PR #1089, opened 2024-02-05
feat: Add support to generate a JSON config file only for applications purposes

### LINKED ISSUE #908: Add support to generate a JSON config file
Similar to the way we output a YAML config file, we could provide a JSON output that can be usable if someone wants to use our library in a Python code directly and not the CLI

Related #901,  #924, #925, #940

helensilva14:
Fixes #908 

New optional CLI flag parameter for any type of validation: 

`[--config-file-json or -cj CONFIG_FILE_JSON]`
     JSON Config File Path to be used for storing validations for application purposes, since we only support generating a JSON config file and all other features are exclusive for YAML config files.

/gcbrun


noxfile.py
```diff
@@ -69,7 +69,10 @@ def unit(session):
         "--cov-config=.coveragerc",
         "--cov-report=term",
         os.path.join("tests", "unit"),
-        env={"PSO_DV_CONFIG_HOME": ""},
+        env={
+            "PSO_DV_CONFIG_HOME": "",
+            "PROJECT_ID": os.environ.get("PROJECT_ID", "pso-kokoro-resources"),
```
nehanene15: I think this may be causing the nox tests to fail. Is this required for the new tests? Usually within the unit test we get the PROJECT_ID env var if needd

helensilva14: I wasn't sure how would be the best way to provide the env variable so I followed something in that sense that we've for other test. But it makes sense, I'll try provide it just for that specific unit test.

tests/unit/test__main.py
```diff
@@ -19,7 +19,9 @@
 
 from data_validation import cli_tools, consts
 from data_validation import __main__ as main
+from tests.system.data_sources.test_bigquery import BQ_CONN
 
+PROJECT_ID = os.getenv("PROJECT_ID")
```
nehanene15: Where do you use this variable?

helensilva14: Only BQ_CONN variable that's been imported needs it, actually

README.md
```diff
@@ -190,7 +193,10 @@ data-validation (--verbose or -v) (--log-level or -ll) validate row
                         If target filter is not provided, the source filter will run on source and target tables.
                         See: *Filters* section
   [--config-file or -c CONFIG_FILE]
-                        YAML Config File Path to be used for storing validations.
+                        YAML Config File Path to be used for storing validations and other features.
+                        See: *Running DVT with YAML Configuration Files* section
+  [--config-file-json or -cj CONFIG_FILE_JSON]
+                        JSON Config File Path to be used for storing validations for application purposes, since we only support generating a JSON config file and all other features are exclusive for YAML config files.
```
nehanene15: We should probably keep this short since it's a CLI command doc.
```suggestion
                        JSON Config File Path to be used for storing validations only for application purposes.
```

data_validation/cli_tools.py
```diff
@@ -922,7 +922,12 @@ def _add_common_arguments(optional_arguments, required_arguments):
     optional_arguments.add_argument(
         "--config-file",
         "-c",
-        help="Store the validation in the YAML Config File Path specified",
+        help="Store the validation config in the YAML File Path specified",
+    )
+    optional_arguments.add_argument(
+        "--config-file-json",
+        "-cj",
+        help="Store the validation config in the JSON File Path specified to be used for application purposes, since we only support generating a JSON config file and all other features are exclusive for YAML config files",
```
nehanene15: Here too - I think the lengthy description can be confusing... maybe we stick to "Store the validation config in the JSON File Path specified for application use cases"

samples/airflow/README.md
```diff
@@ -15,19 +15,14 @@ By default, the DAG will output the results to BigQuery as a result handler.
...
 
-Parameters in a typical YAML config file are slightly different from the raw JSON config, 
-which is generated after DVT parses the YAML. The [build_config_manager()](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/data_validation/config_manager.py#L429) 
-method generates the JSON config and should be used as a reference.
-
-Our Cloud Run sample also expects a raw JSON config in the `'data'` variable shown
-[here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/tree/develop/samples/run#test-cloud-run-endpoint).
+You can get the JSON content specific for your scenario by using our CLI and providing the argument to generate the JSON config file [`--config-file-json` or `-cj <filepath>.json`]. IMPORTANT: do not forget to make the necessary adjustments between JSON and Python objects, check [this link as a reference](https://python-course.eu/applications-python/json-and-python.php).
```
nehanene15, line 25: Does Cloud Run/Airflow require the JSON to be converted to a Python object? I think the request is expected to be JSON... in that case we can remove the disclaimer.

nehanene15 (commented):
Left a couple comments! Were the `samples/tests/` files stale? Saw that those were deleted in this PR

helensilva14:
> Left a couple comments! Were the `samples/tests/` files stale? Saw that those were deleted in this PR

Yes, they were not being used at all and the directory name was quite misleading, as if it had sample code to create test methods using DVT or something like that.

samples/airflow/README.md
helensilva14: They don't necessarily require the conversion, but in both of our samples (Cloud Run, Airflow) we're directly initializing/stating the JSON content inside the Python code as a Dict object. So in these cases is necessary to perform the data types adjustments, but the user can always provide the pure JSON if they want to by using the `json.loads()` method or something similar.

nehanene15 (approved):

PR #1095, opened 2024-02-14
fix: Add support for cx_Oracle's DB_TYPE_LONG_RAW

### LINKED ISSUE #1094: DVT is failing with a "NotImplementedError: Oracle type <cx_Oracle.DbType DB_TYPE_LONG_RAW> is not supported" (Oracle to AlloyDB)
I'm running DVT with Oracle source system and AlloyDB (postgres) destination system. For a few of the tables, DVT is not running and giving the error `NotImplementedError: Oracle type <cx_Oracle.DbType DB_TYPE_LONG_RAW> is not supported`. I have confirmed that table exists in both the source and destination databases. The paths are also correct because DVT is running well for most of the other tables in the schema.

The table contains a string column and a column which contains Long Raw (oracle)/Binary data (alloydb). As per the the [PostgreSQL documentation](https://cx-oracle.readthedocs.io/en/latest/api_manual/module.html#db-api-types), DB_TYPE_LONG_RAW seems to be related to Binary data. 

Can you please help with the issue and give a direction as to what DVT requires in order to handle raw/binary data type? 

Table schema (oracle)
image

Table schema (alloydb)
image

Snapshot of the error log when running the DVT command
image

helensilva14:
Closes #1094 

Reference: https://cx-oracle.readthedocs.io/en/latest/api_manual/module.html#cx_Oracle.DB_TYPE_NCHAR

/gcbrun


nj1973:
I don't think adding LONG RAW to `_type_mapping` is enough on its own.
I suspect we also need to register the type like we do for other data types, however this is not simple because SQL Alchemy doesn't appear to have LONG RAW support either.

nj1973:
The changes in this PR prevent an exception being throw when validating a table that has a LONG RAW column, they do not add functionality to validate the data in LONG RAW columns. That would be a different issue. I don't think row validation is possible on Oracle LONG [RAW] data. We might be able to make column validation work but, like I said, I think that should be a new issue.

nehanene15 (commented):
LGTM!

nj1973 (approved):
LGTM too!

PR #1104, opened 2024-02-23
fix: Add table names to report results when source and/or target dataframes are empty

### LINKED ISSUE #1098: Null table and column names on validation results table when source and/or target tables are empty
Related to #1006 

Validation status is returning the expected (fail), but to have a complete set of information these columns need to be populated and they are not being at the moment

helensilva14:
Closes #1098 

Partially fixing the issue, we still need to figure how to get the column names and most important, if we'll do this change

/gcbrun

### DIFF
--- a/data_validation/combiner.py
+++ b/data_validation/combiner.py
@@ -24,7 +24,6 @@ import json
 import logging
 import ibis
 import ibis.expr.datatypes
-import pandas as pd
 
 from data_validation import consts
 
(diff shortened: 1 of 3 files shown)


data_validation/combiner.py
```diff
@@ -101,9 +102,32 @@ def generate_report(
...
+    )
+
     return result_df
 
 
+def _get_table_name_when_null(
+    validation: metadata.ValidationMetadata, result_type: str
+) -> str:
+    value = validation.get_table_name(result_type)
+    return value if value is not None else np.nan
```
helensilva14: @nehanene15 I just noticed later I could've used `float("nan")` here instead, let me know what do you think it's better

nehanene15: We probably don't need another function for this since there isn't a case where this will be empty.
Instead we can just call `run_metadata.validations[first].get_table_name(consts.RESULT_TYPE_SOURCE)`

helensilva14: That was the initial approach I implemented that day and I thought the same, but actually the table names can be empty for custom-query validations, that's why a Spanner system test was failing

nehanene15: I see - in that case, to make it more consise we can just do `return validation.get_table_name(result_type) or None`

nehanene15 (approved):

PR #1108, opened 2024-03-05
feat: Support GCS files in configs list command

### LINKED ISSUE #1105: Update configs list & configs get to support GCS paths
Currently `data-validation configs list` and `data-validation configs get` are coupled with the environment variable `PSO_DV_CONFIG_HOME`. If this variable points to a GCS directory, the above commands will search within GCS. 

We aim to de-couple the storage of connections and configs within the same GCS root directory and deprecate usage of the env var for _configuration_files.

We should support the following:
`data-validation configs list [--dir gs://path/to/dir]`
`data-validation configs get --config-file gs://path/to/file.yaml`

The `gs://` prefix itself should indicate the file or directory is in GCS rather than stored locally.

nehanene15:
Closes #1105 

- After testing, `configs get` already supported GCS paths so this PR adds support for the `configs list` command
- Update documentation to make more concise and list out all `data-validation configs` options.


helensilva14 (approved):
LGTM!

PR #1126, opened 2024-04-26
fix!: Filters not working correctly in Snowflake

### LINKED ISSUE #1115: Filters not working correctly in Snowflake
There are some issues when comparing Snowflake to BigQuery or vice-versa.
It seems the where-condition does not work correctly to Snowflake.

Command I used for row validations:
```
  -sc bq_conn \
  -tc sf_conn \
  -tbls schema.TABLE1=schema.TABLE1 \
  --concat column1,column2,column3 \
  --primary-keys unique_key \
  --filters 'date(load_dt)=current_date':'date("LOAD_DT")=current_date' \
  --config-file filters.yaml
```

BigQuery:
```
SELECT t4.`concat__all`, t4.`unique_key`
FROM (
  SELECT t3.*,
         ARRAY_TO_STRING([t3.`upper__rstrip__ifnull__cast__column1`, t3.`upper__rstrip__ifnull__cast__column2`, t3.`upper__rstrip__ifnull__cast__column3`], '') AS `concat__all`
  FROM t3
) t4
WHERE date(load_dt)=current_date
```
==> This work fine. There is t3.*, so load_dt is correctly queried.

Snowflake:
```SQL
SELECT t5."concat__all", t5."unique_key" 
FROM (SELECT t4."concat__all" AS "concat__all", t4."UNIQUE_KEY" AS "unique_key" 
FROM t4) AS t5 
WHERE date("LOAD_DT")=current_date
```
==> This does not work fine, as "LOAD_DT" is not queried from t4. 

DDL for the tables:
BigQuery:
```
create table TABLE1 (
   unique_key STRING,
   column1 STRING,
   column2 STRING,
   column3 STRING,
   load_dt DATETIME
)
```

Snowflake:
```
create table TABLE1 (
   unique_key STRING,
   column1 STRING,
   column2 STRING,
   column3 STRING,
   load_dt DATETIME
)
```

nj1973:
The PR makes the following changes:

1) Adds `--filters` tests for `column` and `row` validations across all engines. The filter includes the id column and a comparison column.
2) Moves the application of filters from after application of calculated fields to before them. This has two benefits:
  2.1. We no longer rely on filtered columns bubbling up through the calculated field subqueries, we apply the filter to the actual table before doing anything else.
  2.2 Because we apply the filter to the actual table and not the subqueries we are giving the query optimizer less work to do in trying to push the filter down to the source table.

After adding the tests described in point 1 above all tests passed except for the Snowflake row validation test.

After adding the change described in point 2 above all tests pass.

I'm not sure whether we should flag this change as a breaking change or not. If any users are using filters on calculated fields then these will no longer work. Only filters on real columns are supported.


sundar-mudupalli-work:
Neil,

Looks good to me. Our documentation already says in the filters section `The filter is written in the syntax of the given source and must reference columns in the underlying table, not projected DVT expressions.`

I don't consider this a breaking change.

Sundar Mudupalli

sundar-mudupalli-work (approved):
Thank you.

nj1973:
> and must reference columns in the underlying table, not projected DVT expressions.

For info, I added the words `and must reference columns in the underlying table, not projected DVT expressions.` on this branch.

nehanene15 (approved):
LGTM! I think it's better to flag as a breaking change for any users that apply filters on calculated fields currently, even though this is an anti-pattern

helensilva14:
@nj1973 Please add `BREAKING CHANGE: for any users that apply filters on calculated fields currently, even though this is an anti-pattern` as a footer in the PR's commit message once we're able to merge it, thank you again for the awesome work!

Reference: https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-both--and-breaking-change-footer

helensilva14 (approved):

nj1973:
> @nj1973 Please add `BREAKING CHANGE: for any users that apply filters on calculated fields currently, even though this is an anti-pattern` as a footer in the PR's commit message once we're able to merge it, thank you again for the awesome work!
> 
> Reference: https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-both--and-breaking-change-footer

Of course I forgot to add the `BREAKING CHANGE` message to the PR's commit!

PR #1129, opened 2024-05-03
feat!: Support for GCS config paths decoupled from environment variables

### LINKED ISSUE #1106: Support saving validation configs to GCS with `--config-file` flag
Currently, we rely on the `PSO_DV_CONFIG_HOME` env var to generate config files in GCS with the `--config-file` flag. This can get confusing since the env var doesn't separate storage for connection and configs. 

DVT should support GCS config generation directly within the validate commands (column, row, custom-query, and schema). 
For example, `data-validation validate column ... -c gs://my/path/to/yaml`

nehanene15:
Fixes #1106 

This PR de-couples the usage of `PSO_DV_CONFIG_HOME` env var to manage _both_ connection and configuration files. Instead, the `--config-file` flag supports GCS paths. This will give users more autonomy over where their configs are stored. To reflect this, the StateManager class will only maintain state for getting/setting connections configs in GCS or locally.

Breaking changes:
- The env var `PSO_DV_CONFIG_HOME` is deprecated and renamed to `PSO_DV_CONN_HOME`. It will only manage connection configurations.
- For configurations in GCS, you will need to provide the GCS path directly with `--config-file gs://path/to/file.yaml` or `--config-dir gs://path/to/dir`


data_validation/cli_tools.py
```diff
@@ -1081,47 +1083,47 @@ def get_connection(connection_name):
...
+    elif validation_file_name.endswith("json"):
+        config_str = json.dumps(config)
+    else:
+        raise ValueError(f"Invalid validation file name: {validation_file_name}")
 
-def store_partition(target_file_path, yaml_config, target_folder_path=None):
-    """Store the partition YAML config under the given name."""
-    mgr = state_manager.StateManager(target_folder_path)
-    mgr.create_partition_yaml(target_file_path, yaml_config)
+    os.makedirs(os.path.dirname(validation_path), exist_ok=True)
```
nj1973: I'm not sure this line should be here.

If I write my configs to GCS this still creates an empty `gs:` directory in the CWD. Should this be moved into `gcs_helper` for local fs only?

README.md
```diff
@@ -194,7 +194,7 @@ data-validation (--verbose or -v) (--log-level or -ll) validate row
                         If target filter is not provided, the source filter will run on source and target tables.
                         See: *Filters* section
   [--config-file or -c CONFIG_FILE]
-                        YAML Config File Path to be used for storing validations and other features.
+                        YAML Config File Path to be used for storing validations and other features. Supports GCS and local paths.
                         See: *Running DVT with YAML Configuration Files* section
   [--config-file-json or -cj CONFIG_FILE_JSON]
```
nj1973, line 199: This is probably unrelated to your changes but noting it anyway. If I create a json config file, e.g.:
```
data-validation validate row -sc ora -tc pg \
-tbls=s.c --primary-keys id --hash '*' \
--filter-status fail -cj /tmp/c.json
```

I cannot validate with it:
```
data-validation configs run -cj /tmp/c.json
...
data-validation: error: unrecognized arguments: /tmp/c.json
```

Or am I using it incorrectly?

data_validation/cli_tools.py
```diff
@@ -1081,47 +1083,47 @@ def get_connection(connection_name):
...
 
 
-def get_validation(validation_name, config_dir=None):
-    """Return validation YAML for a specific connection."""
+def get_validation(name: str, config_dir: str = None):
+    """Return validation YAML config."""
     if config_dir:
-        mgr = state_manager.StateManager(file_system_root_path=config_dir)
-        return mgr.get_validation_config(validation_name, config_dir)
+        validation_path = os.path.join(config_dir, name)
```
nj1973, line 1103: Can a config dir also be in GCS?

data_validation/gcs_helper.py
```diff
@@ -0,0 +1,93 @@
+# Copyright 2023 Google LLC
```
nj1973: nit: Should this be 2024?

data_validation/cli_tools.py
```diff
@@ -1081,47 +1083,47 @@ def get_connection(connection_name):
...
-    mgr.create_validation_yaml(validation_file_name, yaml_config)
-
-
-def store_validation_json(validation_file_name, json_config):
-    """Store the validation JSON config under the given name."""
-    mgr = state_manager.StateManager()
-    mgr.create_validation_json(validation_file_name, json_config)
+def store_validation(validation_file_name, config, include_log=True):
+    """Store the validation config under the given name."""
+    validation_path = gcs_helper._get_validation_path(validation_file_name)
```
nj1973: If we are calling `_get_validation_path` from outside of its module (other than for testing) then it should probably be renamed to not have an underscore prefix.

data_validation/gcs_helper.py
```diff
@@ -0,0 +1,93 @@
...
+    except ValueError as e:
+        raise ValueError("GCS Path Failure {} -> {}".format(gcs_file_path, e))
+
+
+def _get_gcs_file_path(gcs_file_path: str) -> str:
+    """
+    Returns relative object file path i.e. `path/to/file.yaml`
+    given full GCS file path with prefix.
+    """
+    return str.join("", gcs_file_path[5:].split("/", 1)[1:])
```
nj1973: nit: I find joining to the string itself rather than via `str` neater, for example:
```
"".join(gcs_file_path[5:].split("/", 1)[1:])
```

But this is just preference so feel free to ignore.

data_validation/state_manager.py
```diff
@@ -236,42 +117,10 @@ def setup(self):
             if not os.path.exists(self._get_connections_directory()):
                 os.makedirs(self._get_connections_directory())
 
-    # GCS File Management Section
     def setup_gcs(self):
-        info = client_info.get_http_client_info()
-        self.storage_client = storage.Client(client_info=info)
         try:
-            self.gcs_bucket = self._get_gcs_bucket()
+            gcs_helper._get_gcs_bucket(self.file_system_root_path)
```
nj1973: Another example of external use of underscore function.

nj1973 (commented):
Just a few minor comments

README.md
nehanene15: Yeah, we only support saving to JSON, not running JSON through the CLI. JSON should only be used for Cloud Run or other application based use cases as a request body

data_validation/cli_tools.py
nehanene15: Yes! We already supported that before this PR though. That's why it was extra confusing - some commands supported direct GCS paths and others didn't 😅

data_validation/gcs_helper.py
nehanene15: I like this a lot better!

data_validation/cli_tools.py
nehanene15: good catch!!

nj1973 (approved):
LGTM.

Raniksingh (approved):

PR #1130, opened 2024-05-05
fix: Issue 1127 configs dir fails with more than 40 files

### LINKED ISSUE #1127: configs run -cdir fails when 40 or more files are present (with Oracle source connection)
Hi,

When I try `data-validation -v configs run -cdir test_yamls` and the directory has more than 40 files (with Oracle source connection), I get the error message
```
05/01/2024 03:55:17 AM-ERROR: Error Connection Type "Oracle" could not connect: (cx_Oracle.DatabaseError) ORA-04031: unable to allocate 226048 bytes of shared memory ("shared pool","unknown object","sga heap(1,0)","Fixed Uga")
(Background on this error at: https://sqlalche.me/e/14/4xp6) occurred while running config file row_access_control_code.yaml. Skipping it for now.
```
Upon review the issue seems to be in the code. What we do when presented with multiple files is to build a list  config_manager objects - one for each yaml file. After we have built the config managers for all the yaml files, we then execute the validation for each config manager. This is in lines 361-371 of the file `data_validation/__main__.py` as follows:
```
            for file in config_file_names:
                config_managers.extend(build_config_managers_from_yaml(args, file))
    else:
        if args.kube_completions:
            logging.warning(
                "--kube-completions or -kc specified, which requires a config directory, however a specific config file is provided."
            )
        config_file_path = _get_arg_config_file(args)
        config_managers = build_config_managers_from_yaml(args, config_file_path)

    run_validations(args, config_managers)
```
This all seems fine - until we have 40 yaml files. So we build 40 config manager objects before we execute any validation - sure not a problem? When we build a config manager object - we open a connection to the database - so we have 40 connections and have not run any validation. The Oracle library seems to have a limit of 39 or so connections and after that complains of running out of shared memory. It appears the right approach might be to build a config_manager object, run the validation and then repeat. The code is in `data_validation/__main__.py` - lines 374-386 as follows:
```
def build_config_managers_from_yaml(args, config_file_path):
    """Returns List[ConfigManager] instances ready to be executed."""
    if "config_dir" in args and args.config_dir:
        yaml_configs = cli_tools.get_validation(config_file_path, args.config_dir)
    else:
        yaml_configs = cli_tools.get_validation(config_file_path)

    source_conn = mgr.get_connection_config(yaml_configs[consts.YAML_SOURCE])
    target_conn = mgr.get_connection_config(yaml_configs[consts.YAML_TARGET])

    source_client = clients.get_data_client(source_conn)
    target_client = clients.get_data_client(target_conn)
```
Suggestions on how to fix ?

sundar-mudupalli-work:
HI,

data validator built all the configurations before calling a validation. "Building" a configuration involves creating a data_client, which connects to the database. When `data-validation` was called with `configs run -cdir <dirnname>` it opens a database connection for each file in the directory *before* calling the first validation.  This results in running out of Shared memory in the case of Oracle (limit seems to be 40 open connections)

Code changed to call a validation immediately, each time after reading the yaml file from the directory. Tested this change on the same directory causing issues with the develop DVT branch. This does not run into Shared memory issues on Oracle. However, not entirely clear how this works - as data-validation does not appear to `disconnect` from each Oracle connection.

Sundar Mudupalli


nehanene15:
LGTM. I'm assuming the connections get garbage collected after the client.execute() is called in SQLAlchemy. Probably something to investigate further in the future.

nehanene15 (approved):

PR #1135, opened 2024-05-15
feat: Add support for random row sampling on binary id columns

nj1973:
This PR adds support for random row sampling on binary id columns by:

1. Casting random binary values to hex
2. Adding a reverse cast back to binary in the IN list added for random binary id filters

Also adds random row options to DVT_BINARY integration tests. There are only 5 rows in the table as we validate with batch size 5 which at least proves that the filter works on each system. We probably need to add a more targeted random row integration tests in the future but that was not the purpose of this PR, I just want to know the casting works on all engines.

Unfortunately I had to disable the Hive test because our Hive test instance is not behaving properly when asked to process a multi-element IN list on a binary column. This reproduces outside of DVT.


nj1973:
We appear to have an issue on Hive, unrelated to our changes.

Single IN list:
```
select * from pso_data_validator.`dvt_binary` where `binary_id` in (unhex('4456542d6b65792d35'));

+-----------------------+--------------------+------------------------+
| dvt_binary.binary_id  | dvt_binary.int_id  | dvt_binary.other_data  |
+-----------------------+--------------------+------------------------+
| DVT-key-5             | 5                  | Row 5                  |
+-----------------------+--------------------+------------------------+
```

Add a second item to the IN list:
```
select * from pso_data_validator.`dvt_binary` where `binary_id` in (unhex('4456542d6b65792d35'),unhex('4456542d6b65792d33');

+-----------------------+--------------------+------------------------+
| dvt_binary.binary_id  | dvt_binary.int_id  | dvt_binary.other_data  |
+-----------------------+--------------------+------------------------+
+-----------------------+--------------------+------------------------+
```

No rows selected!

nehanene15 (approved):
LGTM. Nice job!!

PR #1138, opened 2024-05-20
feat: Control Teradata decimal format when cast to string

### LINKED ISSUE #1134: Row validation failing for TD DECIMAL(15,2) and BQ NUMERIC(15,2)
We have noticed that TD and BQ follow different approaches to store DECIMAL types and also yield different results when casting decimal types to string. 

## Example 1
* TD stores 44.00 as 44.00 
* BQ stores 44.00 as 44

## Example 2
* In TD `CAST(0.07 AS STRING) -> .07`
* In BQ `CAST(0.07 AS STRING) -> 0.07`

## Workaround 
We have been using a custom SQL with the following manipulations

#### Source custom SQL (TD)
The following will remove trailing ceros, e.g. 44.00 -> 44
`TRIM(TRAILING '.' FROM TRIM(TRAILING '0' FROM CAST(QUANTITY AS VARCHAR(15)))) AS QUANTITY`
#### Target custom SQL (BQ)
The following will remove leading ceros before the decimal separator, e.g. 0.07 -> .07
`REGEXP_REPLACE(CAST(QUANTITY AS STRING),r'^0+','') as QUANTITY`

### 1. New feature
Can DVT account for decimal to string casting inconsistencies during row validation? It's not ideal to create custom SQL for each table with decimals as this requires significant manual work. Resolving this issue would benefit future users by preventing them from encountering these inconsistencies and needing to create custom SQL solutions.

### 2. Workaround
Is there any other workaround less cumbersome than having to create custom sql for each table with decimals?

nj1973:
This PR adds a TO_CHAR format to Teradata NUMBERs/DECIMALs to work around two inconsistencies with other engines.

1) SELECT CAST(CAST(7 AS DECIMAL(5,2)) AS VARCHAR(64));
'7.00'
2) SELECT CAST(CAST(-0.7 AS DECIMAL(5,2)) AS VARCHAR(64));
'-.70'

After merging this PR we get these literals from the examples above:
1) '7'
2) '-0.7'


nehanene15 (approved):
LGTM!

Issue #1137, opened 2024-05-17
Increase max partiton generate-table-partitions max value for --partition-num


franklinWhaite:
Currently generate-table-partitions only allows for --partition-num up to 1000 partitions. Given that the recommended partition size is of ~50k rows each, we have a limitation when working with tables with more than 50,000,000 rows. 

At this moment we need to be able to handle tables of up to ~2 billion records (e.g. ~40k partitions)

helensilva14:
Related to #966

sundar-mudupalli-work:
Hi,

Neha is right. We can increase the number of partitions to 10K by changing the following two lines in [cli_tools.py](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/f51f32751bfb8713ac04de6456421a3d4a8ff1ce/data_validation/cli_tools.py#L983):
```
        choices=range(1, 1001),
        metavar="[1-1000]",
```
basically change the limits to 10001 and 10000.
To increase the number of partitions higher - say to 1M partitions - we would need to change the number of characters we use for the name of the yaml files. Currently we use 4 characters - so can do `0000.yaml` to `9999.yaml`. For 1M partitions we would need to use 6 characters, so that we can do `000000.yaml` to `999999.yaml`. I suggest that we pause on this change in DVT - since it is more complex. 
First, we would need to change the way we generate the file names. So this line in [partition_builder.py](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/f51f32751bfb8713ac04de6456421a3d4a8ff1ce/data_validation/partition_builder.py#L336):
```
                target_file_name = "0" * (4 - len(str(pos))) + str(pos) + ".yaml"
```
could be rewritten as:
```
                target_file_name = f"{pos:06d}.yaml"
```
We will also need a change in how we process files in directories if DVT is run via cloud run jobs or K8 - see [`__main__.py`](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/f51f32751bfb8713ac04de6456421a3d4a8ff1ce/data_validation/__main__.py#L351C1-L355C14). There we currently look for file names from `0000.yaml` to `9999.yaml`. We would need to see if the directory contains these files (old `generate-table-partitions` behavior) or from `000000.yaml` to `000000.yaml` (new `generate-table-partitions` behavior). That is these lines:
```
            config_file_path = (
                f"{args.config_dir}{job_index:04d}.yaml"
[...]

PR #1139, opened 2024-05-20
fix: Update to support up to 10K partitions

sundar-mudupalli-work:
Hi,

Per [comment](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1137#issuecomment-2120874591), I have updated the code to support 10K partitions. Supporting more than 10K partitions is a bit more complex as explained in that comment. It turned out to be a bit more complex than mentioned in the doc because the error message with `choices` lists all the choices as stated [here](https://bugs.python.org/issue16418#msg174947), so I have used the suggested approach for large number of values.

I suggest moving this to develop after review if it passes out integration tests. I have tested the code in my environment and it successfully produces 10K files and the first many work with `data-validation configs run -cdir`. So no issues there.

Thanks.

Sundar Mudupalli

### DIFF
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -992,7 +992,7 @@ def _add_common_partition_arguments(optional_arguments, required_arguments):
         help="Number of partitions/config files to generate, a number from 1 to 10,000",
         type=_check_no_partitions,
         metavar="[2-10000]",
-ls    )
+    )
 
     # Optional arguments
     optional_arguments.add_argument(
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -955,7 +955,7 @@ def _add_common_arguments(optional_arguments, required_arguments):
     )
 
 
-def _check_no_partitions(value):
+def _check_no_partitions(value: str) -> int:
     """Check that number of partitions is between [2-10,000]
     Using function to validate rather than choices as error message prints all choices."""
     if not (re.match(r"\d*$", value) and int(value) < 10001 and int(value) > 1):
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -989,7 +989,7 @@ def _add_common_partition_arguments(optional_arguments, required_arguments):
         "--partition-num",
         "-pn",
         required=True,
-        help="Number of partitions/config files to generate, a number from 1 to 10,000",
+        help="Number of partitions/config files to generate, a number from 2 to 10,000",
         type=_check_no_partitions,
         metavar="[2-10000]",
     )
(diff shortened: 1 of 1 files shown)


data_validation/cli_tools.py
```diff
@@ -978,10 +989,9 @@ def _add_common_partition_arguments(optional_arguments, required_arguments):
         "--partition-num",
         "-pn",
         required=True,
-        help="Number of partitions/config files to generate",
-        type=int,
-        choices=range(1, 1001),
-        metavar="[1-1000]",
+        help="Number of partitions/config files to generate, a number from 1 to 10,000",
+        type=_check_no_partitions,
```
nehanene15, line 993: Instead of having a typecheck function, you can use the metavar parameter of `add_argument()` to change the description: https://stackoverflow.com/questions/25295487/python-argparse-value-range-help-message-appearance

sundar-mudupalli-work: Yes - however when you read the comment on that - metavar only helps with usage and help, but not with error - which is the problem I noticed - hence the link I posted in my first message.

data_validation/cli_tools.py
```diff
@@ -978,10 +989,9 @@ def _add_common_partition_arguments(optional_arguments, required_arguments):
         "--partition-num",
         "-pn",
         required=True,
-        help="Number of partitions/config files to generate",
-        type=int,
-        choices=range(1, 1001),
-        metavar="[1-1000]",
+        help="Number of partitions/config files to generate, a number from 1 to 10,000",
```
nehanene15:
```suggestion
        help="Number of partitions/config files to generate, a number from 2 to 10,000",
```

data_validation/cli_tools.py
```diff
@@ -954,6 +955,16 @@ def _add_common_arguments(optional_arguments, required_arguments):
     )
 
 
+def _check_no_partitions(value):
+    """Check that number of partitions less than limit
```
nehanene15:
```suggestion
    """Check that number of partitions is between [2-10,000]
```

data_validation/cli_tools.py
```diff
@@ -954,6 +955,16 @@ def _add_common_arguments(optional_arguments, required_arguments):
     )
 
 
+def _check_no_partitions(value):
```
nehanene15: Is this value a string? If so, lets annotate it
```suggestion
def _check_no_partitions(value: str) -> int:
```

data_validation/cli_tools.py
```diff
@@ -954,6 +955,16 @@ def _add_common_arguments(optional_arguments, required_arguments):
...
+    """Check that number of partitions less than limit
+    Using function to validate rather than choices as error message prints all choices."""
+    if not (re.match(r"\d*$", value) and int(value) < 10001 and int(value) > 1):
```
nehanene15: This may be more readable:
`if value.isdigit() and 2 <= int(value) <= 10000:`

nehanene15 (approved):

PR #1150, opened 2024-05-28
chore: Upgrade from setup.py to pyproject.toml

### LINKED ISSUE #906: Upgrade to pyproject.toml to configure setuptools
The setup.py based install is being deprecated, and now pyproject.toml is the standard way of specifiying project metadata.
This issue will involve migrating to pyproject.toml file.

### LINKED ISSUE #1147: DVT and Airflow Providers Google Library Conflicts
Hello, 

We were looking at executing the Data Validation Tool from within the Composer as per instructions in https://github.com/GoogleCloudPlatform/professional-services-data-validator/tree/develop/samples/airflow. We are getting versions conflict  because DVT needs specific versions of Google Cloud Libraries which conflict with the versions needed by "apache-airflow-providers-google" (Links below). 

https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/210c352ae8a601833a6e34b13972145abc1d49a6/setup.py#L55 

https://github.com/apache/airflow/blob/482c0be434cf4cd6d4769634f859cb5e5a243e9d/airflow/providers/google/provider.yaml#L141

Could you please guide us on finding compatible versions for DVT and Airflow Providers Google that can work on a Cloud Composer version? Or if we are missing something.

Thanks,
Moiz

helensilva14:
Closes #1147, #906 

- Adopts new project configuration file
- Upgrades version of some dependencies


ci/release_latest.sh
```diff
@@ -16,9 +16,9 @@ python3 -m venv rel_venv
 source rel_venv/bin/activate
 
 pip install setuptools wheel
-python setup.py register sdist bdist_wheel
```
nehanene15, line 19: Does this need to be replaced with pyproject.toml?

noxfile.py
```diff
@@ -38,7 +38,6 @@
     "tests",
     "third_party",
     "noxfile.py",
-    "setup.py",
```
nehanene15, line 41: Should this be replaced with pyproject.toml too? That way it can also be linted with black

ci/release_latest.sh
helensilva14: I don't think so because TOML is a file format itself and not a Python script, AFAIK 🤔

noxfile.py
helensilva14: Same thing as I answered for the other question, but maybe Black can lint this format, I'll search bout it

pyproject.toml
```diff
@@ -0,0 +1,66 @@
...
+
+[project.optional-dependencies]
+apache-airflow = ["apache-airflow==1.10.11"]
+pyspark = ["pyspark==3.0.0"]
+
+[project.scripts]
+data-validation = "data_validation.__main__:main"
+
+[tool.setuptools]
+packages = ["data_validation", "data_validation.query_builder", "data_validation.result_handlers"]
```
nehanene15: Should we ad the third_party/ibis packages here as well like we did in setup.py?
https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/setup.py#L68

nehanene15 (approved):

PR #1152, opened 2024-05-30
feat: Add a workaround for a Snowflake IN list limitation

### LINKED ISSUE #1146: Error on Snowflake binary primary key random row validations
When validating tables using random rows we build a list of primary key literal values and then inject that list into the validation queries as an IN list, for example:

```
FROM `db-01.dvt_test.dvt_num_pk` t6
WHERE t6.`ID` IN (1,20,100,500,123456)
```

For tables with binary primary keys we cannot inject binary literals into a SQL statement and therefore convert the binary values to hex first and cast them back to binary in SQL, for example:

```
FROM `db-01.dvt_test.dvt_bin_pk` t7
WHERE t7.`BINARY_ID` IN (FROM_HEX('4456542d6b65792d34'), FROM_HEX('4456542d6b65792d313132'), 
  FROM_HEX('4456542d6b65792d3236'), FROM_HEX('4456542d6b65792d3130'), FROM_HEX('4456542d6b65792d3531'))
```

We have an issue on Snowflake for binary primary keys when the random row sample size is > 50:

```
SQL compilation error: In-list contains more than 50 non-constant values
```

We need a workaround for Snowflake engines.

cc: @henrihem

nj1973:
Add a workaround for a Snowflake IN list limitation including support for splitting IN lists into a series of smaller IN list expressions.

### DIFF
--- a/data_validation/clients.py
+++ b/data_validation/clients.py
@@ -235,10 +235,10 @@ def get_data_client(connection_config):
             consts.GOOGLE_SERVICE_ACCOUNT_KEY_PATH
         )
         if key_path:
-            decrypted_connection_config[
-                "credentials"
-            ] = google.oauth2.service_account.Credentials.from_service_account_file(
-                key_path
+            decrypted_connection_config["credentials"] = (
+                google.oauth2.service_account.Credentials.from_service_account_file(
+                    key_path
+                )
             )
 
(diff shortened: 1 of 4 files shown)


nehanene15 (approved):
Nice!! Innovative solution - LGTM!

PR #1158, opened 2024-06-03
fix: Fixes bug in get_max_in_list_size

### LINKED ISSUE #1157: Random row validations fail on Oracle when sampling 1001+ ids
When validating with `--use-random-row --random-row-batch-size=1001`

The validation fails with:

```
sqlalchemy.exc.DatabaseError: (cx_Oracle.DatabaseError) ORA-01795: maximum number of expressions in a list is 1000
```

For this specific error we can build upon the recent enhancement from https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1146 and add a 1000 IN list limit for Oracle connections.

But we should also think about why this option was added and what sensible limits we should build in.

nj1973:
This PR fixes a bug in the code introduced in https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1146 and also caters for Oracle IN list limit of 1000.

Closes #1157

### DIFF
--- a/data_validation/clients.py
+++ b/data_validation/clients.py
@@ -274,6 +274,23 @@ def get_max_column_length(client):
     return 128
 
 
+def get_max_in_list_size(client, in_list_over_expressions=False):
+    if client.name == "snowflake":
+        if in_list_over_expressions:
+            # This is a workaround for Snowflake limitation:
+            #   SQL compilation error: In-list contains more than 50 non-constant values
+            # getattr(..., "cast") expression above is looking for lists where the contents are casts and not simple literals.
+            return 50
+        else:
+            return 16000
+    elif is_oracle_client(client):
+        # This is a workaround for Oracle limitation:
+        #   ORA-01795: maximum number of expressions in a list is 1000
+        return 1000
+    else:
+        return None
+
+
 CLIENT_LOOKUP = {
     "BigQuery": get_bigquery_client,
     "Impala": impala_connect,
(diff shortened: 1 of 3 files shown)


data_validation/clients.py
```diff
@@ -274,6 +274,23 @@ def get_max_column_length(client):
...
 
+def get_max_in_list_size(client, in_list_over_expressions=False):
+    if client.name == "snowflake":
+        if in_list_over_expressions:
+            # This is a workaround for Snowflake limitation:
+            #   SQL compilation error: In-list contains more than 50 non-constant values
+            # getattr(..., "cast") expression above is looking for lists where the contents are casts and not simple literals.
+            return 50
+        else:
+            return 16000
```
helensilva14, line 285: Curious to know: how did you find this other number for the list size? It would be helpful to put some sort of product documentation as a comment for future reference, but I could only find community posts mentioning it like this one: https://community.snowflake.com/s/question/0D5Do00000O3sJ0KAJ/how-o-get-around-th-16k-limit-on-number-of-expressions-in-in-clause-in-sql-generated-by-a-businessobjects-report

helensilva14 (changes requested):
Great job again Neil!

data_validation/clients.py
nj1973: It was not the same post but I saw another 2 Snowflake community posts.

I don't think we need to document it because this code is ensuring that we don't hit the 16k limit, at 16,000 we'll switch to:

```
WHERE id IN (1, ..., 16000) OR id IN (16001, ..., n)
```

We will find other limits for other engines I'm sure. I started to read about SQL Server but that has a memory limit based on the execution plan and not a hard/fixed limit, so I decided to just include the limits I knew about for now.

helensilva14: Right, it makes sense, it's clearer for me now, thank you! 

And for sure, totally agree that we might face it for other engines but at the moment we can stick with the ones we're aware of the values and usage

helensilva14 (approved):

PR #1164, opened 2024-06-10
fix: Removing t0 alias from column name, while getting schema from query. Adding Integration test for Hive Custom-Query

### LINKED ISSUE #1162: custom-query validations broken on Hive backend
When running custom-query column or row validations, the query generated is invalid. It appends a "t0" prefix to the column names which creates an invalid query. 

Command to reproduce: 
```
data-validation -v validate custom-query column -sc hive -tc hive -sq "select id, name as count_name from default.mascot" -tq "select id, name as count_name from default.mascot"
```

Query generated:
```
SELECT count(1) AS `count`
FROM (
  SELECT t1.`t0.id`, t1.`t0.name`
  FROM (
    select id, name from default.mascot
  ) t1
) t0
```

Error: `impala.error.HiveServer2Error: Error while compiling statement: FAILED: SemanticException [Error 10002]: Line 3:12 Invalid column reference 't0.id'`

We should add a custom query integration test for Hive as part of this PR.

piyushsarraf:
This PR does following: 

- Adding integration test for Hive Custom-Query validation. 
- Fixes the t0 alias issue, while generating schema from custom query, issue pointed in #1162 

Closes #1162

### DIFF
--- a/tests/system/data_sources/test_hive.py
+++ b/tests/system/data_sources/test_hive.py
@@ -339,3 +339,28 @@ def test_row_validation_binary_pk_to_bigquery():
     )
     df = run_test_from_cli_args(args)
     binary_key_assertions(df)
+
+
+@mock.patch(
+    "data_validation.state_manager.StateManager.get_connection_config",
+    new=mock_get_connection_config,
+)
+def test_custom_query_validation_core_types():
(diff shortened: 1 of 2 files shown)


nehanene15 (approved):
LGTM!

PR #1166, opened 2024-06-10
feat: Support `--trim-string-pks` flag for padded string semantics

### LINKED ISSUE #1151: Bug: String primary keys between Teradata and BigQuery are not matching
When a character primary key is used in a Teradata vs BigQuery row validation, the keys do not match. The output table shows trailing whitespaces in the keys from Teradata even though those whitespaces are not present in the Teradata table.

The same error persists when using composite primary keys that contain at least one CHAR typed column.

nehanene15:
Closes #1151 

- Supports flag `--trim-string-pks` intended for use when one engine uses padded string semantics (e.g. CHAR(n)) and the other does not (e.g. VARCHAR(n)).
- Supports random row with trim added
- When flag is provided, an RTRIM() is added on the PKs.


tests/system/data_sources/test_mysql.py
```diff
@@ -356,6 +357,35 @@ def test_row_validation_binary_pk_to_bigquery():
...
+@mock.patch(
+    "data_validation.state_manager.StateManager.get_connection_config",
+    new=mock_get_connection_config,
+)
+def test_row_validation_char_pk_to_bigquery():
+    """MySQL to BigQuery dvt_char_id row validation.
+    This is testing CHAR primary key join columns.
+    Includes random row filter test.
+
+    Note that this test will fail unless PAD_CHAR_TO_FULL_LENGTH SQL mode is enabled.
```
nehanene15, line 370: This test isn't working since we don't enable PAD_CHAR_TO_FULL_LENGTH SQL mode. @nj1973 we may have to skip this test for now...

tests/system/data_sources/test_teradata.py
```diff
@@ -477,6 +478,62 @@ def test_row_validation_binary_pk_to_bigquery():
...
+        [
+            "validate",
+            "row",
+            "-sc=td-conn",
+            "-tc=bq-conn",
+            "-tbls=udf.dvt_char_id=pso_data_validator.dvt_char_id",
+            "--primary-keys=id",
+            "--hash=id,other_data",
+            "--use-random-row",
+            "--random-row-batch-size=5",
```
nehanene15: I did add support for trimming the PKs within random row when the `-tsp` flag is present. This should only be used when the destination is VARCHAR and source is CHAR. 

Random rows are applied at the inner-most level so adding doing random row validation with BQ padded columns is not possible. For this reason, I think we should remove the random row from this test.

tests/system/data_sources/test_mysql.py
nj1973: Agreed, thanks

tests/system/data_sources/test_teradata.py
nj1973: Agreed, thanks

nj1973 (approved):
LGTM

PR #1173, opened 2024-06-19
test: Add unicode character tests

nj1973:
Add row integrations tests for a wider set of characters than standard test data, e.g. data with accents.

For most SQL engines the test is a hash comparison.

For Teradata the hashing UDF does not accept unicode values and therefore that test is based on concat output.

I've skipped the MySQL test because unicode characters are being dropped:
```
FR LYXFJDER vs FÖR LYXFJÄDER
```


nehanene15 (approved):
LGTM!

PR #1176, opened 2024-06-21
docs: Tweaks to Cloud Functions sample

### LINKED ISSUE #1175: Sample Cloud Functions tweaks
While testing this sample I made a few tweaks.

nj1973:
While testing this sample I needed to increase memory limit to be able to run the function. So I made a couple of other tweaks at the same time.

I did contemplate switching the example to gen2 which only requires addition of `--gen2` flag and change of format for `--memory` value. But wasn't sure if we wanted to do this or not.


nehanene15 (approved):
LGTM, this is great! I think updating to gen2 would be a good idea to be prescriptive.

PR #1178, opened 2024-06-24
feat:  generate-table-partitions write multiple partitions to a single yaml file

sundar-mudupalli-work:
Added a new parameter to generate-table-partitions `--parts-per-file` which validates multiple partitions in one yaml file. Containers have some startup time - so having multiple partitions (validations) in one yaml file spreads the cost across multiple partitions. Each yaml file can still be run concurrently with other yaml files.


README.md
```diff
@@ -241,9 +241,11 @@ data-validation (--verbose or -v) (--log-level or -ll) generate-table-partitions
                         Directory Path to store YAML Config Files
                         GCS: Provide a full gs:// path of the target directory. Eg: `gs://<BUCKET>/partitions_dir`
                         Local: Provide a relative path of the target directory. Eg: `partitions_dir`
-  --partition-num [1-1000], -pn [1-1000]
-                        Number of partitions/config files to generate
+  --partition-num INT, -pn INT
```
nehanene15, line 248: It's nice to keep the range here imo, we can do `[2-10000]` with the closed brackets indicating the starting and ending points are included.

README.md
```diff
@@ -241,9 +241,11 @@ data-validation (--verbose or -v) (--log-level or -ll) generate-table-partitions
                         Directory Path to store YAML Config Files
                         GCS: Provide a full gs:// path of the target directory. Eg: `gs://<BUCKET>/partitions_dir`
                         Local: Provide a relative path of the target directory. Eg: `partitions_dir`
-  --partition-num [1-1000], -pn [1-1000]
-                        Number of partitions/config files to generate
+  --partition-num INT, -pn INT 
+                        Number of partitions into which the table should be split, e.g. 1000 or 10000
                         In case this value exceeds the row count of the source/target table, it will be decreased to max(source_row_count, target_row_count)
+  [--parts-per-file INT], [-parts-per-file INT]
```
nehanene15: Even here, what is the min and max range we support?

README.md
```diff
@@ -241,9 +241,11 @@ data-validation (--verbose or -v) (--log-level or -ll) generate-table-partitions
                         Directory Path to store YAML Config Files
                         GCS: Provide a full gs:// path of the target directory. Eg: `gs://<BUCKET>/partitions_dir`
                         Local: Provide a relative path of the target directory. Eg: `partitions_dir`
-  --partition-num [1-1000], -pn [1-1000]
-                        Number of partitions/config files to generate
+  --partition-num INT, -pn INT 
+                        Number of partitions into which the table should be split, e.g. 1000 or 10000
                         In case this value exceeds the row count of the source/target table, it will be decreased to max(source_row_count, target_row_count)
+  [--parts-per-file INT], [-parts-per-file INT] 
+                        Number of partitions in a yaml file. A yaml file can contain one (default) or more partitions. Having multiple partitions in one yaml file speads the cost of container start up accross multiple sequential validations. Different yaml files can be run in multiple containers in parallel.
```
nehanene15: I think this description should be a bit more succinct, maybe just "Number of partitions per yaml file." - We can add more details in the section above this CLI reference on line 219.

data_validation/cli_tools.py
```diff
@@ -954,14 +955,21 @@ def _add_common_arguments(optional_arguments, required_arguments):
     )
 
 
-def _check_no_partitions(value: str) -> int:
-    """Check that number of partitions is between [2-10,000]
-    Using function to validate rather than choices as error message prints all choices."""
-    if value.isdigit() and 2 <= int(value) <= 10000:
-        return int(value)
+def _check_positive(value):
```
nehanene15: Good practice to add parameter type annotations i.e `_check_positive(value: str) -> int`

data_validation/cli_tools.py
```diff
@@ -954,14 +955,21 @@ def _add_common_arguments(optional_arguments, required_arguments):
...
+    ivalue = int(value)
+    if ivalue <= 0:
+        raise argparse.ArgumentTypeError("%s is an invalid positive int value" % value)
+    return ivalue
+
+
+def check_no_yaml_files(partition_num, parts_per_file):
```
nehanene15: Add parameter types + return value annotations here too

data_validation/cli_tools.py
```diff
@@ -1008,6 +1015,13 @@ def _add_common_partition_arguments(optional_arguments, required_arguments):
         "-sa",
         help="Path to SA key file for result handler output",
     )
+    optional_arguments.add_argument(
+        "--parts-per-file",
+        "-parts-per-file",
```
nehanene15: Typically the shortcut flag should be something abbreviated unless the flag itself is <5 characters i.e. `-ppf`

data_validation/partition_builder.py
```diff
@@ -40,17 +41,38 @@ def _get_arg_config_dir(self) -> str:
...
 
-    def _get_yaml_from_config(self, config_manager: ConfigManager) -> Dict:
-        """Return dict objects formatted for yaml validations.
+    def _add_filters_get_yaml(
+        self,
+        config_manager: ConfigManager,
+        source_filters: List[str],
+        target_filters: List[str],
+    ) -> Dict:
+        """Return dict objects formatted for yaml validations. Qualify the configurations by adding the source and
```
nehanene15: Are these the filters that are provided with `data-validation generate-table-partitions -filters 'X'`? Or are these the new generated filters for the table partitions?

data_validation/partition_builder.py
```diff
@@ -40,17 +41,38 @@ def _get_arg_config_dir(self) -> str:
...
+            target filters which validate individual partitions. Each filter pair will result in a DVT execution. Running multiple validations
+            sequentially spreads the startup and shutdown overhead of containers accross multiple partitions.
+            Parallelism can be achieved by running multiple yaml files in parallel.
```
nehanene15: Let's keep the docstring for each function limited to the input/output + goal of the function - otherwise it gets hard to understand. We should avoid documenting best practices/explaining DVT usage here. End users won't be reading this, it should be for developers only.

We need to update the 'Args' in this docstring with source/target_filters. There we can explain which type of filter this is adding to the config (existing or new).

data_validation/partition_builder.py
```diff
@@ -320,27 +341,24 @@ def _add_partition_filters(
...
-                filter_dict = {
-                    "type": "custom",
-                    "source": filter_list[0][pos],
-                    "target": filter_list[1][pos],
-                }
-                # Append partition new filter
-                config_manager.filters.append(filter_dict)
 
+            # Create a list of lists chunked by partitions per file
+            source_filters_list = list_to_sublists(
```
nehanene15, line 367: Do we need to check that len(source_filters_list) == len(target_filters_list)? Or can they have different lengths

data_validation/partition_builder.py
```diff
@@ -311,7 +333,6 @@ def _add_partition_filters(
         Returns:
             yaml_configs_list (List[Dict]): List of YAML configs for all tables
         """
-
```
nehanene15, line 326: I think we need more clarity on what these input data structures are. Not sure I understand what this means...

List for each ConfigManager, containing a List of partition filters which each has a List for source and target filters?
[ConfigManager1 [ PartitionFilter1 [source: filter], [target:filter], PartitionFilter2 [source: filter], [target:filter] ] ]

Would it be cleaner to create a class called PartitionFilter that stores the source and target filter? So we can easily access them (PartitionFilter.source_filter) without nested lists?

As for the return value - how is the Dict structured?
List of Dict {"yaml_files": {{"target_file_name": "file.yaml", "yaml_config": yaml_config}}

data_validation/partition_builder.py
```diff
@@ -320,27 +341,24 @@ def _add_partition_filters(
...
-                config_manager.filters.append(filter_dict)
 
+            # Create a list of lists chunked by partitions per file
+            source_filters_list = list_to_sublists(
+                filter_list[0], self.args.parts_per_file
+            )
+            target_filters_list = list_to_sublists(
+                filter_list[1], self.args.parts_per_file
+            )
+            for i, source_filters in enumerate(source_filters_list):
```
nehanene15: Why did we enumerate only source_filters_list? It's confusing because we use both source and target filters here.
It may be more simple to do:
```
for i in len(source_filters_list):
                yaml_config = self._add_filters_get_yaml(
                    config_manager, source_filters_list[i], target_filters_list[i]
                )
```

nehanene15 (commented):
Need some more clarification on what data structures you are using/updating.

README.md
```diff
@@ -217,7 +217,7 @@ data-validation (--verbose or -v) (--log-level or -ll) validate row
...
 
-The command generates and stores multiple YAML configs that represent chunks of the large table using filters (`WHERE primary_key(s) >= X AND primary_key(s) < Y`). You can then run the configs in the directory serially (or in parallel in multiple containers, VMs) with the `data-validation configs run --config-dir PATH` command as described [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator#yaml-configuration-files).
+The command generates and stores multiple YAML validations each representing a chunk of the large table using filters (`WHERE primary_key(s) >= X AND primary_key(s) < Y`) in one YAML file. The parameter parts -per-file, specifies the number of validations in one YAML file. The partitions are equally divided across many yaml files in one directory. You can then run the validations in the directory serially (or in parallel in multiple containers, VMs) with the `data-validation configs run --config-dir PATH` command as described [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator#yaml-configuration-files).
```
nehanene15: typos, should be "The `parts-per-file` flag specifies the number of validations in one YAML file."

What does this mean? "The partitions are equally divided across many yaml files in one directory."

data_validation/partition_builder.py
```diff
@@ -40,17 +41,41 @@ def _get_arg_config_dir(self) -> str:
...
 
-    def _get_yaml_from_config(self, config_manager: ConfigManager) -> Dict:
-        """Return dict objects formatted for yaml validations.
+    def _add_filters_get_yaml_file(
+        self,
+        config_manager: ConfigManager,
+        source_filters: List[str],
+        target_filters: List[str],
+    ) -> Dict:
+        """Given a ConfigManager object (from the. input args to generate-partitions), add the filters,
```
nehanene15: nit: typo "from the. input"

So this is generating multiple validation blocks per filter and returning one YAML dict correct? Can we document the fact we are creating multiple validations per filter because it's unclear why we input a list of filters rather than a single filter

data_validation/partition_builder.py
```diff
@@ -88,15 +113,14 @@ def _extract_where(table_expr, client) -> str:
         ).replace("t0.", "")
 
     def _get_partition_key_filters(self) -> List[List[List[str]]]:
-        """Generate where clauses for each partition for each table pair. We have to create separate where for the source and
-            target as they may each have a different filter applied to them. We are partitioning the tables based on keys, so that
-            we get equivalent sized partitions that can be compared against each other. With this approach, we can validate the
-            partitions in parallel leading to horizontal scaling of DVT. The design doc for this section is available in
+        """Given self.config_managers - i.e. a list of table pairs for row validation, generate the paritition filter
```
nehanene15: nit: typo - misspelled 'partition filter'

Since you aren't inputting a ConfigManager object, it's confusing to say 'given self.config_managers'.

data_validation/partition_builder.py
```diff
@@ -316,43 +340,39 @@ def _add_partition_filters(
         ConfigManager objects.
 
         Args:
-            partition_filters (List[List[List[str]]]): List of List of Partition filters
-            for all Table/ConfigManager objects, two list of filters for each configManager object,
-            one for the source and one for the target
-
+            self.config_managers is a list of table pairs to which the partition filters must be applied.
```
nehanene15: config_managers isn't a list of table pairs.. are your referring to the fact that each ConfigManager represents a single table validation?

tests/system/data_sources/test_bigquery.py
```diff
@@ -1147,18 +1163,28 @@ def test_bigquery_generate_table_partitions(mock_conn):
...
             "-hash=*",
             "-cdir=/home/users/yaml",
-            "-pn=4",
+            "-pn=9",
+            "-ppf=5",
+            "-filters=quarter_id != 1111",
         ]
     )
     config_managers = main.build_config_managers_from_args(args, consts.ROW_VALIDATION)
     partition_builder = PartitionBuilder(config_managers, args)
```
nehanene15: This test and it's assertions are repeated for all DB types. Can we move the assertions to [common_functions](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/tests/system/data_sources/common_functions.py)?

sundar-mudupalli-work: Yes, I realize that - the only thing that is likely different is the EXPECTED_PARTITION_FILTER - which can be an argument. I was thinking this sort of cleanup is something we should do for the whole test suite.

nehanene15 (approved):
LGTM!

PR #1180, opened 2024-06-27
fix: Fixing multiple function call, to get schema in custom-query validation for Hive

piyushsarraf:
This PR fixes following: 

- Multiple function call to _get_schema_using_query function for custom-query validation. 
- Use custom-query as is with an addition of LIMIT function, instead of having custom-query as a subquery, earlier it was little cost-intensive to customer, if custom-query returns huge amount of rows. 
- Changing fetchall() to fetchone(), just to improve the performance, as we only require the column headers.


third_party/ibis/ibis_impala/api.py
```diff
@@ -193,12 +194,16 @@ def _if_null(op):
     return ops.Coalesce((op.arg, op.ifnull_expr))
 
 
+@lru_cache(maxsize=2)
 def _get_schema_using_query(self, query):
     # Removing LIMIT 0 around query since it returns no results in Hive
-    cur = self.raw_sql(f"SELECT * FROM ({query}) t0 LIMIT 1")
-    cur.fetchall()
+    cur = self.raw_sql(query + " LIMIT 1")
```
nehanene15: Will this cause an issue if the custom query provided already has a LIMIT function within it? We might need to add a check - if it does already have a LIMIT, we can replace it with 'LIMIT 1'.

nehanene15 (commented):
Great work! Suggested one small change.

nehanene15 (approved):
LGTM! You may have already done this but before merging I would test custom query validation between Hive and BQ with + without the LIMIT since our tests only check Hive to Hive.

helensilva14:
Hi all! Just to confirm: there's really no opened issues to be fixed by this PR?

third_party/ibis/ibis_impala/api.py
sundar-mudupalli-work: It appears to me that the existing code `f"SELECT * FROM ({query}) t0 LIMIT 1"` does the same thing as the proposed change and will work correctly if the query has a limit clause. The proposed change will likely throw a syntax error if the custom query had a limit clause.

piyushsarraf: @sundar-mudupalli-work, yes the earlier proposed solution will throw a syntax error if custom-query has a LIMIT keyword, so the `update_query_with_limit` does a regex check that if the query has a LIMIT function, and it changes it to LIMIT 1, else we add LIMIT 1 to the query, to fetch the schema.

piyushsarraf:
> Hi all! Just to confirm: there's really no opened issues to be fixed by this PR?

@helensilva14, changes are made to solve another branch of [this](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1162) issue. 

- To decrease multiple function call to `_get_schema_using_query`
- Avoiding nested query to get schema of a custom-query, which can improve cost implied to user, if the custom-query table is large.

PR #1184, opened 2024-07-03
fix!: Failing validations of strings with extended ascii characters

### LINKED ISSUE #1170: Accented string causes validation_status=fail in TD VS BQ row validation
When running a row validation, the hash for a string in BigQuery is different to the hash of the same string in Teradata if the string has an accented vowel. (I suspect the issue might also occur for any non ascii char).

This causes a failed row validation for strings that are identical.

sundar-mudupalli-work:
DVT while performing row validations converts all columns to upper case before computing the hash function. This is generally OK, except when the upper case function works differently in different databases. For example in Teradata applying upper case to an extended ascii character (e.g. á, â or é) has no effect. Whereas in BQ applying upper case transforms them to Á, Â and É. The hash comparisons then fail even when the columns hold the same values.

BREAKING CHANGE: DVT will default to case-sensitive comparisons when previously they were case-insensitive. To continue running case-insensitive comparisons use the `--case-insensitive-match` flag when running concat or hash validations. This will not affect comparison field validations.

Closes #1170


data_validation/config_manager.py
```diff
@@ -975,7 +975,7 @@ def build_dependent_aliases(self, calc_type, col_list=None):
                 "cast",
                 "ifnull",
                 "rstrip",
-                "upper",
+                # "upper",
```
nj1973: Personally I would remove the line rather than comment it out.

nj1973 (approved):
LGTM

sundar-mudupalli-work:
Team,

Here is how Teradata compares character columns from their documentation. This maybe why we initially did `rtrim` and `upper`.  [Source is here](https://docs.teradata.com/r/Enterprise_IntelliFlex_VMware/SQL-Functions-Expressions-and-Predicates/Comparison-Operators-and-Functions/Character-String-Comparisons/Case-Sensitivity)

For case blind comparisons, any lowercase single byte Latin letters are converted to uppercase before comparison begins. The prepared strings are compared and any trailing pad characters are ignored.

A case blind comparison always considers lowercase and uppercase Cyrillic, Greek and full-width ASCII letters to be equivalent. To distinguish lowercase and uppercase Cyrillic, Greek, and fullwidth ASCII letters you must explicitly declare CASESPECIFIC comparison.

data_validation/cli_tools.py
sundar-mudupalli-work, line 1: Good work removing repeating code.

sundar-mudupalli-work (commented):
LGTM

PR #1186, opened 2024-07-08
Fix: Issue with GoogleSQL regarding dates before 1000CE

### LINKED ISSUE #1181: Date fields starting with 0 (e.g. 1/16/0201) cause failed validations

sundar-mudupalli-work:
GoogleSQL when presented with dates before 1000CE does not print a four digit date. This fix changes the format string used to force a 4 digit date for BigQuery and Spanner backends.

Closes #1181

### DIFF
--- a/third_party/ibis/ibis_addon/operations.py
+++ b/third_party/ibis/ibis_addon/operations.py
@@ -187,9 +187,9 @@ def strftime_bigquery(translator, op):
             arg_formatted,
             arg_type.timezone if arg_type.timezone is not None else "UTC",
         )
-    # Deal with issue 1181 which appears to be due a GoogleSQL bug with dates before 1000 CE 
-    if format_str.value == '%Y-%m-%d' :
-        fmt_string="'%4Y-%m-%d'"
+    # Deal with issue 1181 which appears to be due a GoogleSQL bug with dates before 1000 CE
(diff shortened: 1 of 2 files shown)


third_party/ibis/ibis_addon/operations.py
```diff
@@ -178,6 +178,9 @@ def strftime_bigquery(translator, op):
     arg_type = arg.output_dtype
     strftime_format_func_name = BQ_STRFTIME_FORMAT_FUNCTIONS[type(arg_type)]
     fmt_string = translator.translate(format_str)
+    # Deal with issue 1181 due a GoogleSQL bug with dates before 1000 CE affects both date and timestamp types
+    if format_str.value.startswith("%Y"):
+        fmt_string = fmt_string.replace("%Y", "%4Y", 1)
```
nj1973: BigQuery documentation states we should use "%E4Y":

https://cloud.google.com/bigquery/docs/reference/standard-sql/format-elements#format_elements_date_time

I see "%4Y" also works but don't see it documented, should we use the documented format? Or did I miss some other documentation?

nj1973 (commented):
Just a minor comment for review

nj1973 (approved):
LGTM

PR #1197, opened 2024-07-17
fix: Close source and target connections after executing a validation

### LINKED ISSUE #1195: Multi table validations exhausted PostgreSQL connections
When running a single `validate` command for many tables it was noticed that many PostgreSQL connections are opened, the command failed with the error below:

```
File "/usr/lib/python3.10/site-packages/sqlalchemy/engine/default.py", line 598, in connect
    return self.dbapi.connect(*cargs, **cparams)
  File "/usr/lib/python3.10/site-packages/psycopg2/__init__.py", line 122, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
psycopg2.OperationalError: connection to server at "127.0.0.1", port 5432 failed: FATAL:  remaining connection slots are reserved for non-replication superuser connections
```

I've tested this on PostgreSQL and Oracle and can see the connection count climbing until the command eventually completes and all connections are released.

nj1973:
When multiple tables are validated in a single command we are building up a number of idle connections, at least in Oracle and PostgreSQL we are. Example command where ${TABLES} contains `find-tables` output of 23 tables:
```
data-validation  validate column -sc=ora_local -tc=pg_local -tbls="${TABLES}"
```

When I checked Oracle & PostgreSQL during the command we accumulated > 12 idle connections.

I changed the `DataValidation` class to support being called from a context manager and explicitly closed connections on exit. After this my test command never had more than two connections open at any time.

I also added some code to Oracle client to tag database sessions with "DVT" string.

### DIFF
--- a/data_validation/__main__.py
+++ b/data_validation/__main__.py
@@ -531,49 +531,49 @@ def run_validation(config_manager, dry_run=False, verbose=False):
         dry_run (bool): Print source and target SQL to stdout in lieu of validation.
         verbose (bool): Validation setting to log queries run.
     """
-    validator = DataValidation(
+    with DataValidation(
         config_manager.config,
         validation_builder=None,
         result_handler=None,
         verbose=verbose,
-    )
+    ) as validator:
 
-    if dry_run:
-        sql_alchemy_clients = [
-            "mysql",
-            "oracle",
-            "postgres",
-            "db2",
-            "mssql",
-            "redshift",
-            "snowflake",
-        ]
+        if dry_run:
+            sql_alchemy_clients = [
+                "mysql",
+                "oracle",
+                "postgres",
+                "db2",
+                "mssql",
+                "redshift",
+                "snowflake",
+            ]
 
-        source_query = validator.validation_builder.get_source_query().compile()
-        if config_manager.source_client.name in sql_alchemy_clients:
-            source_query = source_query.compile(
-                config_manager.source_client.con.engine,
-                compile_kwargs={"literal_binds": True},
+            source_query = validator.validation_builder.get_source_query().compile()
+            if config_manager.source_client.name in sql_alchemy_clients:
+                source_query = source_query.compile(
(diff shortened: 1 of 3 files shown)


third_party/ibis/ibis_oracle/__init__.py
```diff
@@ -60,8 +60,15 @@ def do_connect(
...
         self.database_name = sa_url.database
         engine = sa.create_engine(
-            sa_url, poolclass=sa.pool.StaticPool, arraysize=self.arraysize
+            sa_url,
+            poolclass=sa.pool.StaticPool,
+            arraysize=self.arraysize,
         )
+        try:
+            # Identify the session in Oracle as DVT, no-op if this fails.
+            engine.raw_connection().connection.module = "DVT"
```
helensilva14, line 69: That's a very nice touch!

data_validation/__main__.py
```diff
@@ -531,49 +531,49 @@ def run_validation(config_manager, dry_run=False, verbose=False):
...
-            "snowflake",
-        ]
-
-        source_query = validator.validation_builder.get_source_query().compile()
-        if config_manager.source_client.name in sql_alchemy_clients:
-            source_query = source_query.compile(
-                config_manager.source_client.con.engine,
-                compile_kwargs={"literal_binds": True},
-            )
+    ) as validator:
```
helensilva14, line 539: I'm curious: was this change just for "styling" purposes or it was needed for the fix (that new __exit__ method) actually work?

helensilva14 (approved):
LGTM! Just added a few questions out of curiosity

data_validation/__main__.py
nj1973: It ensures the new exit method is called. An alternative would have been to create a public exit or close method and explicitly call that. I think this is nicer.

helensilva14: Indeed, I agree! Thanks for the explanation and great job 🚀

PR #1199, opened 2024-07-19
feat: Add TLS options to PostgreSQL connections add

### LINKED ISSUE #1072: SSL enabled database Connectivity Option:  For connecting Google CloudSQL(postgres) using SSL
As of now, there is no support for connecting to SSL enabled Google Cloudsql(Postgres)  database. Request to enhance / add SSLenabled cloudsql database connectivity by adding 2 more optional parameters "ssl-cert", "ssl-key" to connection command of DVT.

nj1973:
Adds several new options to Postgres `connections add`.

```
[--hostaddr HOSTADDR]        May be required to supply an address that differs from the name in the host option
[--sslcert SSLCERT]          Local path to your client certificate (client-cert.pem) file
[--sslkey SSLKEY]            Local path to your private client key (client-key.pem) file
[--sslrootcert] SSLROOTCERT  Local path to your root certificate (server-ca.pem) file
[--sslmode SSLMODE]          SSL mode
```

`--hostaddr` is one I didn't expect to need but I could only connect via TLS with `--sslmode=verify-full` if I included the real host name in `--host` (which for Cloud SQL was a complex string like "2-ab12ab12-ab12-12a1-1234-1abc12ab12ab.europe-west1.sql.goog") and supplied the IP address via `--hostaddr`. That is why I added that option too.

### DIFF
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -107,6 +107,23 @@ CONNECTION_SOURCE_FIELDS = {
         ["user", "Username to connect to"],
         ["password", "Password for authentication of user"],
         ["database", "Database in PostgreSQL to connect to (default postgres)"],
+        [
+            "hostaddr",
+            "(Optional) May be required to supply an address that differs from the name in the host option",
+        ],
+        [
+            "sslcert",
+            "(Optional) Local path to your client certificate (client-cert.pem) file",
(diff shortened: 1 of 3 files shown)


helensilva14 (approved):
LGTM!

PR #1203, opened 2024-07-23
fix: Support char comparison fields in Teradata

### LINKED ISSUE #1190: Trailing whitespaces in char fields when using --comparison-fields
When using `--comparison-fields ` on a field with datatype `CHAR(N)` the validation fails for all rows. This seems to be because DVT is adding trailing whitespaces to all values from source being compared.

nehanene15:
Closes #1190 

Due to a Teradata client "quirk" extra padding is added to CHAR data types. This adds an rtrim() when doing comparison field row validations. It affects all string comparison fields when Teradata is a source or target system.

### DIFF
--- a/data_validation/__main__.py
+++ b/data_validation/__main__.py
@@ -160,7 +160,7 @@ def get_aggregate_config(args, config_manager: ConfigManager):
     return aggregate_configs
 
 
-def get_calculated_config(args, config_manager):
+def get_calculated_config(args, config_manager) -> List[dict]:
     """Return list of formatted calculated objects.
 
     Args:
@@ -290,9 +290,26 @@ def build_config_from_args(args: Namespace, config_manager: ConfigManager):
             comparison_fields = cli_tools.get_arg_list(
                 args.comparison_fields, default_value=[]
             )
-            config_manager.append_comparison_fields(
-                config_manager.build_config_comparison_fields(comparison_fields)
-            )
+
+            if (
+                config_manager.source_client.name == "teradata"
+                or config_manager.target_client.name == "teradata"
+            ):
+
+                full_comp_fields = config_manager._add_rstrip_for_td_str_comp_fields(
+                    comparison_fields
+                )
+                print(full_comp_fields)
+                comp_config = config_manager.build_config_comparison_fields(
+                    full_comp_fields
+                )
+                print(comp_config)
+                config_manager.append_comparison_fields(comp_config)
+
+            else:
+                config_manager.append_comparison_fields(
+                    config_manager.build_config_comparison_fields(comparison_fields)
+                )
 
         # Append primary_keys
         primary_keys = cli_tools.get_arg_list(args.primary_keys)
--- a/data_validation/__main__.py
+++ b/data_validation/__main__.py
@@ -291,25 +291,19 @@ def build_config_from_args(args: Namespace, config_manager: ConfigManager):
                 args.comparison_fields, default_value=[]
             )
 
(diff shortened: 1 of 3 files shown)


data_validation/__main__.py
```diff
@@ -290,6 +290,17 @@ def build_config_from_args(args: Namespace, config_manager: ConfigManager):
...
                 args.comparison_fields, default_value=[]
             )
+
+            # As per #1190, add rstrip for Teradata string comparison fields
+            if (
+                config_manager.source_client.name == "teradata"
+                or config_manager.target_client.name == "teradata"
+            ):
+
+                comparison_fields = config_manager.add_rstrip_to_comp_fields(
```
nj1973, line 300: I'm probably missing some key point but here are some ignorant questions:

It feels like this detail should be hidden away inside `build_config_comparison_fields()` and not in `__main__`.

Is a valid alternative adding another field to the `column_config` dict like we already do for `consts.CONFIG_CAST`? And only set the new field when source or target is Teradata and type is string?

data_validation/config_manager.py
```diff
@@ -508,6 +508,55 @@ def build_config_manager(
             verbose=verbose,
         )
 
+    def add_rstrip_to_comp_fields(self, comparison_fields: List[str]) -> List[str]:
+        """As per #1190, add an rstrip calculated field for Teradata string comparison fields.
+
+        Parameters:
```
nj1973, line 514: For my education, do you use a VS Code extension to auto create docstrings or type them out yourself?

tests/system/data_sources/test_teradata.py
```diff
@@ -523,6 +523,30 @@ def test_row_validation_char_pk_to_bigquery():
     id_type_test_assertions(df)
 
 
+@mock.patch(
+    "data_validation.state_manager.StateManager.get_connection_config",
+    new=mock_get_connection_config,
+)
+def test_row_validation_char_comp_field_to_bigquery():
```
nj1973, line 530: I wonder if this should be a comparison fields test across all columns rather than restricted to `col_char_2`? That would test for side effects in addition to the "happy path".

I don't think we have any comparison field testing so this could start that process, we could add an issue asking to duplicate this test across other engines?

nj1973 (commented):
Left a few comments for review

data_validation/__main__.py
nehanene15: The tough part about this was that I wanted to add the rstrip() to validations where comparison fields where explicitly passed in. When running a hash or concat validation, the comparison field automatically populates as 'hash__all' or 'concat__all' respectively. If I updated the `column_config` dict, it will add an rtrim on top of the hash/concat values which we don't want. 

I do agree with moving the code into config_manager() though - will make that change!

data_validation/config_manager.py
nehanene15: Type them out myself :)

data_validation/__main__.py
nehanene15: As I was implementing this, I realized that this snippet does have to remain in main.py. If I move the 'add_rstrip_to_comp_fields' logic into the `build_config_comparison_fields()` method, it adds the rstrip to hash and concat values since that function is also called when building the comparison fields for hash and concat.

tests/system/data_sources/test_teradata.py
nehanene15: good idea - will add this --comp-fields test to all TD columns

nj1973 (approved):
LGTM

PR #1204, opened 2024-07-24
fix: Oracle cast(decimal to string) caters for non-zero values <1

### LINKED ISSUE #1201: Row hash validation on numeric columns less than 1 produces different results in Oracle vs Postgres
The discrepancy in row hash value occurs when the row in question contains any numerical columns with a decimal value of less between (-1, 1) (e.g., `0.71`, `-0.3` but NOT `1.5`)

Based on my investigation, when DVT calculate the row hash, it perform a set of operations on each columns. For numeric columns, it first CAST its value to VARCHAR before concatenating its value with the rest of the columns.

There is a discrepancy in behaviour in terms of how CASTing a numeric value into VARCHAR works in the two database engines. In Postgres, CAST(`0.71` as VARCHAR) would result in `"0.71"` while in Oracle, the equivalent expression would result in `".71"`
DVT then uses this resulting string to calculate the row hash. Since the strings are different, the hash is also different.

Example:

```
$ data-validation query \
--conn oracle-db \
--query "SELECT
                t0.id AS id
                ,t0.cage_equivalents AS cage_equivalents
                ,CAST(t0.cage_equivalents AS VARCHAR2(4000 CHAR)) AS cast__cage_equivalents
                ,coalesce(CAST(t0.cage_equivalents AS VARCHAR2(4000 CHAR)), 'DEFAULT_REPLACEMENT_STRING') AS ifnull__cast__cage_equivalents
                ,rtrim(coalesce(CAST(t0.cage_equivalents AS VARCHAR2(4000 CHAR)), 'DEFAULT_REPLACEMENT_STRING')) AS rstrip__ifnull__cast__cage_equivalents
                ,upper(rtrim(coalesce(CAST(t0.cage_equivalents AS VARCHAR2(4000 CHAR)), 'DEFAULT_REPLACEMENT_STRING'))) AS upper__rstrip__ifnull__cast__cage_equivalents
                ,lower(standard_hash(convert(upper(rtrim(coalesce(CAST(t0.cage_equivalents AS VARCHAR2(4000 CHAR)), 'DEFAULT_REPLACEMENT_STRING'))), 'UTF8'), 'SHA256')) AS hash__all
        FROM some_table t0
        WHERE id = 150389"
[(150389, Decimal('0.71'), '.71', '.71', '.71', '.71', '32c62276b4ad988e5e62d117b1fa809a03571e97368d3a4cb7268efde740be88')]
$ 



$ data-validation query \
--conn postgres-db \
--query "SELECT 
                t7.id AS id
                ,t7.cage_equivalents as cage_equivalents
                ,CAST(t7.cage_equivalents AS VARCHAR(4000)) AS cast__cage_equivalents
                ,coalesce(CAST(t7.cage_equivalents AS VARCHAR(4000)), 'DEFAULT_REPLACEMENT_STRING') AS ifnull__cast__cage_equivalents
                ,rtrim(coalesce(CAST(t7.cage_equivalents AS VARCHAR(4000)), 'DEFAULT_REPLACEMENT_STRING'), ' \t \r\u000b\f') AS rstrip__ifnull__cast__cage_equivalents
                ,upper(rtrim(coalesce(CAST(t7.cage_equivalents AS VARCHAR(4000)), 'DEFAULT_REPLACEMENT_STRING'), ' \t \r\u000b\f')) AS upper__rstrip__ifnull__cast__cage_equivalents
                ,concat_ws('', upper(rtrim(coalesce(CAST(t7.cage_equivalents AS VARCHAR(4000)), 'DEFAULT_REPLACEMENT_STRING'), ' \t \r\u000b\f'))) AS concat__all
                ,encode(sha256(convert_to(concat_ws('', upper(rtrim(coalesce(CAST(t7.cage_equivalents AS VARCHAR(4000)), 'DEFAULT_REPLACEMENT_STRING'), ' \t \r\u000b\f'))), 'UTF8')), 'hex') AS hash__all
        FROM some_table AS t7
        WHERE id = 150389"
[(Decimal('150389'), Decimal('0.71'), '0.71', '0.71', '0.71', '0.71', '0.71', 'c68fa0c5dd8c9a2cf6555708f938a5d592d20b0d81f0d40ac6221d07500ae145')]
$ 
```

nj1973:
This PR fixes two issues:

1) Oracle non-zero NUMBERs < 1 (or > -1) were missing a leading zero when cast to string.
2) PostgreSQL non-zero DECIMALs < 1 (or > -1) were missing a leading zero when cast to string, but only if they had precision/scale.

My test plan is to:

1) Change row 1 in dvt_core_types.col_dec_10_2 to contain a value < 1. This is testing a column with a precision and scale. It will weed out any issues with other engines. Other issues that are identified will result in new issues and disabled tests.
2) Change row 1 in dvt_ora2pg_types.col_num to contain a negative value > -1. This will test columns that have no precision/scale which I believe applies to Oracle and PostgreSQL only. It also tests a negative value.

Because our test tables are not version controlled I need to time this correctly and after merging this PR, merge `develop` into other feature branches.

### DIFF
--- a/third_party/ibis/ibis_addon/operations.py
+++ b/third_party/ibis/ibis_addon/operations.py
@@ -378,7 +378,10 @@ def sa_cast_decimal_when_scale_padded_fmt_fm(t, op):
         #     return (sa.cast(sa.func.trim_scale(arg), typ))
         precision = arg_dtype.precision or 38
         fmt = (
-            "FM" + ("9" * (precision - arg_dtype.scale)) + "." + ("9" * arg_dtype.scale)
+            "FM"
+            + ("9" * (precision - arg_dtype.scale - 1))
+            + "0."
+            + ("9" * arg_dtype.scale)
         )
         return sa.func.rtrim(sa.func.to_char(sa_arg, fmt), ".")
     return None
--- a/third_party/ibis/ibis_addon/operations.py
+++ b/third_party/ibis/ibis_addon/operations.py
@@ -480,23 +480,6 @@ def sa_cast_mysql(t, op):
     return sa_fixed_cast(t, op)
 
 
-def sa_cast_oracle(t, op):
-    arg = op.arg
-    typ = op.to
-    arg_dtype = arg.output_dtype
-
-    sa_arg = t.translate(arg)
-    if arg_dtype.is_binary() and typ.is_string():
-        # Binary to string cast is a "to hex" conversion for DVT.
-        return sa.func.lower(sa.func.rawtohex(sa_arg))
-    elif arg_dtype.is_string() and typ.is_binary():
-        # Binary from string cast is a "from hex" conversion for DVT.
-        return sa.func.hextoraw(sa_arg)
-
-    # Follow the original Ibis code path.
-    return sa_fixed_cast(t, op)
-
-
 def sa_cast_snowflake(t, op):
     custom_cast = sa_cast_decimal_when_scale_padded_fmt_fm(t, op)
     if custom_cast is not None:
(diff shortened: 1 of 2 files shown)


nehanene15 (approved):
Makes sense - LGTM!

PR #1206, opened 2024-07-25
fix: More robust __exit__ on DataValidation()

### LINKED ISSUE #1195: Multi table validations exhausted PostgreSQL connections

nj1973:


data_validation/data_validation.py
```diff
@@ -81,8 +81,10 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_value, exc_traceback):
         if hasattr(self, "config_manager"):
             try:
-                self.config_manager.source_client.con.dispose()
-                self.config_manager.target_client.con.dispose()
+                if hasattr(self.config_manager.source_client, "con"):
+                    self.config_manager.source_client.con.dispose()
```
nehanene15: We might want to check if dispose() is a supported method too. I know in Teradata I get the following now: `PM-WARNING: Exception closing connections: 'TeradataCursor' object has no attribute 'dispose'`

sundar-mudupalli-work: @nj1973 - this might be a problem specific to Postgres. I did not see this issue on Oracle.

Please take a look at my [PR regarding too many files in a config directory](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1130). There by changing the way we did validation, I was able to process about 150 files in sequence.  I was impressed that the python oracle driver and/or the garbage collection were working well.

Sundar Mudupalli

nj1973: In my testing I could see the connection count climbing in both Oracle and PostgreSQL.

nj1973: > We might want to check if dispose() is a supported method too. I know in Teradata I get the following now: `PM-WARNING: Exception closing connections: 'TeradataCursor' object has no attribute 'dispose'`

Hmm, I think I should wrap the `dispose()` calls in a check for Oracle and PostgreSQL. They are the systems with a known problem. I made the assumption that there would be consistency across other engine's interfaces but that was clearly a bad assumption.

nehanene15 (approved):

PR #1208, opened 2024-07-26
fix: Support DATE '0001-01-01' comparison fields

### LINKED ISSUE #1193: [TD to BQ] Row validation with -comp-fields fails for date "0001-01-01"
Fields with datatype DATE and value of `0001-01-01` cause unexpected behaviors on `-comp-fields`.

### Scenario 1
If value 0001-01-01 is present in TD but not present in BQ and date_field is used in in `-pk` the following error is produced: 

```
ValueError: You are trying to merge on object and datetime64[ns] columns. 
If you wish to proceed you should use pd.concat
```

### Scenario 2
If the value 0001-01-01 present in both TD and BQ and date_field in `-comp-fields`, then the following error is produced

```
 File "/home/user/professional-services-data-validator/third_party/ibis/ibis_addon/operations.py", line 556, in string_to_epoch
    parsed_ts = isoparse(ts).astimezone(UTC)
ValueError: year 0 is out of range
```

nehanene15:
Fixes #1193 

Previously dates starting with zero i.e. DATE '0001-01-01' in comparison fields would throw an error when convertin to UTC`isoparse(timestamp).astimezone(UTC)` since it's an invalid UTC timestamp.

This fix implements a fix so date columns can calculate epoch seconds without UTC conversion. The timedelta generated is the same since DATE fields don't support timezones innately.

### DIFF
--- a/third_party/ibis/ibis_addon/operations.py
+++ b/third_party/ibis/ibis_addon/operations.py
@@ -579,8 +579,13 @@ def string_to_epoch(ts: str):
     from dateutil.tz import UTC
     from datetime import datetime, timezone
 
-    parsed_ts = isoparse(ts).astimezone(UTC)
-    return (parsed_ts - datetime(1970, 1, 1, tzinfo=timezone.utc)).total_seconds()
+    try:
+        parsed_ts = isoparse(ts).astimezone(UTC)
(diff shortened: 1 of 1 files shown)


helensilva14 (approved):
LGTM!

Issue #1214, opened 2024-07-31
Need standard format for INTERVAL columns in row validations


nj1973:
Casting interval columns to string gives differing values based on the SQL engine. For example:

- Oracle: `+02 03:04:55.666`
- PostgreSQL: `2 days 03:04:55.666`
- BigQuery: `0-0 2 3:4:55.666`

There does not appear to be formatting functions like there are for timestamps so we need to think about how to approach this.

Also consider whether the changes for this issue can cater for column min/max validations.

Search for "issue-1214" in our tests to locate our currently disabled tests.

sundar-mudupalli-work:
Repeating comment made in issue 1547 here

Oracledb package supports Interval data type as documented [here](https://python-oracledb.readthedocs.io/en/latest/user_guide/appendix_a.html#supported-oracle-database-data-types). Interval days, minutes and seconds is mapped to python timedelta type. Oracle assumes 24 hours per day. IntervalYM has its own Oracle type which is a tuple with Years and months as integers.

Postgres on the other hand Internally, interval values are stored as three integral fields: months, days, and microseconds. These fields are kept separate because the number of days in a month varies, while a day can have 23 or 25 hours if a daylight savings time transition is involved.

BigQuery's Interval type is pre GA as of this writing. See [here](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#interval_type)

Interesting.

Sundar Mudupalli

nj1973:
For hash/concat validation we'll need a common string based format for intervals. I did a little research and can't find a common way to format the output.

PostgreSQL
```
SET intervalstyle = 'sql_standard';
select cast(col_interval_ds as text),cast(col_interval_ym as text) from pso_data_validator.dvt_intervals order by id;

 col_interval_ds | col_interval_ym
-----------------+-----------------
 2:03:44         | 0-2
 1 2:03:44       | 1-2
 2 2:03:44.123   | 2-2
 30 22:33:44     | 30-11

SET intervalstyle = 'iso_8601';
select cast(col_interval_ds as text),cast(col_interval_ym as text) from pso_data_validator.dvt_intervals order by id;

 col_interval_ds | col_interval_ym
-----------------+-----------------
 PT2H3M44S       | P2M
 P1DT2H3M44S     | P1Y2M
 P2DT2H3M44.123S | P2Y2M
 P30DT22H33M44S  | P30Y11M
```

BigQuery
```
[...]

PR #1215, opened 2024-07-31
fix: Oracle INTERVAL exception in validate schema

### LINKED ISSUE #1211: Oracle exception validating table with INTERVAL DAY TO SECOND column
Example validate schema command on table with INTERVAL DAY TO SECOND column, we tried to exclude the column to make the command pass but still get the exception:
```
data-validation validate schema --source-conn ora --target-conn pg \
--tables-list owner.t1 \
--exclusion-columns col_interval
...
  File "C:\Users\u\DVT\lib\site-packages\multipledispatch\dispatcher.py", line 269, in __call_:
    func = self._cache[types]
KeyError: (<class 'sqlalchemy.dialects.oracle.cx_oracle.OracleDialect_cx_oracle'>, <class 'sqlalchemy.dialects.oracle.base.INTERVAL'>)
...
  raise NotImplementedError(
NotImplementedError: Could not find signature for dtype: <OracleDialect_cx_oracle, INTERVAL>
```

The reported problem is for `validate schema`, we should also test `column` and `row` to understand what is required there.

nj1973:
This PR fixes an exception when validating an Oracle table containing an INTERVAL DAY TO SECOND column.

It also adds the column to the `dvt_ora2pg_types` test table.

It does not:

1. Does not enable INTERVAL YEAR TO MONTH. This data type is not supported at all by cx-Oracle and therefore needs a different approach
2. Does not fix row or column validation. This is a bigger, not Oracle specific issue: https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1214


helensilva14:
> Does not enable INTERVAL YEAR TO MONTH. This data type is not supported at all by cx-Oracle and therefore needs a different approach

Regarding this observation: does the customer have columns with this other data type at the moment?

helensilva14 (approved):
LGTM!

nj1973:
> > Does not enable INTERVAL YEAR TO MONTH. This data type is not supported at all by cx-Oracle and therefore needs a different approach
> 
> Regarding this observation: does the customer have columns with this other data type at the moment?

Not that they have shared with me. They have 100s of databases though so it would not surprise me if it came up. As I said above though, the solution will be different so I don't think it should be connected to this fix.

PR #1220, opened 2024-08-08
fix: TD to BQ generate partitions with date PKs

### LINKED ISSUE #1191: [TD] Invalid literal for pk filter in YAML from generate-partitions using date field as pk
When using Teradata as a source and using DATE field as primary key the generate table partitions commands will generate a yaml with a filter field like the one below

```
filters:
  - source: date_field <='2025-07-03'
    target: date_field <='2025-07-03'
    type: custom
  - source: ' ( date_field <= ''2025-07-03'' ) AND ( "date_field" < ''2024-01-20T00:00:00''
      )'
    target: ' ( date_field <= ''2025-07-03'' ) AND ( `date_field` < CAST(''2024-01-20
      00:00:00'' AS DATETIME) )'
    type: custom
```

The  `"date_field" < ''2024-01-20T00:00:00''` filter is not valid and will cause the following error on TD:

`teradatasql.OperationalError: [Version 20.0.0.12] [Session 4806] [Teradata Database] [Error 3535] A character string failed conversion to a numeric value.`

nehanene15:
Fixes #1191 

- Fixes rtrim() log message so that it only is logged when string comparison fields are present
- Adds check for date columns in generate partitions so that date literals are generated. This will only affect date PKs.


sundar-mudupalli-work:
Hi,

I updated the generate partitions test case to include dates and datetime and both rendered correctly in the partition filters. I noticed that strings which have embedded quotes are not handled properly and opened [issue 1221](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1221).

Otherwise good to go.

Sundar Mudupalli

sundar-mudupalli-work (approved):
Changes look good. I added test cases. Hopefully this can be merged into main.

Good work.

helensilva14 (approved):
LGTM as well! Changes make sense

PR #1226, opened 2024-08-11
fix: TD to BQ - Support hash row validation with Latin and Unicode Characters in Strings

### LINKED ISSUE #1170: Accented string causes validation_status=fail in TD VS BQ row validation

### LINKED ISSUE #1225: Row Hash validation of TD to BQ with non latin characters fails
If a Teradata column contains a string which has Unicode Characters beyond the Latin character set, DVT fails with the following error:
```
 [Version 20.0.0.12] [Session 8285] [Teradata Database] [Error 6706] The string contains an untranslatable character.
```
If the row with the offending character is excluded from the validation, the validation succeeds. 

The problem is with [this line](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/89ef449d9f68281fcf2a0eb765b1c425f7fb792f/third_party/ibis/ibis_addon/operations.py#L163)  
```
 return f"rtrim(hash_sha256(TransUnicodeToUTF8(TRANSLATE({arg} using latin_to_unicode WITH ERROR))))") 
```
This line assumes that all characters can be converted into Latin - which is then converted to Unicode and UTF8. When a non latin character is presented Teradata throws the error.  One solution is to have the user provide an option to DVT `--beyond-latin` or `-byl` which would indicate that table columns are Unicode and/or some table columns have characters beyond the latin character set. In that situation, the backend can generate the following SQL for Teradata
```
return f"rtrim(hash_sha256(TransUnicodeToUTF8({arg})))")
```
If we did the above when there are columns with latin encoding and latin character set, the validation will fail. 

Sundar Mudupalli

sundar-mudupalli-work:
Fixes [1170 correctly](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1170) and [issue 1225](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1225)

Adds test case for checking latin hash row, also hash validation for dvt pangrams.

This requires an updated `sha256` function that supports VARBYTE in teradata.  That is available separately.


helensilva14:
> This requires an updated sha256 function that supports VARBYTE in teradata. That is available separately.

@sundar-mudupalli-work shouldn't we provide the code for this updated function here somehow?

nehanene15 (approved):
LGTM!

PR #1233, opened 2024-08-19
feat: Auto split row concat/hash validations when many columns

### LINKED ISSUE #1216: Exception in PostgreSQL row validation for table with many columns
Exception in PostgreSQL row validation for table with many columns:
```
sqlalchemy.exc.OperationalError: (psycopg2.errors.TooManyArguments) cannot pass more than 100 arguments to a function

LINE 222: ...per__dvt_calc_col_103 AS upper__dvt_calc_col_103, concat_ws(...
```

I was able to reproduce this easily enough on PostgreSQL but also had problems with Oracle:
```
cx_Oracle.DatabaseError: ORA-01489: result of string concatenation is too long
```

I wonder if we should break up validations with > 100 columns into multiple validations and not try to manipulate complex concat expressions for each supported engine. The down side to this would be multiple hashes for each primary key. Needs some discussion.

nj1973:
Closes #1216.

A few of our supported SQL engines have limits on the number of columns that can be concatenated into a string.

For example:

- PostgreSQL: `psycopg2.errors.TooManyArguments) cannot pass more than 100 arguments to a function`
- Oracle: `ORA-01489: result of string concatenation is too long`
- SQL Server: `The concat function requires 2 to 254 arguments. (189) (SQLExecDirectW)`
- Teradata: `[Error 3556] Too many columns defined for this table`

This PR introduces a new option for row validations (including custom-query) to limit the number of columns processed in a single query: `--max-concat-columns`. If the number of columns to be processed by a validation is greater than `--max-concat-columns` then the validation is implicitly split into multiple validations.

The `--max-concat-columns` option had an adaptive default. For engines with no known limit it is uncapped and prior behaviour is retained. For engines defined in the new constant `consts.MAX_CONCAT_COLUMNS_DEFAULTS` we impose a limit. Users can override this at runtime. If we find new limits for other engines then the user has the runtime option as a workaround and we can easily modify the constant.

When a validation is split into multiple validations due to the number of columns we output an INFO level log message to inform the user.

I've added a new integration test table `pso_data_validator.dvt_many_cols` and tests for both standard and custom-query validations. Note that I have disabled the Hive tests due to time they take and that they often cause our test infrastructure to become unstable.

Even without the Hive tests this PR adds 16 more integration tests, I wonder if we don't need custom-query tests for every engine. Just one or two of the ones with a lower limit perhaps?


data_validation/cli_tools.py
```diff
@@ -1293,8 +1400,41 @@ def get_pre_build_configs(args: Namespace, validate_cmd: str) -> List[Dict]:
...
+            )
+            additional_pre_build_configs = _concat_column_count_configs(
+                cols,
+                pre_build_configs,
+                consts.CONFIG_ROW_HASH if args.hash else consts.CONFIG_ROW_CONCAT,
+                _max_concat_columns(
+                    args.max_concat_columns, source_client, target_client
+                ),
+            )
+            if len(additional_pre_build_configs) > 1:
```
helensilva14: For my understanding by the variable name, should we check for `>= 1`? Such as if have only 1 additional file it shows the INFO log, not only for 2 or more files. If not it means the default single file is also counted on `additional_pre_build_configs`?

data_validation/clients.py
```diff
@@ -155,28 +172,29 @@ def get_ibis_query(client, query):
...
         return client.get_schema(table_name, schema_name)
 
 
+def get_ibis_query_schema(client, query_str) -> "sch.Schema":
+    if client.name in IBIS_ALCHEMY_BACKENDS:
+        ibis_query = get_ibis_query(client, query_str)
+        return ibis_query.schema()
+    else:
+        # NJ: I'm not happy about calling a private method but don't see how I can avoid it.
+        return client._get_schema_using_query(query_str)
```
helensilva14, line 196: We can discuss more about that!

tests/resources/sqlserver_test_tables.sql
```diff
@@ -124,3 +124,408 @@ INSERT INTO pso_data_validator.dvt_pangrams VALUES (2,'Polish', 'Pchnąć w tę
...
+, col_391 decimal(1)
+, col_392 decimal(1)
+, col_393 decimal(1)
+, col_394 decimal(1)
+, col_395 decimal(1)
+, col_396 decimal(1)
+, col_397 decimal(1)
+, col_398 decimal(1)
+, col_399 decimal(1)
+);
```
helensilva14, line 530: SQL Server table is missing the description provided for the others: `'Integration test table used to test validating many columns.'`

third_party/ibis/ibis_cloud_spanner/tests/ddl.sql
```diff
@@ -79,3 +79,405 @@ CREATE TABLE dvt_binary
...
+, col_391 INT64
+, col_392 INT64
+, col_393 INT64
+, col_394 INT64
+, col_395 INT64
+, col_396 INT64
+, col_397 INT64
+, col_398 INT64
+, col_399 INT64
+) PRIMARY KEY (id);
```
helensilva14, line 483: Spanner table is missing the description provided for the others: `'Integration test table used to test validating many columns.'`

tests/system/data_sources/test_mysql.py
```diff
@@ -398,7 +399,7 @@ def test_row_validation_pangrams_to_bigquery():
             "row",
             "-sc=mysql-conn",
             "-tc=bq-conn",
-            "-tbls=PSO_DATA_VALIDATOR.PUBLIC.DVT_PANGRAMS=pso_data_validator.dvt_pangrams",
+            "-tbls=pso_data_validator.dvt_pangrams",
```
helensilva14, line 402: There was a change for the MySQL table's schema?

tests/system/data_sources/test_hive.py
```diff
@@ -375,3 +391,36 @@ def test_custom_query_validation_core_types():
...
+)
+def test_row_validation_many_columns():
+    """Hive dvt_many_cols row validation.
+    Using BigQuery target because Hive queries are really slow.
+    When executed individually this test passes but when executed during a full
+    integration test run I get:
+      impala.error.HiveServer2Error: java.lang.NullPointerException
+    This must be down to minimal resources for our small Hive instance, disabling for now.
+    """
+    pytest.skip("Skipping test_row_validation_many_columns due to resource issues.")
```
helensilva14, line 408: I suggest adding a TODO comment for us to remember to remove this skip eventually

helensilva14 (changes requested):
Just minor comments, super cool implementations Neil, good job!

data_validation/cli_tools.py
nj1973: I didn't want to re-use the name `pre_build_configs` because it was also used as an input into the function. I've changed the new from `additional_pre_build_configs` to `new_pre_build_configs`, I hope that makes it clearer. I still only want to inform the user if there are `> 1` of them, if there's only 1 then we haven't split the validations and there's no need to output a message.

data_validation/clients.py
nj1973: Ibis does not expose a public method like it does for get_schema(). I didn't feel it was right for me to add a public method to Ibis objects, I know we already patch the private method but adding new methods seemed a step too far. Happy to be influenced in another direction though.

tests/resources/sqlserver_test_tables.sql
nj1973: As far as I can tell SQL Server doesn't have an easy way to add table comments. You'll notice none of the tables in this file have comments.

third_party/ibis/ibis_cloud_spanner/tests/ddl.sql
nj1973: I can't see in Spanner DDL how to add table comments. You'll notice none of the tables in this file have comments.

tests/system/data_sources/test_mysql.py
nj1973: No, the test was wrong but disabled so it didn't really matter. I just noticed it had the Snowflake schema name so I fixed it.

tests/system/data_sources/test_hive.py
nj1973: Agreed

data_validation/cli_tools.py
helensilva14: Ahh got it, the new name makes easier to understand that, thanks!

tests/resources/sqlserver_test_tables.sql
helensilva14: That's right, I didn't notice the other tables don't have it. I did some search too and yeah, apparently it's something more UI based it seems? All good then!

data_validation/clients.py
helensilva14: Well, since you added a comment about this current implementation I'd be okay to move forward as it is so we can unblock the customer. It seems like something we can analyze further later on.

helensilva14 (approved):
LGTM!

Issue #1161, opened 2024-06-06
docs: Add list of log levels options to README.md


helensilva14:
Taking this merged PR as reference https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/577, describe in our main documentation page what are the log levels the users can pass as parameters

PR #1240, opened 2024-08-23
feat: BigQuery result handler logs textual output at DEBUG instead of INFO

### LINKED ISSUE #871: Bigquery result handler to use Text result handler for stdout
Currently the Biquery result handler outputs the results to stdout by running 
```
            result_df.drop(consts.COLUMN_FILTER_LIST, axis=1).to_markdown(
                tablefmt="fancy_grid", index=False
            )
```
which is nearly identical to the default table output from the Text result handler. 

A useful enhancement would be for this to call the Text result handler instead, which would then mean the stdout format could be set with the --format flag when using the Bigquery result handler as well. Currently this flag is ignored unless the Text result handler is being used. 

For example I might want to store results in BigQuery, but also have the stdout output formatted as JSON. This isn't currently possible because the BigQuery result handler only outputs to stdout using a table format as above.

nj1973:
This PR:

1. Changes the BigQuery result handler log level to DEBUG instead of INFO for the textual results output.
2. Allows the `--format` option to control the logged output, I don't think this will be used but it was requested in the issue and easy to do.
3. Writes a simple INFO log message stating the results were written to BigQuery, e.g.:

```
$ data-validation validate row -sc=ora -tc=bq \
-tbls=pso_data_validator.dvt_core_types --primary-keys=id \
--concat='id,col_dec' -bqrh=project-01.dvt_test.results --format=csv
08/23/2024 02:02:12 PM-INFO: Results written to BigQuery, run id: 836f7cce-94e8-43bb-86a8-02655d83c6a3
$
```


helensilva14 (approved):
Makes sense to me! @nehanene15 what do you think?

helensilva14 (approved):

nehanene15 (approved):
Code looks great. I think we could also add some documentation on the --log-level flag in the README (#1161) but not necessary

PR #1241, opened 2024-08-23
fix: support generate partitions for one row per partition

nehanene15:
Support generate partitions with one row per partition i.e. num_partitions = number of rows.


data_validation/partition_builder.py
sundar-mudupalli-work, line 1: Neha,

I have a question - back in [this line](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/cd831647d7ab3d199d62e7e4c7d80c7a0efaed71/data_validation/partition_builder.py#L172) we check that the number of parititions is smaller than the number of rows in the source table. If that is true, how is it that we are getting `first_elements` to be null ? I don't understand.

Thanks.

Sundar Mudupalli

nehanene15: In this case the query that populates first_elements returned 0 rows. In your comments you wrote: row number * # of partitions % total number of rows has to be > 0 and <= number of partitions. Because source row count = target_row_count = partition_num it's failing...

sundar-mudupalli-work: Yeah - I think the problem is [this line](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/cd831647d7ab3d199d62e7e4c7d80c7a0efaed71/data_validation/partition_builder.py#L205).
It should read
```
cond = True if source_count == number_of_part else ...
```
Basically, cond will be true (choose all rows) if source_count == number_of_part else <current expression>

That should fix the issue.

Sundar Mudupalli

nehanene15: This worked! Using 'True' didn't work to select all rows, but using 'rownum_table' so that the condition is `rownum_table[rownum_table].order_by(source_pks)` does select all rows. ([Ibis doc](https://ibis-project.org/tutorials/ibis-for-sql-users#select-equivalent))

`cond = rownum_table if source_count == number_of_part else ...`

helensilva14 (approved):
LGTM! Great collab Neha & Sundar!

PR #1243, opened 2024-08-26
feat: Adding exclude_columns flag for row validation hash

mloeberc:
Closes issue https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/880

Added parameter --exclude-columns or -ec  as a flag that identifies the column list as an exclusion, and not an inclusion when calculating hash operations in row validation and custom query row.
Updated Read Me with new parameter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,7 +185,7 @@ data-validation (--verbose or -v) (--log-level or -ll) validate row
   --hash COLUMNS        Comma separated list of columns to hash or * for all columns
   --concat COLUMNS      Comma separated list of columns to concatenate or * for all columns (use if a common hash function is not available between databases)
   [--exclude-columns or -ec]
-                        Flag to indicate the list of columns provided should be excluded and not included.
+                        Flag to indicate the list of columns provided should be excluded from hash or concat instead of included.
   [--bq-result-handler or -bqrh PROJECT_ID.DATASET.TABLE]
                         BigQuery destination for validation results. Defaults to stdout.
                         See: *Validation Reports* section
@@ -409,7 +409,7 @@ data-validation (--verbose or -v) (--log-level or -ll) validate custom-query row
   --primary-key or -pk JOIN_KEY
                        Common column between source and target tables for join
   [--exclude-columns or -ec]
-                        Flag to indicate the list of columns provided should be excluded and not included.
+                        Flag to indicate the list of columns provided should be excluded from hash or concat instead of included.
   [--bq-result-handler or -bqrh PROJECT_ID.DATASET.TABLE]
                         BigQuery destination for validation results. Defaults to stdout.
                         See: *Validation Reports* section
--- a/README.md
+++ b/README.md
@@ -184,6 +184,8 @@ data-validation (--verbose or -v) (--log-level or -ll) validate row
                         See: *Calculated Fields* section for details
   --hash COLUMNS        Comma separated list of columns to hash or * for all columns
   --concat COLUMNS      Comma separated list of columns to concatenate or * for all columns (use if a common hash function is not available between databases)
+  [--exclude-columns or -ec]
+                        Flag to indicate the list of columns provided should be excluded and not included.
   [--bq-result-handler or -bqrh PROJECT_ID.DATASET.TABLE]
                         BigQuery destination for validation results. Defaults to stdout.
                         See: *Validation Reports* section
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -184,6 +184,8 @@ data-validation (--verbose or -v) (--log-level or -ll) validate row
                         See: *Calculated Fields* section for details
   --hash COLUMNS        Comma separated list of columns to hash or * for all columns
   --concat COLUMNS      Comma separated list of columns to concatenate or * for all columns (use if a common hash function is not available between databases)
+  [--exclude-columns or -ec]
```
helensilva14, line 187: Since it's a flag and not a list of cols, only for `column` and `row` validations, I think it should have a specific name to differentiate it from the feature `[--exclusion-columns or -ec EXCLUSION_COLUMNS] Comma separated list of columns to be excluded from the schema validation, e.g.: col_a,col_b.` available for `schema` validations.

```suggestion
  [--exclude-columns-list or -ecl]
```

Comment applies for lines 109, 188, 347, 411 as well

data_validation/__main__.py
```diff
@@ -174,14 +174,18 @@ def get_calculated_config(args, config_manager: ConfigManager) -> List[dict]:
             if config_manager.hash == "*"
             else cli_tools.get_arg_list(config_manager.hash)
         )
-        fields = config_manager.build_dependent_aliases("hash", col_list)
+        fields = config_manager.build_dependent_aliases(
+            "hash", col_list, args.exclude_columns
```
helensilva14, line 178:
```suggestion
            "hash", col_list, config_manager.exclude_columns_list
```

Change will work only after having a new `CONFIG_EXCLUDE_COLUMNS_LIST` variable added to `consts.py` and a equivalent function added to `config_manager.py` to get it. Suggested structure for the function:


```python
    def exclude_columns_list(self):
        """Return if the validation should exclude list of columns provided instead of including"""
        return self._config.get(consts.CONFIG_EXCLUDE_COLUMNS_LIST) or False
```

Comment applies for lines 104, 111, 120, 129, 138, 147, 156, 187 as well

data_validation/config_manager.py
```diff
@@ -1031,17 +1033,35 @@ def build_dependent_aliases(self, calc_type: str, col_list=None) -> List[Dict]:
...
+                    for (k, v) in casefold_source_columns.items()
+                    if k in casefold_col_list
+                }
+                casefold_target_columns = {
+                    k: v
+                    for (k, v) in casefold_target_columns.items()
+                    if k in casefold_col_list
+                }
+        else:
+            if exclude_cols:
```
```suggestion
        elif exclude_cols:
```

data_validation/config_manager.py
```diff
@@ -1031,17 +1033,35 @@ def build_dependent_aliases(self, calc_type: str, col_list=None) -> List[Dict]:
...
-                for (k, v) in casefold_source_columns.items()
-                if k in casefold_col_list
-            }
-            casefold_target_columns = {
-                k: v
-                for (k, v) in casefold_target_columns.items()
-                if k in casefold_col_list
-            }
+            if exclude_cols:
+                # Filter columns based on col_list if provided
```
```suggestion
                # Exclude columns based on col_list, as provided
```

data_validation/config_manager.py
```diff
@@ -1031,17 +1033,35 @@ def build_dependent_aliases(self, calc_type: str, col_list=None) -> List[Dict]:
...
+                    for (k, v) in casefold_source_columns.items()
+                    if k not in casefold_col_list
+                }
+                casefold_target_columns = {
+                    k: v
+                    for (k, v) in casefold_target_columns.items()
+                    if k not in casefold_col_list
+                }
+            else:
+                # Filter columns based on col_list if provided
```
```suggestion
                # Include columns based on col_list, as provided
```

data_validation/__main__.py
helensilva14: Ah I forgot to mention: it seems this explicit call for `args.exclude_columns` can be the root cause behind the integration tests failure for MySQL

README.md
nehanene15: Yeah the `exclusion-columns` fore schema validation is different than the `-exclude-columns` for row/column validation. Maybe we can standardize the approach in another PR but for now this one is standard across column + row validations

data_validation/__main__.py
nehanene15: The issue here was that the exclude_columns flag wasn't added to the arg parser - it is now! Similar to the existing column validation flag, it isn't stored in the config manager...it's only used to make the CLI easier. If you add the -ec flag, it won't be stored in the YAML config, but it will change the columns added to the YAML itself

README.md
helensilva14: For sure, it's not critical for this feature to function, I can open a follow-up issue for us to perform this change!

data_validation/__main__.py
helensilva14: Ahhh got it! Makes more sense haha thanks Neha!

helensilva14 (approved):
LGTM!

README.md
helensilva14: Issue #1244 created!

PR #1245, opened 2024-08-28
fix: Oracle validations of tables with hyphen in name

### LINKED ISSUE #1192: Exception validating Oracle table with hyphen in name
Oracle test table:
```
create table dvt_test."TAB-HYPHEN"
(id number(5) primary key, data varchar2(30));
insert into dvt_test."TAB-HYPHEN" values (1,'Row 1');
commit;
```

PostgreSQL test table:
```
create table dvt_test."tab-hyphen"
(id bigint primary key, data varchar(30));
insert into dvt_test."tab-hyphen" values (1,'Row 1');
```

Test command, PostgreSQL vs PostgreSQL:
```
data-validation validate column -sc=pg -tc=pg -tbls=dvt_test.tab-hyphen

Success
```

Test command, Oracle vs Oracle:
```
data-validation validate column -sc=ora -tc=ora -tbls=dvt_test.tab-hyphen

  File "/professional-services-data-validator/.venv/bin/data-validation", line 8, in <module>
    sys.exit(main())
             ^^^^^^
...
  File "/professional-services-data-validator/data_validation/validation_builder.py", line 391, in get_source_query
    table = self.config_manager.get_source_ibis_table()
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
  File "/professional-services-data-validator/.venv/lib/python3.11/site-packages/sqlalchemy/engine/reflection.py", line 789, in reflect_table
    raise exc.NoSuchTableError(table_name)
sqlalchemy.exc.NoSuchTableError: tab-hyphen
```

nj1973:
Before this PR, SQL Alchemy has conflicting behaviour around Oracle table names.

1. Any table name is uppercased before querying the data dictionary **unless** it contains special characters. Really the presence of special characters is irrelevant for this decision. We either uppercase or don't uppercase, special characters should not be a factor.
2. Any table name is double quoted during query generation if it contains special characters but it is not upper cased. The double quotes force Oracle to look for a lower case table name and fail. If we are forcing an upper case name for data dictionary (item 1 in this list) then we may as well force upper case in item 2 also.

This PR changes both code paths above so we always upper case the name in item 1 and be sure to upper case it if adding double quotes in item 2.

This PR also adds a test table and tests across many engines although some tests have been left inactive and referenced in this issue: https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1271


helensilva14 (approved):
Great work, LGTM!

PR #1259, opened 2024-09-09
fix: Using datetime constants in filters

### LINKED ISSUE #1255: Filter is not working with Datetime in column validation
`--filters "update_timestamp >= '2024-04-01 16:00:00 UTC'"`

when specified with column validation option it does not work and throws,
` ValueError("Unable to parse filter arguments.") line 1090, in get_filters`
 
Its splitting the timestamp into multiple values due to colon, --filter is trying to split the timestamp assuming there are source and target filters (as per the documentation --filter can be used with **source_filter : target_filter** )

it seems issue is with this line :https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/d01a40b14e5e090c1a99aa44fd8625fa8e515427/data_validation/cli_tools.py#L1073 

its not allowing to escape/handle the colons.

sundar-mudupalli-work:
Correctly parses `--filters` to handle datetime constants or other string literals that could have embedded ':'

Closes [1255](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1255)

### DIFF
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -1065,7 +1065,7 @@ def get_labels(arg_labels):
     return labels
 
 
-def get_filters(filter_value: str) -> list[dict]:
+def get_filters(filter_value: str) -> List[Dict]:
     """Returns filters for source and target from --filters argument.
     A filter is the condition that is used in a SQL WHERE clause.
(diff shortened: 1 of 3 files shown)


nehanene15 (approved):
LGTM!

PR #1263, opened 2024-09-11
fix: big query run id message

nick-redfearn:
I started out on this PR by trying to add additional unit test scenarios around running row level validations where I'm writing results to BQ.

In particular these scenarios...

1. log level = INFO + some results returned with filter_status = fail -> Does not write results to console + we get the run id message
2. log level = INFO + no results returned with filter_status = fail -> Does not write results to console + we get the "No results" message
3. log level = DEBUG + some results returned with filter_status = fail -> Does write results to console + we get the run id message
4. log level = DEBUG + no results returned with filter_status = fail -> Does not write results to console (because there are none) + we get the "No results" message

I care about this stuff because I don't want to write sensitive info to the console and I want to keep track of the run IDs (or lack of one where there is no row validation failure)

I found that I was getting an exception when I was setting filter_status = fail.
Once I looked into it, I found that `result_df["run_id"][0]` was looking for a key = 0, which did not exist in my test case.
`result_df.iloc[0]["run_id"]` returns the run id for the first item in the DF so should get around this bug.

Related PR -> https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1240

### DIFF
--- a/tests/unit/test_data_validation.py
+++ b/tests/unit/test_data_validation.py
@@ -853,7 +853,7 @@ def test_console_data_shown_for_matching_validation_with_result_written_to_bq_in
     # Then...
     # 0 failures returned
     assert len(result_df) == 0
-    # The "Results written" message happens + "Empty DataFrame" because there are no failures to display
+    # The "No results" message happens + "Empty DataFrame" because there are no failures to display
     assert len(caplog.records) == 2
     assert caplog.records[0].message == "No results to write to BigQuery"
     assert caplog.records[1].message.startswith("Empty DataFrame")
--- a/tests/unit/test_data_validation.py
+++ b/tests/unit/test_data_validation.py
@@ -401,10 +401,10 @@ SAMPLE_ROW_CONFIG_BQ_FAILURES = {
         },
     ],
     consts.CONFIG_RESULT_HANDLER: {
-            consts.CONFIG_TYPE: "BigQuery",
-            consts.PROJECT_ID: "my-project",
-            consts.TABLE_ID: "dataset.table_name",
-        },
+        consts.CONFIG_TYPE: "BigQuery",
+        consts.PROJECT_ID: "my-project",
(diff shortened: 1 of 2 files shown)


nick-redfearn:
@nj1973 @nehanene15 @helensilva14 - Happy to review?

tests/unit/test_data_validation.py
```diff
@@ -363,6 +366,49 @@
...
+        },
+        {
+            consts.CONFIG_FIELD_ALIAS: "text_value",
+            consts.CONFIG_SOURCE_COLUMN: "text_value",
+            consts.CONFIG_TARGET_COLUMN: "text_value",
+            consts.CONFIG_CAST: None,
+        },
+    ],
+    consts.CONFIG_RESULT_HANDLER: {
+            consts.CONFIG_TYPE: "BigQuery",
```
nj1973: We have some over-indentation here. We check formatting with `black` as part of running tests. Can you run the command below and commit the changes please:

```
python3 -m black tests/unit/test_data_validation.py
```

nj1973 (changes requested):
Thanks for fixing this @nick-redfearn , I introduced this issue recently, my bad :-(

Just a small formatting issue and then I'll approve. Cheers

tests/unit/test_data_validation.py
nicky6s: Thanks @nj1973 - I'll sort this out 👍

nick-redfearn: Thanks @nj1973

Cleaned up here -> https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1263/commits/2266a47c6a4ae3be9b09eba05d731afc3d033024

nj1973 (approved):
LGTM, thanks again.

nick-redfearn:
Thanks all 🙇

PR #1265, opened 2024-09-13
feat: Enhance --tables-list to support schema.* shorthand for find-tables command

### LINKED ISSUE #1254: Exception calling DVT with find-tables results: Argument list too long
A customer is feeding the results of a `find-tables` command directly into a validation command and hitting a BASH exception due to the length of the command:

```
"-bash: /some-path/bin/data-validation: Argument list too long"
```

I created a shell script for them as a workaround but am raising this issue for two reasons:

1. Should I put the admittedly simple shell script in the repo under samples for future reference?
2. We could simplify the interaction between DVT commands and `find-tables` output but adding a shortcut directly into validation commands. For example the `-tbls` option could support having a table name as asterisk which would trigger an implicit call to `find-tables` logic. e.g.:

```
data-validation validate column --source-conn=ora--target-conn=bq \
      --tables-list="test_schema.*" --count="*"
```

DVT would expand `test_schema.*` to all tables in the schema using existing logic. If we did this we might also need to add an exclude tables option in the future.

nj1973:
Two recent projects for different customers have demonstrated the need for this. In both projects the end user wanted to run column validation across the whole schema. They ended up using `find-tables` to pull a list of tables and then feed that into validation commands. The output of find tables can quickly exceed BASH shell limits.

This PR adds a shortcut for validation of all tables in a schema by supplying an asterisk for the table name, for example:

```
data-validation validate column --source-conn=ora --target-conn=bq  \
--tables-list="pso_data_validator.*" --min="*" --count="*"
```

When we see a table name of "str.*" we run the same code that the `find-tables` command runs to get the same output, but without the need for the user to run multiple commands. Because we are using the same code we cannot match partial wildcards, for example `"pso_data_validator.dvt_c*"` matches no tables.

This shortcut is active for all validation types but is only really useful for column and schema validations. I couldn't see a reason to block it for row validations but it isn't useful because of the need for individual primary key values. It is also active when generating config files.

I did consider adding a `--tables-list-exclude` option to allow the user to skip certain tables but didn't because I think that is an independent requirement and not part of this change. Happy to raise an issue for that if others agree it is needed.
 
I only added tests to `test_oracle.py` because I don't think the logic differs across SQL engines, I have tests for Oracle vs BigQuery and Oracle vs PostgreSQL.

Other test changes:
- I added a test for supplying two tables on the command line, not really relevant to this PR but I felt we should have it.
- Initially I was executing validations and refactored how column tests are executed so I could share code. I then moved away from that as I don't need to execute the tests to test these changes. I left the new `column_validation_test` function 
in `common_functions.py` as I think we can roll it out across all test files. I'll create a new issue for this after merging this PR.

### DIFF
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -1397,9 +1397,10 @@ def get_pre_build_configs(args: Namespace, validate_cmd: str) -> List[Dict]:
         filter_status = None
 
     pre_build_configs_list = []
-    tables_list = find_tables.expand_tables_of_asterisk(
-        tables_list, source_client, target_client
-    )
+    if config_type != consts.CUSTOM_QUERY:
+        tables_list = find_tables.expand_tables_of_asterisk(
+            tables_list, source_client, target_client
+        )
     for table_obj in tables_list:
         pre_build_configs = {
             "config_type": config_type,
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -56,8 +56,9 @@ from argparse import Namespace
 from typing import Dict, List, Optional
 from yaml import Dumper, Loader, dump, load
 
-from data_validation import clients, consts, state_manager, gcs_helper
+from data_validation import clients, consts, find_tables, state_manager, gcs_helper
 from data_validation.validation_builder import list_to_sublists
+
 
 CONNECTION_SOURCE_FIELDS = {
     "BigQuery": [
@@ -645,7 +646,7 @@ def _configure_column_parser(column_parser):
         "-tbls",
         default=None,
         required=True,
-        help="Comma separated tables list in the form 'schema.table=target_schema.target_table'",
+        help="Comma separated tables list in the form 'schema.table=target_schema.target_table'. Or shorthand schema.* for all tables.",
     )
     _add_common_arguments(optional_arguments, required_arguments)
 
@@ -1151,7 +1152,7 @@ def get_tables_list(arg_tables, default_value=None, is_filesystem=False):
 
     tables_list = []
     tables_mapping = list(csv.reader([arg_tables]))[0]
-    source_schema_required = False if is_filesystem else True
+    source_schema_required = bool(not is_filesystem)
 
     for mapping in tables_mapping:
         tables_map = mapping.split("=")
@@ -1388,6 +1389,9 @@ def get_pre_build_configs(args: Namespace, validate_cmd: str) -> List[Dict]:
         filter_status = None
 
     pre_build_configs_list = []
+    tables_list = find_tables.expand_tables_of_asterisk(
+        tables_list, source_client, target_client
+    )
     for table_obj in tables_list:
         pre_build_configs = {
             "config_type": config_type,
--- a/data_validation/find_tables.py
+++ b/data_validation/find_tables.py
@@ -116,7 +116,7 @@ def expand_tables_of_asterisk(
 
     We can be very specific in this function, we only expand arguments that are:
       {"schema_name": (str), "table_name": "*"}.
-    No partital wildcards or args that include target_schema/table_name are expanded.
+    No partial wildcards or args that include target_schema/table_name are expanded.
 
     Args:
         tables_list (list[dict]): List of schema/table name dicts.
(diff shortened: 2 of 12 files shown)


data_validation/cli_tools.py
```diff
@@ -646,7 +647,7 @@ def _configure_column_parser(column_parser):
         "-tbls",
         default=None,
         required=True,
-        help="Comma separated tables list in the form 'schema.table=target_schema.target_table'",
+        help="Comma separated tables list in the form 'schema.table=target_schema.target_table'. Or shorthand schema.* for all tables.",
```
helensilva14, line 650: This information is useful to add in our README documentation as well, but as you said in your PR description, maybe let's add only for `column` and `schema` validation sections

data_validation/find_tables.py
```diff
@@ -0,0 +1,147 @@
...
+        list: New version of tables_list with expanded "table_name": "*" entries.
+    """
+    new_list = []
+    for mapping in tables_list:
+        if (
+            mapping
+            and mapping[consts.CONFIG_SCHEMA_NAME]
+            and mapping[consts.CONFIG_TABLE_NAME] == "*"
+            and consts.CONFIG_TARGET_SCHEMA_NAME not in mapping
+            and consts.CONFIG_TARGET_TABLE_NAME not in mapping
```
helensilva14: Not sure if I understood these conditions specially, could you please clarify why they are  needed/used for? 🤔

tests/unit/test_find_tables.py
```diff
@@ -0,0 +1,141 @@
...
+                {"schema_name": "s1", "table_name": "t1"},
+            ],
+            [
+                {
+                    "schema_name": "s1",
+                    "table_name": "t1",
+                },
+            ],
+        ),
+        # Test that asterisk mixed woth other characters is not expanded
```
helensilva14:
```suggestion
        # Test that asterisk mixed with other characters is not expanded
```

data_validation/find_tables.py
```diff
@@ -0,0 +1,147 @@
...
+def expand_tables_of_asterisk(
+    tables_list: list,
+    source_client: "ibis.backends.base.BaseBackend",
+    target_client: "ibis.backends.base.BaseBackend",
+) -> list:
+    """Pre-processes tables_mapping expanding any entries that are "schema.*". A shorthand for "find-tables" command.
+
+    We can be very specific in this function, we only expand arguments that are:
+      {"schema_name": (str), "table_name": "*"}.
+    No partital wildcards or args that include target_schema/table_name are expanded.
```
helensilva14:
```suggestion
    No partial wildcards or args that include target_schema/table_name are expanded.
```

helensilva14 (commented):
Great feature and awesome improvements to our tests, good job Neil!

data_validation/cli_tools.py
nj1973: Good suggestion, thanks.

data_validation/find_tables.py
nj1973: I added a comment and changed the logic slightly in the hope of making it clearer. Good suggestion again, thanks.

data_validation/find_tables.py
```diff
@@ -132,8 +132,9 @@ def expand_tables_of_asterisk(
             mapping
             and mapping[consts.CONFIG_SCHEMA_NAME]
             and mapping[consts.CONFIG_TABLE_NAME] == "*"
-            and consts.CONFIG_TARGET_SCHEMA_NAME not in mapping
-            and consts.CONFIG_TARGET_TABLE_NAME not in mapping
+            # Looking for schema.* without a target side qualifier.
+            and not mapping.get(consts.CONFIG_TARGET_SCHEMA_NAME, None)
+            and not mapping.get(consts.CONFIG_TARGET_TABLE_NAME, None)
```
helensilva14, line 137: Ahhh I get it now, makes sense, thanks! Only the comment already clarifies a lot, we can go back to the initial approach since it's simpler to read and understand, what do you think?

nj1973: I think we leave the new code because it is more robust.

helensilva14: That's right, no prob. Maybe it's the case to add a test to guarantee we skip this scenario? Something similar to the new one `test_column_multi_table_all_config_managers`

nj1973: I've added an extra unit test. Good suggestion, thanks.

helensilva14 (approved):
Awesome, LGTM!

PR #1281, opened 2024-09-27
fix: cast count to bigint when generating partitions

### LINKED ISSUE #1280: Error when attempting to create partitions on large Teradata tables (2B+ rows)
When creating partitions for a large table in a Teradata to BigQuery migration, the following error was encountered:

teradatasql.OperationalError: [Version 20.0.0.15] [Session 64925450] [Teradata Database] [Error 2616] Numeric overflow occurred during computation

image

DVT is failing when generating the partitions as it is trying to do count(*) before partition creation.

According to the documentation, a TD integer can only be between -2,147,483,648 and 2,147,483,647. The table in question has 2,234,011,695 rows, which is outside the Integer range(https://docs.teradata.com/r/Enterprise_IntelliFlex_VMware/SQL-Data-Types-and-Literals/Numeric-Data-Types/INTEGER-Data-Type).

Suggestion is to change the count check to cast(count(1) as BIGINT) which should resolve the error.

nehanene15:
Closes #1280 

Casts count to BIGINT so that large count values will not throw a NumericOverflow error.

### DIFF
--- a/data_validation/query_builder/partition_row_builder.py
+++ b/data_validation/query_builder/partition_row_builder.py
@@ -62,4 +62,4 @@ class PartitionRowBuilder(object):
 
     def get_count(self) -> int:
         """Return a count of rows of primary keys - they should be all distinct"""
-        return self.query[self.primary_keys].count().execute()
+        return self.query[self.primary_keys].count().force_cast("int64").execute()


helensilva14 (approved):
LGTM, a very precise change and fix!

PR #1285, opened 2024-10-01
fix: Add proper handling for GCPSecretManager exceptions

### LINKED ISSUE #1242: secret-manager failed to take the secrets for db2 connection
I am working on connecting DB2 and BQ and created a connection json for DB2 like below
{
"source_type":"DB2",
"secret_manager_type":"GCP",
"secret_manager_project_id":"12345678910",
"host":"",
"port":"",
"user":"XXXX",
"password":"secret_name",
"database":""
}

I have a secret name like 'secret_name' in the project with values as json like below,
projects/12345678910/secrets/secret_name

**secret_name: password**

When i hardcoded the user name and password and add the connection.json in bucket and i am able to connect DB2 and BQ. However when i tried to use the secret manager i am unable to connect. Could you please guide me on this?

And also, if there are any additional packages needed, please let us know.

helensilva14:
Related to #1242 

/gcbrun

### DIFF
--- a/data_validation/secret_manager.py
+++ b/data_validation/secret_manager.py
@@ -12,6 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
+
+from google.api_core import exceptions
+
 
 class SecretManagerBuilder:
     def build(self, client_type):
@@ -37,10 +41,11 @@ class GCPSecretManager:
         # Create the Secret Manager client.
         self.client = secretmanager.SecretManagerServiceClient()
 
-    def maybe_secret(self, project_id, secret_id, version_id="latest"):
-        """
-        Get information about the given secret.
-        :return String value with the secret value or the secret id if the secret value if not exists
+    def maybe_secret(self, project_id: str, secret_id: str, version_id="latest") -> str:
+        """Get information about the given secret.
+
+        Returns:
+            str: Secret value as a string or the secret id if the secret does not exist.
         """
         try:
             # Build the resource name of the secret.
@@ -50,5 +55,6 @@ class GCPSecretManager:
             # Return the decoded payload.
             payload = response.payload.data.decode("UTF-8")
             return payload
-        except BaseException:
+        except exceptions.NotFound:
+            # The secret does not exist and therefore we can assume it is just a token to be returned.
             return secret_id
--- a/data_validation/secret_manager.py
+++ b/data_validation/secret_manager.py
@@ -12,8 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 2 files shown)


data_validation/secret_manager.py
```diff
@@ -50,5 +53,6 @@ def maybe_secret(self, project_id, secret_id, version_id="latest"):
             # Return the decoded payload.
             payload = response.payload.data.decode("UTF-8")
             return payload
-        except BaseException:
+        except exceptions.NotFound:
```
nj1973, line 56: The main reason I didn't create a PR previously is the risk of this line.

I think it is better to tell the user if they have a real problem and therefore I think this change is good. My only concern is that if we have customers who have incorrectly included the secret project id in their connection files then it will currently be silently failing. But after this change the exception will bubble to the top.

On balance I still think this code is better than before. But we should expect someone somewhere to see an exception they did not know they were hitting.

nj1973 (approved):
LGTM on the basis that we expect someone somewhere to see an exception they didn't know they had.

data_validation/secret_manager.py
helensilva14: Agreed! I wonder if it would be the case of marking it as a breaking change? But since it's something not in the main and critical path, maybe it's not worth it...

nj1973: Yeah, I think it's not worth marking as a breaking change. Just something for us to be aware of.

sundar-mudupalli-work:
Helen,

Looking at [Exceptions thrown with this call](https://cloud.google.com/python/docs/reference/secretmanager/1.0.2/google.cloud.secretmanager_v1.SecretManagerServiceClient#google_cloud_secretmanager_v1_SecretManagerServiceClient_access_secret_version), it apperas that all you will know is if the "If the request failed for any reason.", not that the particular secret is not there. This is deliberate - so that you cannot by repeatedly querying the secret manager catalog all the secrets it holds. I am not sure how this code works. Can this be tested ?

Sundar Mudupalli

nj1973:
> This is deliberate - so that you cannot by repeatedly querying the secret manager catalog all the secrets it holds.

Where did you see this documented. When I manually tested the code a month or so ago we got a `NotFound` exception for secrets that did not exist.

Prior to these changes misconfigurations by the user were silently ignored which caused users to blame DVT rather than investigate real issues they may have.

> Can this be tested ?

I presume this means "can we add automated tests?". You are right it would be nice to have integration tests for this to ensure API changes do not break our code. I couldn't think of a nice way to do this at the time but perhaps we could do something if I tried a little harder. We've already merged this PR though so I think it needs to be a new issue.

PR #1286, opened 2024-10-01
chore: Add integration tests for DB2 and missing changes for setup.py usage

### LINKED ISSUE #649: Add integration tests for DB2
Implement a test suite for DB2 in the same way we've for our other supported connections: https://github.com/GoogleCloudPlatform/professional-services-data-validator/tree/develop/tests/system/data_sources

helensilva14:
This PR merges the following changes:

- Adds changes regarding setup.py usage across files that ended up missing in PR #1156 that was created to revert PR https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1150
   - To see only these files, check commit 747efe2bbf60feab01777cc567a2d46e95ab19d6
   
- Adds a new Cloud Build step to run the Nox session specific for DB2 integration tests
- Adds `tests/system/data_sources/test_db2.py` file with 9 tests for schema, column, row and custom-query validations
   - Tests for `generate-table-partitions` still pending, not prioritized for now but issue #1300 created for it
   - Bug for column validations: #1295 (TODO comments added where I think this should be addressed)
   - Bug for row validations: #1296 (TODO comments added where I think this should be addressed)
- Updates `tests/resources/db2_test_tables.sql` to reflect the current state of our test instance
   - INSERT statement for table [db2inst1.dvt_large_decimals](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/a98b7b556a7557f4abe3fc9ba3ecac2b30863ad6/tests/resources/db2_test_tables.sql#L79) was commented because I was having problems with the values' precision and scales
- Updates all files in `tests/system/data_sources` to use the common functions


tests/resources/db2_test_tables.sql
```diff
@@ -32,42 +32,85 @@ CREATE TABLE dvt_core_types
 ,   col_string      VARCHAR(32000)
 ,   col_date        DATE
 ,   col_datetime    TIMESTAMP(3)
-,   col_tstz        TIMESTAMP(3)
+,   col_tstz        TIMESTAMP(3) -- Unable to create col_tstz with time zone on our test database therefore test data is adjusted.
```
helensilva14, line 35:
```
[db2inst1@db2server ~]$ db2 "CREATE TABLE db2inst1.dvt_date_timestamp(id INTEGER NOT NULL PRIMARY KEY, col_date DATE, col_datetime TIMESTAMP(3), col_tstz TIMESTAMP(3) WITH TIME ZONE);"
DB21034E  The command was processed as an SQL statement because it was not a 
valid Command Line Processor command.  During SQL processing it returned:
SQL0104N  An unexpected token "TIME" was found following "tz TIMESTAMP(3) 
WITH".  Expected tokens may include:  "REFERENCES".  SQLSTATE=42601
```
Adding here the output we get when trying to create a column with type TIMESTAMP WITH TIME ZONE since it seems it's not supported by DB2 LUW (image below), only by DB2 z/OS.

image

References: 

- https://infocenter.sybase.com/help/index.jsp?topic=/com.sybase.help.sqlanywhere.12.0.1/mlserver/ibm-datatypes-ml-ref.html,
- https://www.ibmmainframeforum.com/db2/topic12878.html#:~:text=I%20searched%20a%20bit%20more%2C%20it%20seems%20that%20TIMESTAMP%20WITH%20TIME%20ZONE%20is%20supported%20by%20zOS%20version%20only.
- https://techdocs.broadcom.com/us/en/ca-mainframe-software/database-management/ca-fast-unload-for-db2-for-z-os/20-0/reference/keyword-descriptions/timestamp-with-time-zone-specify-the-data-format.html

tests/system/data_sources/common_functions.py
sundar-mudupalli-work, line 1: More pythonic way of doing the same thing
```
build_arg_list = [
        "validate",
        "custom-query",
        f"{validation_type}",
        "-sc=mock-conn",
        f"-tc={tc}",
        f"--source-query={source_query}",
        f"--target-query={target_query}",
        "--filter-status=fail",
            ...
        f"--count={count_cols}" if count_cols else None,
        ]
cli_arg_list = [arg for arg in build_arg_list if arg]

sundar-mudupalli-work (changes requested):
The updates to the tests and the introduction of the db2 tests look good to me.
The changes to RELEASING.md, cli/release_latest.sh, cloudbuild.yaml, noxfile.py, samples/docker/build_docker.sh and tests/local_check.sh - I am not sure about this.
Should we get input from Neha or Neil ?

Thanks.

Sundar Mudupalli

nj1973:
> The changes to RELEASING.md, cli/release_latest.sh, cloudbuild.yaml, noxfile.py, samples/docker/build_docker.sh and tests/local_check.sh - I am not sure about this.
> Should we get input from Neha or Neil ?

My only comment is on `cloudbuild.yaml` and the values we use for `waitFor`, this extends beyond the new DB2 tests.

We can run the DB2 tests in parallel to other integration tests by using `waitFor: ['-']`. I think we should have also included that for Teradata, Snowflake and Hive when we added those in the past, unless we intentionally wanted them to run serially after other tests are complete, I can't think why.

It also looks like the `waitFor` lines for MySQL and Oracle need to be switched but that is not the fault of this PR.

It feels like we should also review `PYTHON_VERSIONS = ["3.8", "3.9", "3.10"]` in `noxfile.py` but that is **not** something for this PR.

Edit:

I've added https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1303 to cover my suggestions above.

LGTM.

tests/system/data_sources/common_functions.py
nj1973: I've changed most of these but left a couple that I thought were more readable as they are.

nj1973 (approved):
LGTM

helensilva14:
> The updates to the tests and the introduction of the db2 tests look good to me. The changes to RELEASING.md, cli/release_latest.sh, cloudbuild.yaml, noxfile.py, samples/docker/build_docker.sh and tests/local_check.sh - I am not sure about this. Should we get input from Neha or Neil ?
> 
> Thanks.
> 
> Sundar Mudupalli

Those changes are just putting back the original values they had previously, but they are just documentation or sample files and don't affect at all our features and such. You can see this history in the two past PRs I mentioned in the description. Anyways @nehanene15 also has the context about it.

tests/system/data_sources/common_functions.py
helensilva14: Agreed, thank you very much @nj1973 for doing the changes for me! I'm currently only able to check Github using my phone because I'm traveling.

helensilva14:
Even though Neil already gave the approval, apparently @sundar-mudupalli-work will need to provide an approval too since his review was the "request changes" type

sundar-mudupalli-work (approved):

PR #1289, opened 2024-10-02
feat: Partition Custom Queries

### LINKED ISSUE #1145: generate-table-partitions does not support custom query row validation
Hi,

generate-table-partitions is helpful in scaling row validations by breaking up large row validations into smaller chunks. Smaller chunks can be validated without running out of memory. Smaller chunks can also be validated in parallel speeding up row validations. Generate-table-partitions does not support custom query row validations. The problem is that the existing logic uses the tables and primary keys to generate where clauses to partition the table.

There are two options:
* require the user to provide a -tbls argument so that generate table partitions can generate the where clauses based on the underlying tables. This may not work in all cases since the underlying tables may have fewer or additional rows than in the custom query.
* partition the custom query itself - that is, make the custom query a subquery - extract the primary key(s) and generate the where clauses to partition the custom query.

Sundar Mudupalli

### LINKED ISSUE #1282: Validate column command accepts --use-random-row --random-row-batch-size parameters
Hi,

`--use-random-row` `--random-row-batch-size` are not valid parameters for validate column and should be removed.

`data-validation validate column -h` outputs the following (some text deleted):
```
data-validation validate column
       [-h] [--count COUNT] [--sum SUM] [--avg AVG] [--min MIN] [--max MAX] [--bit_xor BIT_XOR] [--std STD]
       [--grouped-columns GROUPED_COLUMNS] [--exclude-columns] [--threshold THRESHOLD] [--filters FILTERS] [--use-random-row]
       [--random-row-batch-size RANDOM_ROW_BATCH_SIZE] [--wildcard-include-string-len] [--wildcard-include-timestamp] [--cast-to-bigint]
       --tables-list TABLES_LIST --source-conn SOURCE_CONN --target-conn TARGET_CONN [--bq-result-handler BQ_RESULT_HANDLER] [--labels LABELS]
       [--service-account SERVICE_ACCOUNT] [--config-file CONFIG_FILE] [--config-file-json CONFIG_FILE_JSON] [--format FORMAT]
       [--filter-status FILTER_STATUS]

options:
  -h, --help            show this help message and exit

optional arguments:
  --count COUNT, -count COUNT
                        Comma separated list of columns for count 'col_a,col_b' or * for all columns
  --sum SUM, -sum SUM   Comma separated list of columns for sum 'col_a,col_b' or * for all columns
  --avg AVG, -avg AVG   Comma separated list of columns for avg 'col_a,col_b' or * for all columns
  --min MIN, -min MIN   Comma separated list of columns for min 'col_a,col_b' or * for all columns
  --max MAX, -max MAX   Comma separated list of columns for max 'col_a,col_b' or * for all columns
  --bit_xor BIT_XOR, -bit_xor BIT_XOR
                        Comma separated list of columns for hashing a concatenate 'col_a,col_b' or * for all columns
  --std STD, -std STD   Comma separated list of columns for standard deviation 'col_a,col_b' or * for all columns
  --grouped-columns GROUPED_COLUMNS, -gc GROUPED_COLUMNS
                        Comma separated list of columns to use in GroupBy 'col_a,col_b'
  --exclude-columns, -ec
                        Flag to indicate the list of columns should be excluded from validation and not included.
  --threshold THRESHOLD, -th THRESHOLD
                        Float max threshold for percent difference
  --filters FILTERS, -filters FILTERS
                        Filters in the format source_filter:target_filter
  --use-random-row, -rr
                        Finds a set of random rows of the first primary key supplied.
  --random-row-batch-size RANDOM_ROW_BATCH_SIZE, -rbs RANDOM_ROW_BATCH_SIZE
                        Row batch size used for random row filters (default 10,000).
  --wildcard-include-string-len, -wis
                        Include string fields for wildcard aggregations.
  --wildcard-include-timestamp, -wit
                        Include timestamp/date fields for wildcard aggregations.
  --cast-to-bigint, -ctb
                        Cast any int32 fields to int64 for large aggregations.
  --bq-result-handler BQ_RESULT_HANDLER, -bqrh BQ_RESULT_HANDLER
                        BigQuery result handler config details
  --labels LABELS, -l LABELS
                        Key value pair labels for validation run
  --service-account SERVICE_ACCOUNT, -sa SERVICE_ACCOUNT
                        Path to SA key file for result handler output
  --config-file CONFIG_FILE, -c CONFIG_FILE
                        Store the validation config in the YAML File Path specified
  --config-file-json CONFIG_FILE_JSON, -cj CONFIG_FILE_JSON
                        Store the validation config in the JSON File Path specified to be used for application use cases
  --format FORMAT, -fmt FORMAT
                        Set the format for printing command output, Supported formats are (text, csv, json, table). Defaults to table
  --filter-status FILTER_STATUS, -fs FILTER_STATUS
                        Comma separated list of statuses to filter the validation results. Supported statuses are (success, fail). If no list
                        is provided, all statuses are returned

required arguments:
  --tables-list TABLES_LIST, -tbls TABLES_LIST
                        Comma separated tables list in the form 'schema.table=target_schema.target_table'. Or shorthand schema.* for all
                        tables.
  --source-conn SOURCE_CONN, -sc SOURCE_CONN
                        Source connection name
  --target-conn TARGET_CONN, -tc TARGET_CONN
                        Target connection name
```
Sundar Mudupalli

### LINKED ISSUE #1283: validate custom-query column does not accept `--grouped-columns` parameter.
Hi,

The command `validate column` accepts `--grouped-columns` parameter and the `validate custom-query column` does not. It should accept the `--grouped-columns` parameter. The help message for `validate custom-query column` shows the following (some text deleted):
```
data-validation validate custom-query column
       [-h] [--count COUNT] [--sum SUM] [--avg AVG] [--min MIN] [--max MAX] [--bit_xor BIT_XOR] [--std STD] [--exclude-columns]
       [--wildcard-include-string-len] [--wildcard-include-timestamp] [--cast-to-bigint] [--filters FILTERS] [--threshold THRESHOLD]
       [--use-random-row] [--random-row-batch-size RANDOM_ROW_BATCH_SIZE]
       (--source-query-file SOURCE_QUERY_FILE | --source-query SOURCE_QUERY)
       (--target-query-file TARGET_QUERY_FILE | --target-query TARGET_QUERY) --source-conn SOURCE_CONN --target-conn TARGET_CONN
       [--bq-result-handler BQ_RESULT_HANDLER] [--labels LABELS] [--service-account SERVICE_ACCOUNT] [--config-file CONFIG_FILE]
       [--config-file-json CONFIG_FILE_JSON] [--format FORMAT] [--filter-status FILTER_STATUS]

options:
  -h, --help            show this help message and exit

optional arguments:
  --count COUNT, -count COUNT
                        Comma separated list of columns for count 'col_a,col_b' or * for all columns
  --sum SUM, -sum SUM   Comma separated list of columns for sum 'col_a,col_b' or * for all columns
  --avg AVG, -avg AVG   Comma separated list of columns for avg 'col_a,col_b' or * for all columns
  --min MIN, -min MIN   Comma separated list of columns for min 'col_a,col_b' or * for all columns
  --max MAX, -max MAX   Comma separated list of columns for max 'col_a,col_b' or * for all columns
  --bit_xor BIT_XOR, -bit_xor BIT_XOR
                        Comma separated list of columns for hashing a concatenate 'col_a,col_b' or * for all columns
  --std STD, -std STD   Comma separated list of columns for standard deviation 'col_a,col_b' or * for all columns
  --exclude-columns, -ec
                        Flag to indicate the list of columns should be excluded from validation and not included.
  --wildcard-include-string-len, -wis
                        Include string fields for wildcard aggregations.
  --wildcard-include-timestamp, -wit
                        Include timestamp/date fields for wildcard aggregations.
  --cast-to-bigint, -ctb
                        Cast any int32 fields to int64 for large aggregations.
  --filters FILTERS, -filters FILTERS
                        Filters in the format source_filter:target_filter
  --threshold THRESHOLD, -th THRESHOLD
                        Float max threshold for percent difference
  --use-random-row, -rr
                        Finds a set of random rows of the first primary key supplied.
  --random-row-batch-size RANDOM_ROW_BATCH_SIZE, -rbs RANDOM_ROW_BATCH_SIZE
                        Row batch size used for random row filters (default 10,000).
  --bq-result-handler BQ_RESULT_HANDLER, -bqrh BQ_RESULT_HANDLER
                        BigQuery result handler config details
  --labels LABELS, -l LABELS
                        Key value pair labels for validation run
  --service-account SERVICE_ACCOUNT, -sa SERVICE_ACCOUNT
                        Path to SA key file for result handler output
  --config-file CONFIG_FILE, -c CONFIG_FILE
                        Store the validation config in the YAML File Path specified
  --config-file-json CONFIG_FILE_JSON, -cj CONFIG_FILE_JSON
                        Store the validation config in the JSON File Path specified to be used for application use cases
  --format FORMAT, -fmt FORMAT
                        Set the format for printing command output, Supported formats are (text, csv, json, table). Defaults to table
  --filter-status FILTER_STATUS, -fs FILTER_STATUS
                        Comma separated list of statuses to filter the validation results. Supported statuses are (success, fail). If no list
                        is provided, all statuses are returned

required arguments:
  --source-query-file SOURCE_QUERY_FILE, -sqf SOURCE_QUERY_FILE
                        File containing the source sql query
  --source-query SOURCE_QUERY, -sq SOURCE_QUERY
                        Source sql query
  --target-query-file TARGET_QUERY_FILE, -tqf TARGET_QUERY_FILE
                        File containing the target sql query
  --target-query TARGET_QUERY, -tq TARGET_QUERY
                        Target sql query
  --source-conn SOURCE_CONN, -sc SOURCE_CONN
                        Source connection name
  --target-conn TARGET_CONN, -tc TARGET_CONN
                        Target connection name
```

sundar-mudupalli-work:
Hi,

When validating rows of large datasets, validations can run out of memory. Large datasets can be validating by partitioning them into smaller chunks. This feat extends partitioning to queries. 

For Row and Column validations, DVT used a separate command `validate custom-query row` and `validate custom-query column` when validations are required on queries instead of table. This PR integrates the `--source-query` and `--target-query` into the existing `generate-table-partitions` command. A user may provide either `--tables-list` or `--source-query` and `--target-query` (or the corresponding file options), but not both. Due to the limitations of `argparse`, the check to see which option was chosen (tables or query) is done after parsing. This leads to less duplicated code.

Closes #1145.

This also includes code cleanup and enhancements including:

1. Closes #1282 and #1283 
2. Removes duplicate code from `build_conf_from_args` in `data_validation/__main__.py`
3. Moves argument parsing/validation and default functions (for `filters` and `threshold`) as parameters to `add_argument` 
4. Moves the testing of yaml file creation and storage to unit tests 

Thanks.

Sundar Mudupalli


data_validation/__main__.py
```diff
@@ -225,6 +225,8 @@ def build_config_from_args(args: Namespace, config_manager: ConfigManager):
         config_manager (ConfigManager): Validation config manager instance.
     """
 
+    # If configuring generate-table-partitions, validation_type will be set to underlying validation either Row
+    # or custom query and, if custom query - args.custom_query_type will also be set.
```
helensilva14: This statement would be better placed as part of the function's documentation, since it's not a direct comment for a code line or block

data_validation/__main__.py
```diff
@@ -234,9 +236,10 @@ def build_config_from_args(args: Namespace, config_manager: ConfigManager):
             )
         config_manager.append_allow_list(args.allow_list, args.allow_list_file)
 
-    # Append CUSTOM_QUERY configs
+    # Append configs specific to CUSTOM_QUERY (i.e. query strings or strings from files)
     if config_manager.validation_type == consts.CUSTOM_QUERY:
-        config_manager.append_custom_query_type(args.custom_query_type)
+        if config_manager.validation_type == consts.CUSTOM_QUERY:
```
helensilva14, line 241:
```suggestion
```
It looks this additional conditional is not actually needed

data_validation/__main__.py
```diff
@@ -234,9 +236,10 @@ def build_config_from_args(args: Namespace, config_manager: ConfigManager):
...
+            config_manager.append_custom_query_type(args.custom_query_type)
```
helensilva14: As of now this condition is right after an exact same one, this can be misleading and maybe be seen as an error/duplicate line. At least, it was my first thought before I read the entire function and really understood its actual purpose. 

I think it's better to place it at the end of the function.

data_validation/cli_tools.py
```diff
@@ -242,6 +277,46 @@ def _configure_partition_parser(subparsers):
         help="Number of partitions into which the table should be split",
         type=_check_positive,
     )
+    # User can provide tables or custom queries, but not both
+    # However, Argprase does not support adding an argument_group to an argument_group or adding a
```
helensilva14:
```suggestion
    # However, Argparse does not support adding an argument_group to an argument_group or adding a
```

data_validation/cli_tools.py
```diff
@@ -1362,16 +1424,10 @@ def cols_from_arg(concat_arg: str, client, table_obj: dict, query_str: str) -> l
...
-    use_random_rows = None
-    random_row_batch_size = None
-    if (
-        args.command != "generate-table-partitions"
-        and config_type != consts.SCHEMA_VALIDATION
-    ):
-        use_random_rows = getattr(args, "use_random_row", False)
-        random_row_batch_size = getattr(args, "random_row_batch_size", None)
+    # Get random row arguments. In row validations these attributes will be present,
+    # otherwise not present.
```
helensilva14:
```suggestion
    # Get random row arguments. Only in row validations these attributes can be present.
```

README.md
```diff
@@ -262,6 +273,7 @@ data-validation
                         Directory Path to store YAML Config Files
                         GCS: Provide a full gs:// path of the target directory. Eg: `gs://<BUCKET>/partitions_dir`
                         Local: Provide a relative path of the target directory. Eg: `partitions_dir`
+                        If invoked with -tbls parameter, the validations are stored in a directory named <schema>.<table>, otherwise the directory is named `custom.<random_prefix>`
```
helensilva14:
```suggestion
                        If invoked with -tbls parameter, the validations are stored in a directory named <schema>.<table>, otherwise the directory is named `custom.<random_string>`
```

tests/system/data_sources/common_functions.py
```diff
@@ -346,17 +342,56 @@ def generate_partitions_test(
...
-    # 2 files were created with upto 5 validations in each file
-    assert len(yaml_configs_list[0]["yaml_files"]) == 2
-    # 5 validations in the first file
-    assert len(yaml_configs_list[0]["yaml_files"][0]["yaml_config"]["validations"]) == 5
-    # 4 validations in the second file
-    assert len(yaml_configs_list[0]["yaml_files"][1]["yaml_config"]["validations"]) == 4
+
+def partition_query_test(
+    expected_filter: str,
+    tmp_path,
```
helensilva14: What is the expected data type? 🤔

tests/system/data_sources/common_functions.py
```diff
@@ -346,17 +342,56 @@ def generate_partitions_test(
...
+    Teradata has a different syntax for inequality and Postgres has different column names/types for primary keys.
+    The unit tests in tests/unit/test_partition_builder.py check if the filters are split into configs and
+    stored in the filesystem correctly.
+    """
+    tables_list = tables.split("=")
+    source_table_name = tables_list[0]
+    target_table_name = tables_list[1] if len(tables_list) == 2 else tables_list[0]
+    target_query = f"select * from {target_table_name}"
+    source_query = f"select * from {source_table_name}"
+    source_query_file = tmp_path / "source_query.sql"
```
helensilva14, line 371: I asked about the data type above because it's not clear to me what's happening at this line. 

I tested on my own and I got:
```
>>> tmp_path = "/teste"
>>> print(tmp_path / "source_query.sql")
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for /: 'str' and 'str'
```

tests/system/data_sources/test_bigquery.py
```diff
@@ -1154,9 +1155,10 @@ def test_custom_query():
     "data_validation.state_manager.StateManager.get_connection_config",
     return_value=BQ_CONN,
 )
-def test_bigquery_generate_table_partitions(mock_conn):
-    """Test generate table partitions on BigQuery"""
-    generate_partitions_test(EXPECTED_PARTITION_FILTER)
+def test_generate_partitions(mock_conn, tmp_path):
```
helensilva14: Where is `tmp_path` coming from? I looked at all variables available at the top and it's not there, I'm super curious here haha same question applies to the other systems too

tests/system/data_sources/test_hive.py
```diff
@@ -128,9 +131,10 @@ def test_schema_validation_core_types():
     "data_validation.state_manager.StateManager.get_connection_config",
     new=mock_get_connection_config,
 )
-def test_hive_generate_table_partitions():
+def test_generate_partitions(tmp_path):
     """Test generate table partitions on Hive"""
```
helensilva14:
```suggestion
    """Test generate partitions on Hive, first on table, then on custom query"""
```

tests/system/data_sources/test_mysql.py
```diff
@@ -113,9 +116,10 @@ def test_mysql_count_invalid_host():
     "data_validation.state_manager.StateManager.get_connection_config",
     new=mock_get_connection_config,
 )
-def test_mysql_generate_table_partitions():
-    """Test generate table partitions on mysql"""
-    generate_partitions_test(EXPECTED_PARTITION_FILTER)
+def test_generate_partitions(tmp_path):
+    """Test generate partitions first table, then query on mysql"""
```
helensilva14:
```suggestion
    """Test generate partitions on MySQL, first on table, then on custom query"""
```

tests/system/data_sources/test_oracle.py
```diff
@@ -139,14 +142,40 @@ def mock_get_connection_config(*args):
...
 
 @mock.patch(
     "data_validation.state_manager.StateManager.get_connection_config",
     new=mock_get_connection_config,
 )
-def test_oracle_generate_table_partitions():
-    """Test generate table partitions on Oracle"""
-    generate_partitions_test(EXPECTED_PARTITION_FILTER)
+def test_generate_partitions(tmp_path):
+    """Test generate table partitions first on table, then custom query on Oracle"""
```
helensilva14:
```suggestion
    """Test generate partitions on Oracle, first on table, then on custom query"""
```

tests/system/data_sources/test_sql_server.py
```diff
@@ -238,11 +238,14 @@ def mock_get_connection_config(*args):
     "data_validation.state_manager.StateManager.get_connection_config",
     new=mock_get_connection_config,
 )
-def test_sqlserver_generate_table_partitions(cloud_sql):
-    """Test generate table partitions on sqlserver"""
-    generate_partitions_test(
+def test_generate_partitions(cloud_sql, tmp_path):
+    """Test generate table partitions first table then custom query on sqlserver"""
```
helensilva14:
```suggestion
    """Test generate partitions on SQL Server, first on table, then on custom query"""
```

helensilva14 (commented):
Finished my review, good job @sundar-mudupalli-work! Please take a look at my comments and suggestions, thanks!

data_validation/__main__.py
nehanene15: +1, I think we should move this comment to a more relevant spot or delete it. All custom-query validations should populate the `args.custom_query_type` so not sure it's worth calling out.

data_validation/__main__.py
nehanene15: +1 we don't need the outer conditional here checking for the same condition.

data_validation/__main__.py
```diff
@@ -252,53 +255,29 @@ def build_config_from_args(args: Namespace, config_manager: ConfigManager):
...
-            config_manager.append_primary_keys(
-                config_manager.build_column_configs(primary_keys)
-            )
-
-    # Append COLUMN_VALIDATION configs
-    if config_manager.validation_type == consts.COLUMN_VALIDATION:
+    # Append COLUMN_VALIDATION configs, including custom-query column validation
+    if (
+        config_manager.validation_type == consts.COLUMN_VALIDATION
+        or config_manager.validation_type == consts.CUSTOM_QUERY
```
nehanene15, line 264: Shouldn't this be grouping the last two statements? Right now, it might require args.custom_query_type='column' since it's in the 'and'

```
   if (
        config_manager.validation_type == consts.COLUMN_VALIDATION
        or (config_manager.validation_type == consts.CUSTOM_QUERY
        and args.custom_query_type == consts.COLUMN_VALIDATION.lower())
    ):
```

data_validation/__main__.py
```diff
@@ -252,53 +255,29 @@ def build_config_from_args(args: Namespace, config_manager: ConfigManager):
...
+        and args.custom_query_type == consts.COLUMN_VALIDATION.lower()
+    ):
         config_manager.append_aggregates(get_aggregate_config(args, config_manager))
-        if args.grouped_columns is not None:
+        if (
+            config_manager.validation_type == consts.COLUMN_VALIDATION
+            and args.grouped_columns
```
nehanene15: You can also write `if config_manager.validation_type == consts.COLUMN_VALIDATION and args.grouped_columns` instead of writing 'is not None' for readability

data_validation/__main__.py
```diff
@@ -252,53 +255,29 @@ def build_config_from_args(args: Namespace, config_manager: ConfigManager):
...
             config_manager.append_query_groups(
                 config_manager.build_column_configs(grouped_columns)
             )
 
-    # Append ROW_VALIDATION configs
-    if config_manager.validation_type == consts.ROW_VALIDATION:
+    # Append ROW_VALIDATION configs, including custom-query row validation
+    if (
+        config_manager.validation_type == consts.ROW_VALIDATION
+        or config_manager.validation_type == consts.CUSTOM_QUERY
```
nehanene15, line 280: Same as above, shouldn't the last two custom query statements be inside parentheses?

data_validation/__main__.py
```diff
@@ -252,53 +255,29 @@ def build_config_from_args(args: Namespace, config_manager: ConfigManager):
...
+        and args.custom_query_type == consts.ROW_VALIDATION.lower()
+    ):
         # Append calculated fields: --hash/--concat
         config_manager.append_calculated_fields(
```
nehanene15, line 284: Are there any effects of appending the calculated fields _before_ comparison fields? Before this PR, the order was reversed for custom query.

data_validation/config_manager.py
```diff
@@ -386,7 +390,7 @@ def get_source_ibis_table_from_query(self):
     def get_source_ibis_calculated_table(self, depth=None):
         """Return mutated IbisTable from source
         n: Int the depth of subquery requested"""
-        if self.validation_type == consts.CUSTOM_QUERY:
+        if self.source_table is None:  # either custom query or partition custom query
```
nehanene15: Why do we have to change this? Won't generate-partitions custom query also set the validation_type to 'CUSTOM_QUERY'?

data_validation/validation_builder.py
```diff
@@ -384,11 +384,10 @@ def get_source_query(self):
             "source_query": self.config_manager.source_query,
         }
 
-        if self.validation_type == consts.CUSTOM_QUERY:
-            table = self.config_manager.get_source_ibis_table_from_query()
-        else:
+        if self.config_manager.source_table:
```
nehanene15: Same as above - why do we need to check if source_table exists rather than keeping `if self.validation_type == consts.CUSTOM_QUERY:`

nehanene15 (commented):
Left a couple of comments. Overall great optimizations and amazing job!!

data_validation/__main__.py
sundar-mudupalli-work: Fixed

data_validation/__main__.py
sundar-mudupalli-work: Fixed

data_validation/__main__.py
sundar-mudupalli-work: That grouping is already there since and has higher precedence than or.

data_validation/__main__.py
sundar-mudupalli-work: and has higher precedence than or - so it is already there.

data_validation/__main__.py
sundar-mudupalli-work: I saw that as well, but forgot to double check it. Now I have. Calculated_fields (hash or concat) and comparison_fields are mutually exclusive - in the argparse, so only one of these if statements is ever triggered.

data_validation/config_manager.py
sundar-mudupalli-work: Good catch. I had made this change before I decide to code the validation_type. Now that I made that change, this code can revert back to the more elegant way of doing things. Thank you.

data_validation/validation_builder.py
sundar-mudupalli-work: Yes, you are correct. It turns out this code is not called by generate-partitions at all, it is only called when doing a dry-run or executing the query, so I need not have made this change.

data_validation/__main__.py
sundar-mudupalli-work: Done

data_validation/validation_builder.py
```diff
@@ -406,10 +405,10 @@ def get_target_query(self):
             "target_query": self.config_manager.target_query,
         }
 
-        if self.validation_type == consts.CUSTOM_QUERY:
```
nehanene15, line 409: Here too, we can keep the `validation_type == consts.CUSTOM_QUERY`

data_validation/config_manager.py
```diff
@@ -416,7 +422,7 @@ def get_target_ibis_table_from_query(self):
     def get_target_ibis_calculated_table(self, depth=None):
         """Return mutated IbisTable from target
         n: Int the depth of subquery requested"""
-        if self.validation_type == consts.CUSTOM_QUERY:
+        if self.target_table is None:  # either custom query or partition custom query
```
nehanene15: Can't this also remain `if self.validation_type == consts.CUSTOM_QUERY`?

data_validation/config_manager.py
```diff
@@ -386,7 +390,9 @@ def get_source_ibis_table_from_query(self):
     def get_source_ibis_calculated_table(self, depth=None):
         """Return mutated IbisTable from source
         n: Int the depth of subquery requested"""
-        if self.validation_type == consts.CUSTOM_QUERY:
+        if (
```
nehanene15: nit: doesn't have to be in parenthesis

nehanene15 (approved):
3 small updates that might have been missed. Will approve so you can merge afterwards.

data_validation/config_manager.py
sundar-mudupalli-work: That was due to the comment there black was inserting the parantheses. The comment is no longer necessary since the code itself says that. Removing the comment - removed the ()

data_validation/validation_builder.py
sundar-mudupalli-work: Done

data_validation/config_manager.py
sundar-mudupalli-work: It should not be changed there, but later on the in the code. I still found the code very confusing therefore opened an [issue](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1294)

PR #1290, opened 2024-10-04
fix: Combiner did not take boolean datatypes into account

### LINKED ISSUE #1262: Error while validating Postgres BOOLEAN datatype with Teradata CHAR datatype in row validation
Customer facing another issue while validating Postgres with Teradata. They've a column,ISCURRENTFORTYPE, defined boolean in Postgres and CHAR(1) in Teradata. When they're validating this column using comp field comparison using below command, it's throwing error : 

--source-conn pg_conn \
--target-conn td_conn \
--tables-list public.eodassetlocation=Z_OD_ALF_LIVE_M_Q.Z0325_ODSASSETLOC_CURR \
-tsp \
--primary-keys ID \
--comparison-fields ID,ASSETID,ISCURRENTFORTYPE \
--bq-result-handler prj-dfdl-528-fcmis-d-528.bq_dvt_validation_results_raw.bq_postgres_td_dvt_validation_results \
-l Tables=public.eodassetlocation:Z_OD_ALF_LIVE_M_Q.Z0325_ODSASSETLOC_CURR,run_date=20240910-1

**Error :** 
ibis.common.exceptions.IbisTypeError: Given arguement with datatype boolean is not implicitly castable to string.

Later they created a view on TD to type cast column ISCURRENTFORTYPE to BYTEINT and used view in DVT validation. Got below error : 

**TypeError:** Arguments differences_target_value:int64 and differences_source_value:boolean are not comparable

nj1973:
Fix recently merged boolean implementation which was only working by accident. It worked if the boolean was on the target side and the pseudo boolean was on the source size. It failed with the opposite configuration.

### DIFF
--- a/data_validation/combiner.py
+++ b/data_validation/combiner.py
@@ -23,7 +23,7 @@ import functools
 import json
 import logging
 import ibis
-import ibis.expr.datatypes
+import ibis.expr.datatypes as dt
 
 from data_validation import consts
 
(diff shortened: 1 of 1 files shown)


helensilva14 (approved):
LGTM! That was a very fast bug fix, good job! 🚀

PR #1291, opened 2024-10-04
feat: Add support for -comp-fields=*, which also adds support for --exclude-columns

### LINKED ISSUE #1222: Comparison fields option does not take asterisk as a parameter
Comparison fields option does not take asterisk as a parameter, for example:
```
$ data-validation validate row -sc=ora_local -tc=pg_local -tbls=pso_data_validator.dvt_core_types --primary-keys=id --comparison-fields='*'
...
ibis.common.exceptions.IbisTypeError: Column '*' is not found in table. Existing columns: 'id', 'col_int8', 'col_int16', 'col_int32', 'col_int64', 'col_dec_20', 'col_dec_38', 'col_dec_10_2', 'col_float32', 'col_float64', 'col_varchar_30', 'col_char_2', 'col_string', 'col_date', 'col_datetime', 'col_tstz'.
```

If we decide this should support '*' then we need to review how the expansion works because the option does not accept the primary key therefore we would need to exclude any columns defined as primary keys:
```
$ data-validation validate row -sc=ora_local -tc=pg_local -tbls=pso_data_validator.dvt_core_types --primary-keys=id --comparison-fields='id,col_int8,col_int16'
...
    raise IntegrityError(f'Duplicate column name(s): {duplicate_names}')
ibis.common.exceptions.IntegrityError: Duplicate column name(s): ['id']
```

Finally we don't appear to have any integration tests for `--comparison-fields`.

nj1973:
This PR adds support for `-comp-fields=*`, which also adds support for `--exclude-columns`.

Tests added in Oracle and PostgreSQL test sets. We could (should?) add more but the asterisk change is not engine specific and tests are slow enough as it is.

### DIFF
--- a/data_validation/__main__.py
+++ b/data_validation/__main__.py
@@ -160,7 +160,7 @@ def get_aggregate_config(args, config_manager: ConfigManager):
     return aggregate_configs
 
 
-def get_calculated_config(args, config_manager: ConfigManager) -> List[dict]:
+def _get_calculated_config(args, config_manager: ConfigManager) -> List[dict]:
     """Return list of formatted calculated objects.
 
     Args:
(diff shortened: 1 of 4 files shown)


helensilva14 (commented):
LGTM, very nice! Just an observation that this PR will probably need to be updated if PR #1286 is merged before this one.

helensilva14 (approved):
Approving already since the main changes have been reviewed!

PR #1304, opened 2024-10-22
fix: Correctly cater for Oracle FLOAT in custom-query validations

### LINKED ISSUE #1302: Validate custom query row throws exception for Oracle FLOAT(126) column
Test table:
```
create table dvt_test.tab_float_126 (id number(5), col_flt_126 float(126));
```

Standard table validation is successful:
```
data-validation validate row --source-conn ora --target-conn ora \
--tables-list "dvt_test.tab_float_126" --primary-keys id --concat="*"
```

But `custom-query` validation throws an exception:
```
data-validation validate custom-query row --source-conn ora_local --target-conn ora_local \
-sq "select id,col_flt_126 from dvt_test.tab_float_126" \
-tq "select id,col_flt_126 from dvt_test.tab_float_126" \
-pk id --hash="*"
...
    raise ValueError('Decimal type scale cannot be negative')
ValueError: Decimal type scale cannot be negative
```

nj1973:
Fix two issues:

1) Use of Oracle FLOAT in a custom query validation
2) Use of PostgreSQL numeric in a custom query validation

Also adds a test covering both issues.

### DIFF
--- a/third_party/ibis/ibis_oracle/datatypes.py
+++ b/third_party/ibis/ibis_oracle/datatypes.py
@@ -50,8 +50,8 @@ def _get_type(col: _FieldDescription) -> dt.DataType:
         raise NotImplementedError(f"Oracle type {typename} is not supported")
 
     if typename == cx_Oracle.DB_TYPE_NUMBER:
-        if col[4] == 0 and col[5] == -127:
-            # This will occur if type is NUMBER with no precision/scale
+        if col[5] == -127:
+            # This will occur if type is NUMBER with no precision/scale or if type is FLOAT.
             typ = partial(typ)
         else:
             typ = partial(typ, precision=col[4], scale=col[5])
--- a/third_party/ibis/ibis_postgres/client.py
+++ b/third_party/ibis/ibis_postgres/client.py
@@ -99,7 +99,13 @@ def _get_type(typestr: str) -> dt.DataType:
     else:
         typestr_wo_length = typestr_wob
     typ = _type_mapping.get(typestr_wo_length)
-    if typ is not None:
+    # Added != "numeric" below for issue:
+    #   https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1302
+    # We do not want to map all numerics to decimal(None, None), we want "numeric" to be
+    # passed to the _parse_numeric() function.
+    # An alternative was to remove "numeric" from _type_mapping but that would be yet more monkey
+    # patching, at least this function is already patched.
+    if typ is not None and typestr_wo_length != "numeric":
         return dt.Array(typ) if is_array else typ
     return _parse_numeric(typestr)
 
--- a/tests/system/data_sources/test_oracle.py
+++ b/tests/system/data_sources/test_oracle.py
@@ -602,6 +602,7 @@ def test_custom_query_row_validation_oracle_to_postgres():
     )
     custom_query_validation_test(
         validation_type="row",
+        tc="pg-conn",
         source_query=f"select {hash_cols} from pso_data_validator.dvt_ora2pg_types",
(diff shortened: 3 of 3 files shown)


luispavaogoogle (approved):
LGTM, everything is clear and well commented

third_party/ibis/ibis_oracle/datatypes.py
```diff
@@ -50,8 +50,8 @@ def _get_type(col: _FieldDescription) -> dt.DataType:
         raise NotImplementedError(f"Oracle type {typename} is not supported")
 
     if typename == cx_Oracle.DB_TYPE_NUMBER:
-        if col[4] == 0 and col[5] == -127:
-            # This will occur if type is NUMBER with no precision/scale
+        if col[5] == -127:
```
sundar-mudupalli-work, line 53: I know this was pre-existing. Is it possible to point to some documentation about these values - ie. -127

third_party/ibis/ibis_postgres/client.py
```diff
@@ -99,7 +99,13 @@ def _get_type(typestr: str) -> dt.DataType:
...
         typestr_wo_length = typestr_wob
     typ = _type_mapping.get(typestr_wo_length)
-    if typ is not None:
+    # Added != "numeric" below for issue:
+    #   https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1302
+    # We do not want to map all numerics to decimal(None, None), we want "numeric" to be
+    # passed to the _parse_numeric() function.
+    # An alternative was to remove "numeric" from _type_mapping but that would be yet more monkey
+    # patching, at least this function is already patched.
+    if typ is not None and typestr_wo_length != "numeric":
```
sundar-mudupalli-work, line 108: Is there any Postgres doc we can point to regarding `typestr_wo_length` != "numeric"` or is this something from DVT/ibis

sundar-mudupalli-work (approved):
Neil,

This looks good to me. I made some comments on specific lines, but they are not showstoppers.

Thanks again for the great work.

third_party/ibis/ibis_oracle/datatypes.py
nj1973: I don't believe it is documented anywhere. Perhaps in an Oracle support note but, to my knowledge, nowhere publicly visible.

third_party/ibis/ibis_postgres/client.py
nj1973: The value comes from the PostgreSQL catalog and is referenced as a string in Ibis code:
```
_type_mapping = {
    "bigint": dt.int64,
    "boolean": dt.bool,
    "bytea": dt.binary,
    "character varying": dt.string,
    "character": dt.string,
    "character(1)": dt.string,
    "date": dt.date,
    "double precision": dt.float64,
    "geography": dt.geography,
    "geometry": dt.geometry,
    "inet": dt.inet,
    "integer": dt.int32,
    "interval": dt.interval,
    "json": dt.json,
    "jsonb": dt.json,
    "line": dt.linestring,
    "macaddr": dt.macaddr,
    "macaddr8": dt.macaddr,
    "numeric": dt.decimal,
    "point": dt.point,
    "polygon": dt.polygon,
    "real": dt.float32,
    "smallint": dt.int16,
    "text": dt.string,
    # NB: this isn't correct because we're losing the "with time zone"
    # information (ibis doesn't have time type that is time-zone aware), but we
    # try to do _something_ here instead of failing
    "time with time zone": dt.time,
    "time without time zone": dt.time,
    "timestamp with time zone": dt.Timestamp("UTC"),
    "timestamp without time zone": dt.timestamp,
    "uuid": dt.uuid,
}
```
I don't find any of this code in recent Ibis versions so this might all go away if/when we upgrade.

PR #1310, opened 2024-11-04
feat: user defined input `run_id` via config

nick-redfearn:
I would like to be able to define `run_id`s before I start individual row validations, so that I can better keep track of individual runs.

Before this change = The `run_id` is generated as part of `RunMetadata` object using `str(uuid.uuid4())`

After this change = The `run_id` can be input via config, in which case it is assigned as part of `data_validation.py`, otherwise if the input config value is None then `data_validation.py` will continue to use `str(uuid.uuid4())`

To reviewers...
Please note that the majority of my changes are to `test_data_validation.py` where Ive added extra assertions and cleaned up things up by using references to `consts` values rather than strings.


nick-redfearn:
@nj1973 - Would you mind explaining your 2 commits for my understanding please?

- https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1310/commits/55f5e645cd67e635d257c37ab72183e8f124aec0
- https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1310/commits/6857a27d8ac2f0074b42c317e2b588975c3446b6

nick-redfearn:
@nj1973 Im considering adding this to `cli_tools.py`, under `_add_common_arguments`
image

nj1973:
> @nj1973 - Would you mind explaining your 2 commits for my understanding please?
> 
> * [55f5e64](https://github.com/GoogleCloudPlatform/professional-services-data-validator/commit/55f5e645cd67e635d257c37ab72183e8f124aec0)
> * [6857a27](https://github.com/GoogleCloudPlatform/professional-services-data-validator/commit/6857a27d8ac2f0074b42c317e2b588975c3446b6)

Hi @nick-redfearn, happy to. In the BigQuery integration tests there are some tests which store validations in a YAML config file. They include an assertion that the number of lines in the YAML file match a constant. With the addition of the new run_id attribute the constants needed increasing by 1.

I'm not convinced that checking the number of lines in the config file is the best way for these tests to identify success but I don't want to dig into that right now.

You are unlikely to be able to run integration tests due to needing many different database types therefore you won't see knock on effects like this. I figured the simplest thing to do was to adjust the constants and update you once the tests were passing.

I'll review the PR properly later today.

nick-redfearn:
@nj1973 Thanks very much 🙇

nj1973:
> @nj1973 Im considering adding this to `cli_tools.py`, under `_add_common_arguments` 

Yep, I think that's a good idea.

nick-redfearn:
> Yep, I think that's a good idea.

@nj1973 Added in [this commit](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1310/commits/75def1e518da24e9530ca2ddcd283c315d30003e)

nj1973:
> > Yep, I think that's a good idea.
> 
> @nj1973 Added in [this commit](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1310/commits/75def1e518da24e9530ca2ddcd283c315d30003e)

Hi, the value from the command line option is not being passed through. You need to add some code to `get_pre_build_configs()` to pass the argument value through to the config manager. And perhaps also add a test to `test_config_manager.py` (similar to `test_get_threshold_property`).

data_validation/cli_tools.py
```diff
@@ -990,6 +990,12 @@ def _add_common_arguments(
         # TODO: update if we start to support other statuses
         help="Comma separated list of statuses to filter the validation results. Supported statuses are (success, fail). If no list is provided, all statuses are returned",
     )
+    optional_arguments.add_argument(
+        "--run-id",
+        "-rid",
+        default=None,
+        help="Set a string for the run_id, if None is input then an randomly generated UUID will be used, which is the default behaviour",
```
nj1973:
```suggestion
        help="Set a string for the run_id, if None is input then a randomly generated UUID will be used, which is the default behaviour",
```

nick-redfearn:
@nj1973 Ive added these commits...

- Cleanup in `cli_tools.py` to use `consts` values -> https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1310/commits/6908c7ee70d16eba956126a45611dc1772a5b694
- Adding to `get_pre_build_configs` -> https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1310/commits/6908c7ee70d16eba956126a45611dc1772a5b694
- Adding to `test_config_manager.py`, similar to `test_get_threshold_property` + formatting in `get_pre_build_configs` -> https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1310/commits/b6b08d4f940d809bd8963b8cd4d4dc32021f7050

nj1973 (approved):
LGTM

But my suggested earlier change ended up using my personal email address for the commit instead of my work one, normally this is no issue but for some reason it is preventing merge of the PR (sigh). The co-authored path must take a different path to a normal commit. I'll try and figure out how to proceed.

nj1973:
Thanks @nick-redfearn for your continued contributions, it's very much appreciated.

nick-redfearn:
@nj1973 Thanks a lot for your help, much appreciated 🙇

PR #1311, opened 2024-11-04
test: Update Cloud Build test sequencing

### LINKED ISSUE #1303: Refine test sequencing and versions
**Integration test sequence**
Currently Teradata, Snowflake, DB2 and Hive integration tests run sequentially after all other tests have completed. This is particularly painful because the Hive tests are the longest running tests.

We should try changing `cloudbuild.yaml` to include `waitFor: ['-']` so that they all run concurrently. Hopefully the higher concurrency doesn't cause any infrastructure issues. If it does then perhaps we can rearrange things slightly so Hive at least starts as soon as possible.

**MySQL wantFor incorrect**
MySQL uses the proxy to connect and should have the following `waitFor` line:
```
waitFor: ['proxy-install']
```

**PYTHON_VERSIONS**
We should review PYTHON_VERSIONS = ["3.8", "3.9", "3.10"] in `noxfile.py`.

I think we should add 3.11 and 3.12 to this list.

helensilva14:
Closes #1303 

/gcbrun


nj1973 (approved):
LGTM.

PR #1316, opened 2024-11-05
fix: Hardcode SQLAlchemy max_identifier_length=128 for Oracle connections

### LINKED ISSUE #1250: Oracle 11g row validation exception ORA-00904 invalid identifier
A customer is running a row validation against an Oracle 11g database, something like:
```
data-validation validate row -sc ora -tc pg \
-tbls dvt_test.tab_00904 \
-pk pkcol_1234567890123456789012 \
--concat='pkcol_1234567890123456789012'
```

They get exception:
```
sqlalchemy.exc.DatabaseError: (cx_Oracle.DatabaseError) ORA-00904: "T3"."PKCOL_1234567890123456789012": invalid identifier
```

There are two factors in play here:

1) The Oracle database is <= version 12.1 and therefore has a 30 character identifier size limit
2) The primary key column is defined as NUMBER(*), i.e. no precision or scale has been specified. This results in DVT injecting a CASE expression to work around an inconsistency in how Oracle TO_CHAR treats fractional numbers >0 and <1. https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1201

Due to item 1 above SQL Alchemy is (unnecessarily) truncating the column alias:

```
sqlalchemy.exc.DatabaseError: (cx_Oracle.DatabaseError) ORA-00904: "T3"."PKCOL_1234567890123456789012": invalid identifier
[SQL: WITH t0 AS
(SELECT t4.pkcol_1234567890123456789012 AS pkcol_123456789012345678_1, t4.col2 AS col_2, CASE WHEN t4.pkcol_1234567890123456789012>0 AND t4.pkcol_1234567890123456789012<1 THEN '0'||TO_CHAR(t4.pkcol_1234567890123456789012) WHEN t4.pkcol_1234567890123456789012>-1 AND t4.pkcol_1234567890123456789012<0 THEN '-0'||TO_CHAR(ABS(t4.pkcol_1234567890123456789012)) ELSE TO_CHAR(t4.pkcol_1234567890123456789012) END AS cast__dvt_calc_col_0
FROM dvt_test.tab_00904 t4),
t1 AS
...
 SELECT t3.concat__all, CASE WHEN t3.pkcol_1234567890123456789012>0 AND t3.pkcol_1234567890123456789012<1 THEN '0'||TO_CHAR(t3.pkcol_1234567890123456789012) WHEN t3.pkcol_1234567890123456789012>-1 AND t3.pkcol_1234567890123456789012<0 THEN '-0'||TO_CHAR(ABS(t3.pkcol_1234567890123456789012)) ELSE TO_CHAR(t3.pkcol_1234567890123456789012) END AS pkcol_1234567890123456789012
FROM (SELECT t2.pkcol_123456789012345678_1 AS pkcol_123456789012345678_1, t2.col2 AS col2, t2.cast__dvt_calc_col_0 AS cast__dvt_calc_col_0, t2.ifnull__cast__dvt_calc_c_7 AS ifnull__cast__dvt_calc_c_7, t2.rstrip__dvt_calc_col_0 AS rstrip__dvt_calc_col_0, rstrip__dvt_calc_col_0 AS concat__all
FROM t2) t3]
```

Notice `t4.pkcol_1234567890123456789012 AS pkcol_123456789012345678_1` above.

The CASE expression in `t0` is fine because the column is available. But in the final query block the CASE expression is incorrect because it uses the original name `t3.pkcol_1234567890123456789012` and not the truncated name `pkcol_123456789012345678_1`.

nj1973:
I'm not happy about resorting to this "fix" because it is not really a fix.

Explanation of the problem this PR works around. This is a real edge case based on a number of colliding factors:

- Oracle version <= 12.1 (i.e. max identifier length 30). Oracle versions <= 12.1 are not exactly rare but not current and will become rarer over time
- Primary key column name >= 24 characters. Primary key names tend not to be very wordy
- Primary key defined as NUMBER(*). Primary key columns tend to be defined as integers of some kind and not the scale unknown NUMBER(*)
- For fractional columns we use a CASE expression when casting to string to cater for an Oracle issue that omits the zero for numbers less than ABS(1), e.g. .123 instead of 0.123. We don't know the scale for a loosely defined NUMBER(*) column so have to check it with the CASE expression.
- SQLAlchemy has a weird behaviour of renaming (by alias) columns with a name within 6 characters of the limit. For the Oracle version in play that is a 24 character limit.
- Primary key columns with a greater precision than bigint/double are lossy in Pandas and therefore need to be cast to string. Again a loosely defined NUMBER(*) column does not tell us the precision and we therefore have to assume it is large.
- The cast to string fails because SQLAlchemy renamed the column because of the long(ish) name
If the column was correctly defined as NUMBER(p,0) then we would no end up in this edge case.

I've invested many hours trying to find an elegant solution and cannot really justify spending more time on it for such an edge case, especially one that will become less likely as we move further away from Oracle versions <= 12.1.


helensilva14 (approved):
Very specific case and you added all the explanation about this approach, LGTM!

Issue #659, opened 2023-01-12
Error handling for --config-dir YAMLs


nehanene15:
When running YAMLs from a config directory, we should catch and handle errors so that users can pinpoint which config failed.
Since each YAML runs as a separate validation, we should log the name of the YAML file that is currently being processed and if there was a failure.

kanhaPrayas:
PR: https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/663

PR #1317, opened 2024-11-06
fix: Issue 1312 incorrect status code

### LINKED ISSUE #1312: Success status returned when validations in YAML files throw exception
Hi,

DVT returns success status (i.e. exit (0)) even when the YAML file throws an exception (after printing message "Error %s occurred while running config file %s. Skipping it for now.").

This is problematic when the YAML file is run using Cloud Run and there is an intermittent error (like failure writing to BQ). The expectation is that DVT would return a failure status code (non zero exit status) - which could result in the task being retried. 

Sundar Mudupalli

sundar-mudupalli-work:
Hi,

This fixes the issue that DVT returns status code (0) even if there is an exception while validating the yaml file. This results in Cloud Run considering the validation a success even when there is a transient error such as writing to BigQuery.

The issue [Error handling for --config-dir YAMLs](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/659) was raised. The issue is that when customers are validating multiple yaml files in a directory, customers had to fix one yaml at a time. When one yaml failed validation and raised an exception, the rest of the yaml files were not processed. This was fixed in [PR605](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/663). The fix was to catch exceptions raised, throw an error and skip to processing the next file. However, the  exception raised in processing one of the files was lost. This "losing the exception" happened even if only one yaml file was being processed. 

The fix is to move the exception processing logic to where the config runner is processing multiple files, so this behavior is only when multiple files are being processed. A flag is set to signal that one of the files had an exception and an exception is raised after processing all the files. In this way, the status is reported correctly. A config runner unit test has also been added.

Sundar Mudupalli


data_validation/__main__.py
```diff
@@ -388,9 +388,24 @@ def config_runner(args):
...
+                    run_validations(args, config_managers)
+                except Exception as e:
+                    errors = True
+                    logging.error(
+                        "Error %s occurred while running config file %s. Skipping it for now.",
+                        str(e),
+                        file,
+                    )
+            if errors:
+                raise Exception("Some of the validations raised an exception")
```
nj1973: Rather than raising Exception I wonder if it would be better to add an exception to `exceptions.py` and raise that instead? I think raising a vanilla `Exception` is frowned upon.

tests/unit/test__main.py
```diff
@@ -162,3 +172,35 @@ def test_config_runner_3(mock_args, mock_build, mock_run, caplog):
...
+    args = cli_tools.get_parsed_args()
+    caplog.clear()
+    with pytest.raises(Exception) as e_info:
+        main.config_runner(args)
+    # assert that exception message was output for the failed validation
+    # validation is called four times, once for each file
+    # After all four files were validated, an exception was raised back to main to return status
+    assert (
+        caplog.messages[0]
+        == "Error Boom! occurred while running config file 0001.yaml. Skipping it for now."
```
nj1973: Seeing as we are using this text as an assertion for a test I wonder if we should convert it to a constant?

For example:
```
assert (
        caplog.messages[0]
        == CONFIG_RUNNER_EXCEPTION_TEXT.format("Boom!, "0001.yaml")
)
```

Then if the message text changes the assertion automatically follows suit.

nj1973 (commented):
Left a couple of minor discussion points.

data_validation/__main__.py
```diff
@@ -388,9 +388,24 @@ def config_runner(args):
...
+                try:
+                    logging.info(
+                        "Currently running the validation for YAML file: %s",
+                        file,
+                    )
+                    run_validations(args, config_managers)
+                except Exception as e:
+                    errors = True
+                    logging.error(
+                        "Error %s occurred while running config file %s. Skipping it for now.",
```
```suggestion
                        "Error '%s' occurred while running config file %s. Skipping it for now.",
```
To make it clearer what part of the string is the exactly error message returned.

data_validation/__main__.py
helensilva14: Agreed! Something like `YAMLConfigFileException`

sundar-mudupalli-work: Done

tests/unit/test__main.py
sundar-mudupalli-work: Done

data_validation/__main__.py
sundar-mudupalli-work: Thank you

nj1973 (approved):
LGTM

PR #1320, opened 2024-11-07
refactor: Optimize connection management to share connections

### LINKED ISSUE #1319: Column validation of many small tables is slow
A customer has reported that running a simple count column validation command over many small tables is very slow. They cannot see a lot of database activity and feel the time is being spent in DVT code.

nj1973:
This PR shares connections between multiple configuration managers. Prior to this PR a command with multiple tables supplied via `--tables-list` or one that is split due to a high column count will create new source and target connections for each `ConfigManager`. After this PR they will all share the same two connections.

The PR also enabled SQLAlchemy's pessimistic disconnect handling:
https://docs.sqlalchemy.org/en/20/core/pooling.html#disconnect-handling-pessimistic

The thinking here is that re-using connections increases the likelihood of sessions being dropped due to network blips or idle timeouts. SQLAlchemy can protect us from this.

I tested this on 100 empty tables with both source and target systems in the same cloud region, i.e. this is best case. The customer that raised this issues was validating across different geographic locations.

Schema validation
```
$ time data-validation validate schema -sc=ora -tc=pg --tables-list "dvt_many_tabs.*"  > /tmp/dvt_many_tabs.out
```

Elapsed times:
- Prior code: 0m16.201s
- This PR code: 0m4.764s

Column validation
```
$ time data-validation validate column -sc=ora -tc=pg --tables-list "dvt_many_tabs.*"  --min="*" > /tmp/dvt_many_tabs.out
```

Elapsed times:
- Prior code: 0m28.710s
- This PR code: 0m21.570s


nj1973:
I've converted this back to draft while I think about it some more. I am happy that the change above is sound in most cases but am concerned about idle connections being dropped and then failing subsequent validations.

For example, what if row validation is executed and split into two because of a high number of columns. The first validation pulls the data and then spends 15 minutes in Python code joining Dataframes and recording success/failure status. Then we move onto the second ConfigManager and find that the connection has been dropped due to a firewall or database policy.

My current options are:

1. Modify the connection sharing to be active only for schema and column validations where the gap between database activity will always be short
2. Try to simulate the disconnect behaviour I have described above and work on a "reconnect if disconnected" strategy
3. Do nothing and hope for the best :-)

I'm thinking option 1 is the best short term way forward but will think on it for a little while first.

nj1973:
I've worked around my concerns in comment https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1320#issuecomment-2464263398 by only introducing connection sharing for backends managed by SQLAlchemy, this is because SQLAlchemy handles disconnected session for us.

I think this is a good solution to my concerns.

data_validation/__main__.py
```diff
@@ -504,19 +504,32 @@ def convert_config_to_json(config_managers) -> dict:
...
-def run_validation(config_manager, dry_run=False, verbose=False):
+def run_validation(config_manager: ConfigManager, dry_run=False, verbose=False):
     """Run a single validation.
 
     Args:
         config_manager (ConfigManager): Validation config manager instance.
         dry_run (bool): Print source and target SQL to stdout in lieu of validation.
         verbose (bool): Validation setting to log queries run.
     """
+    # Only used cached connection for SQLAlchemy backends that manage reconnects for us.
```
```suggestion
    # Only use cached connection for SQLAlchemy backends that manage reconnects for us.
```

helensilva14 (approved):
LGTM, a nifty solution! Just a NIT suggestion, so approving it already.

PR #1323, opened 2024-11-08
fix: Fix --comp-fields bug losing option value before validation

### LINKED ISSUE #1322: Comparison fields option broken by issue 1222
Issue https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1222 introduced a bug that prevented `-comp-fields` option from working correctly. It now always assumes `*`.

nj1973:
Fixes a bug introduced in the prior release where we lose the `--comparison-fields` option value and always validate all columns.

### DIFF
--- a/tests/unit/test_config_manager.py
+++ b/tests/unit/test_config_manager.py
@@ -608,3 +608,27 @@ def test_get_none_run_id(module_under_test):
         SAMPLE_CONFIG, MockIbisClient(), MockIbisClient(), verbose=False
     )
     assert config_manager.run_id is None
+
+
+@mock.patch(
+    "data_validation.config_manager.ConfigManager.get_source_ibis_calculated_table",
+    new=lambda x: MockIbisTable(),
+)
+def test_build_comp_fields(module_under_test):
+    config_manager = module_under_test.ConfigManager(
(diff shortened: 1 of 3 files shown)


tests/unit/test_config_manager.py
```diff
@@ -608,3 +608,27 @@ def test_get_none_run_id(module_under_test):
...
+    config_manager = module_under_test.ConfigManager(
+        SAMPLE_CONFIG, MockIbisClient(), MockIbisClient(), verbose=False
+    )
+
+    # With exclude_columns=False
+    comparison_fields = config_manager.build_comp_fields(
+        ["a", "c", "e"],
+        False,
+    )
+    assert comparison_fields == {"a": "a", "c": "c"}
```
helensilva14, line 629: To confirm my understanding: column "e" isn't in the expected dict because  it only exists in the source table and not in the target? If not, not sure I understood here.

helensilva14 (approved):
Left just one question to better understand the new unit test and what's the expected behavior for the function. Besides that, LGTM!

tests/unit/test_config_manager.py
nj1973: Column "e" is not in either table and therefore is not expecting in the final list, I've added a comment to that effect.

PR #1324, opened 2024-11-11
feat: Auto populate -pks when not provided by user

### LINKED ISSUE #1253: Row validation primary keys option could be automated
I have a customer who would like to run row validation for multiple tables in a single command but they cannot because the primary key columns differ from table to table.

What they have ended up doing is running `find-tables` to a BASH variable and looping through the contents capturing primary key information from the Oracle data dictionary via `sqlplus` commands. They then construct the a CSV for the `-pks` option and run DVT commands in the loop.

The customer asked if this could be built in to DVT. Obviously our timeline would not fit theirs so the custom shell script route still needs to be worked on but it does seem like a reasonable idea. Many validation source tables will have primary keys defined and therefore option population could be automated.

Other thoughts:

- Not all tables have primary keys therefore some failures are to be expected
- Not all systems support primary keys, e.g. Hive, and would be out of scope
- Finding the correct dictionary queries for each backend and plugging them into Ibis add-ons could be tricky. These things are never as simple as they first appear.

nj1973:
This PR merged in code and tests to auto populate the --primary-keys option when not provided by the user.

It add a new `list_primary_key_columns()` method to Ibis `Backend` class for each supported engine. For a few engines, e.g. BigQuery/Hive, the new method returns None. For most engines we run an appropriate dictionary query and retrieve the primary columns.

In DVT we first run the `list_primary_key_columns()` method for the source client. If that cannot give us the columns then we run the method on the target client. If we cannot get a list of columns from either location then we throw an exception requesting the user provides the id columns in the usual way.


data_validation/cli_tools.py
```diff
@@ -590,8 +590,12 @@ def _configure_row_parser(
     required_arguments.add_argument(
         "--primary-keys",
         "-pk",
-        required=True,
-        help="Comma separated list of primary key columns 'col_a,col_b'",
+        required=is_custom_query,
+        help=(
+            "Comma separated list of primary key columns 'col_a,col_b'" ""
```
helensilva14:
```suggestion
            "Comma separated list of primary key columns 'col_a,col_b'"
```
I think there's this additional string was added unintentionally

third_party/ibis/ibis_biquery/api.py
```diff
@@ -0,0 +1,10 @@
+from ibis.backends.bigquery import Backend as BigQueryBackend
+
+
+def _list_primary_key_columns(self, database: str, table: str) -> list:
+    """Return a list of primary key column names."""
+    # TODO nj@2024-11-08 It's not clear if this is possible, we should revisit if it becomes a requirement.
```
helensilva14:
```suggestion
    # TODO: Related to issue-1253, it's not clear if this is possible, we should revisit if it becomes a requirement.
```

third_party/ibis/ibis_cloud_spanner/__init__.py
```diff
@@ -214,6 +214,11 @@ def drop_view():
     def fetch_from_cursor():
         pass
 
+    def list_primary_key_columns(self, database: str, table: str) -> list:
+        """Return a list of primary key column names."""
+        # TODO nj@2024-11-08 It's not clear if this is possible, we should revisit if it becomes a requirement.
```
helensilva14:
```suggestion
        # TODO: Related to issue-1253, it's not clear if this is possible, we should revisit if it becomes a requirement.
```

third_party/ibis/ibis_redshift/__init__.py
```diff
@@ -105,6 +105,11 @@ def _get_temp_view_definition(
     ) -> str:
         yield f"CREATE OR REPLACE TEMPORARY VIEW {name} AS {definition}"
 
+    def list_primary_key_columns(self, database: str, table: str) -> list:
+        """Return a list of primary key column names."""
+        # TODO nj@2024-11-08 It's not clear if this is possible, we should revisit if it becomes a requirement.
```
helensilva14:
```suggestion
        # TODO: Related to issue-1253, it's not clear if this is possible, we should revisit if it becomes a requirement.
```

third_party/ibis/ibis_postgres/client.py
```diff
@@ -126,6 +126,23 @@ def list_schemas(self, like=None):
...
 
+def _list_primary_key_columns(self, database: str, table: str) -> list:
+    """Return a list of primary key column names."""
+    # From https://wiki.postgresql.org/wiki/Retrieve_primary_key_columns
+    list_pk_col_sql = f"""
+        SELECT a.attname
+        FROM   pg_index i
+        JOIN   pg_attribute a ON a.attrelid = i.indrelid
+                             AND a.attnum = ANY(i.indkey)
+        WHERE  i.indrelid = '{database}.{table}'::regclass
```
helensilva14: Thanks for adding the reference link! I checked there, but just wondering: wouldn't parameter binding work here too?

data_validation/config_manager.py
```diff
@@ -1177,3 +1177,20 @@ def build_comp_fields(self, col_list: list, exclude_cols: bool = False) -> dict:
         )
 
         return casefold_source_columns
+
+    def default_primary_keys(self) -> list:
```
```suggestion
    def auto_list_primary_keys(self) -> list:
```
I think saying default brings a different connotation to the method, like if it would decide by itself which columns would be the PKs 🤔

data_validation/__main__.py
```diff
@@ -314,18 +312,23 @@ def build_config_from_args(args: Namespace, config_manager: ConfigManager):
...
-        # Append Comparison fields
-        if args.comparison_fields:
-            config_manager.append_comparison_fields(
-                _get_comparison_config(args, config_manager)
-            )
-
         # Append primary_keys
         primary_keys = cli_tools.get_arg_list(args.primary_keys)
+        if not primary_keys:
+            primary_keys = config_manager.default_primary_keys()
```
helensilva14: Maybe we could put the `config_manager.default_primary_keys()` method calling directly in line 316 in a lambda format? Because I'm not sure about having duplicate `if not primary_keys:` blocks in sequence and with that change we'd only have the exception handling one

helensilva14 (commented):
Few suggestions and a question

third_party/ibis/ibis_postgres/client.py
nj1973: Great suggestion, thanks.

data_validation/config_manager.py
nj1973: Yep, I agree. Thanks

data_validation/__main__.py
nj1973: I don't quite understand. This is the current code:
```
316        primary_keys = cli_tools.get_arg_list(args.primary_keys)
317        if not primary_keys:
318            primary_keys = config_manager.auto_list_primary_keys()
319        if not primary_keys:
320            raise ValueError("--primary-keys argument is required for this validation")
```

- We set primary keys to user input (L316).
- If the user didn't give us input (L317) we set primary keys from the source or target (L318).
- If we didn't get primary keys from source/target (L319) we fail the execution

It could be condensed to below if you think that is less confusing?
```
316        primary_keys = cli_tools.get_arg_list(args.primary_keys) or config_manager.auto_list_primary_keys()
317        if not primary_keys:
318            raise ValueError("--primary-keys argument is required for this validation")
```

helensilva14: Yes, I was thinking about the condensed version you put! Not necessarily because it's confusing right now, I just thought having both methods in one line would be more straightforward than having two identical `if` blocks one after the other. But at the end it's up to you, both approaches work and are understandable!

nj1973: Understood. I've implemented the condensed version. Thanks

helensilva14 (approved):
LGTM!

data_validation/__main__.py
sundar-mudupalli-work: Line 324 (old line #316) - is strange!! - Why not use `args.primary_keys`, just like in the old line # 318 `args.comparison_fields`. Since primary_keys is now optional, we should have a default value of None - see comment below.

data_validation/cli_tools.py
sundar-mudupalli-work: Please add `default=None` before new line #593.

data_validation/cli_tools.py
```diff
@@ -590,8 +590,12 @@ def _configure_row_parser(
...
+            if is_custom_query
+            else " (defaults to table primary key if available)"
```
sundar-mudupalli-work: I suggest we make --primary-keys an optional argument, not be required just for custom query. Generate table partitions can use custom queries, in which case primary keys are required, but if it uses tables, primary-keys may not be required (could trigger a different error in __main__.py line 321 above).

The better place to make this check is in [cli_tools.py](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/292420241b3670807520708378b8d079c5b9163c/data_validation/cli_tools.py#L210). Basically at that point you have received all the arguments, they are OK syntactically. There is something not quite right with the "semantics" - that is - if it is custom query row or generate-table-partitions with custom-query, primary keys are required - in all other cases, primary keys are not required. So that check can be made there.

`_check_custom_query_args` does something very similar, and is currently only used by generate-table-partitions. In the future, it can be extended for all validations if we decide to allow the user to specify either `-tbls` or `--source-query` and infer that it is `validate row` or `validate custom-query row` based on this specification. Today, anything other than generate-table-validations will fall to [this line](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/292420241b3670807520708378b8d079c5b9163c/data_validation/cli_tools.py#L203) because if it is `validate row` and `validate column` only `--tables-list` is allowed and if it is  `validate custom query row` and `validate custom query column`, `source-query` will be an attribute even if it is `None`, `tables_list` will not be an attribute.

Also Helen the extra `""` at the end of the first line is not a typo - it is actually `"" if is_custom_query else " (defaults to primary key(s) of the table if available)"`.  The linter for some reason split it in the wrong place. In light of the earlier comment, can you say something like "Primary keys when not specified will be inferred from the source or target table if available" - it is not specifically for custom query only.

data_validation/__main__.py
```diff
@@ -314,18 +312,24 @@ def build_config_from_args(args: Namespace, config_manager: ConfigManager):
...
-            )
-
         # Append primary_keys
-        primary_keys = cli_tools.get_arg_list(args.primary_keys)
+        primary_keys = (
+            cli_tools.get_arg_list(args.primary_keys)
+            or config_manager.auto_list_primary_keys()
+        )
+        if not primary_keys:
+            raise ValueError("--primary-keys argument is required for this validation")
```
sundar-mudupalli-work: Let us be more explicit here. Perhaps something like "No primary keys were provided on the command line and neither the source or target tables have primary keys. Please specify primary keys using --primary-keys argument"

sundar-mudupalli-work (changes requested):
Overall - looks good - I would like another quick look at the changes made before approving.

The documentation needs to be updated - there are 10 references to primary keys and some of them are now optional. 

data_validation/cli_tools.py
nj1973: Good points, thanks.

data_validation/__main__.py
```diff
@@ -314,18 +312,25 @@ def build_config_from_args(args: Namespace, config_manager: ConfigManager):
...
-                _get_comparison_config(args, config_manager)
-            )
-
         # Append primary_keys
         primary_keys = cli_tools.get_arg_list(args.primary_keys)
+        if not primary_keys and config_manager.validation_type != consts.CUSTOM_QUERY:
+            primary_keys = config_manager.auto_list_primary_keys()
+        if not primary_keys:
+            raise ValueError(
+                "No primary keys were provided and neither the source or target tables have primary keys. Please include --primary-keys argument"
```
sundar-mudupalli-work, line 321: I tried a custom query without any primary keys and got the following error `ValueError: No primary keys were provided and neither the source or target tables have primary keys. Please include --primary-keys argument`. Is this sufficient or should we say `Custom query validations must provide primary keys. No primary keys were provided and neither the source or target tables have primary keys. Please include --primary-keys argument`

I am OK either way.

Sundar Mudupalli

sundar-mudupalli-work (approved):
LGTM

data_validation/__main__.py
nj1973: I left it as it is, it's not totally accurate but makes it clear you need to provide the primary keys. Thanks for the good review, both you and Helen provided valuable direction for me.

Issue #1327, opened 2024-11-11
CLI Unit tests could cover more (all?) CLI options


sundar-mudupalli-work:
Hi,

[Issue 1325](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1325) highlighted that a bug introduced in [PR 1310](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1310)  could have been easily caught with a more comprehensive CLI unit test. We have CLI unit tests, but don't cover all the options - just one column validation CLI test.

Let us make a CLI unit test will the documented options.

Thanks.

Sundar Mudupalli

safe049:
i'd like to do this
may you assign me?

helensilva14:
Hi @safe049, assignment done! Thank you for your collaboration!

safe049:
Here you are!
https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1421

PR #1328, opened 2024-11-11
fix: Issue 1325

### LINKED ISSUE #1325: Random row filter is not working in latest release
In commit https://github.com/GoogleCloudPlatform/professional-services-data-validator/commit/783d1b09359490f1977ac66534c92dddd55a4e74#diff-ce217c386e702709267508ebe1bc725106c6bcba326c1ccb29951d506855ae6aR1434 the following change was made:

```
-    use_random_rows = getattr(args, "use_random_row", False)
-    random_row_batch_size = getattr(args, "random_row_batch_size", None)
+    use_random_rows = getattr(args, consts.CONFIG_USE_RANDOM_ROWS, False)
+    random_row_batch_size = getattr(args, consts.CONFIG_RANDOM_ROW_BATCH_SIZE, None)
```

Notice that `"use_random_row"` is not the same as the value of constant `consts.CONFIG_USE_RANDOM_ROWS`. The value is `"use_random_rows"` (with a trailing "s").

We need to either:

1. Revert the change on that line to `"use_random_row"` or
2. Change the name of the option to `--use-random-rows`

I'm surprised we don't have tests to cover this, but perhaps first we just fix this at the highest priority and create a new issue to add a test.

sundar-mudupalli-work:
Hi,

Random Row validation was not possible due to a CLI bug introduced in [PR 1310](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1310). Basically, the command line would only accept `--use-random-row` or `-rr`, while the backend could only understand `use_random_rows` which was always false because `'use_random_row' != 'use_random_rows'`. 

In order to not create a breaking change - the CLI accepts `--use-random-row` and `-rr` and translates them into `use_random_rows`, which the backend can understand.

We have tests for random rows, for multiple engines - they all bypass the CLI. We have a CLI unit test, but only for column validation. Random rows is an argument to Row validation, therefore opened [issue 1327](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1327)

Sundar Mudupalli

### DIFF
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -1431,7 +1431,9 @@ def get_pre_build_configs(args: Namespace, validate_cmd: str) -> List[Dict]:
     format = args.format if args.format else "table"
 
     # Get random row arguments. Only in row validations these attributes can be present.
-    use_random_rows = getattr(args, consts.CONFIG_USE_RANDOM_ROWS, False)
+    # Bad coding here, but keeping it so as not to introduce a breaking change. See
+    # consts.py Line 17 for a more detailed explanation.
+    use_random_rows = getattr(args, "use_random_row", False)
     random_row_batch_size = getattr(args, consts.CONFIG_RANDOM_ROW_BATCH_SIZE, None)
 
     # Get table list. Not supported in case of custom query validation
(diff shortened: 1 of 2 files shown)


nj1973 (approved):
LGTM, thanks for quick turnaround

PR #1329, opened 2024-11-12
chore: adding unit test coverage to `__main__.py`

nick-redfearn:
I originally started this PR with exit codes in mind, but I can see now that this is not needed.

Along the way, I added a bunch of unit test coverage to `__main__.py`. I feel like this is a useful addition for the future so I'm keeping the PR.

### DIFF
--- a/tests/unit/test__main.py
+++ b/tests/unit/test__main.py
@@ -329,10 +329,7 @@ def test_config_runner_4(mock_args, mock_build, mock_run, caplog):
     assert e_info.value.args[0] == "Some of the validations raised an exception"
 
 
-@mock.patch(
-    "argparse.ArgumentParser.parse_args",
-    return_value=argparse.Namespace(**VALIDATE_COLUMN_CONFIG),
-)
+@mock.patch("data_validation.__main__.run_validation")
 @mock.patch(
     "data_validation.__main__.build_config_managers_from_args",
     return_value=[
@@ -341,17 +338,17 @@ def test_config_runner_4(mock_args, mock_build, mock_run, caplog):
         )
     ],
 )
-@mock.patch("data_validation.__main__.run_validation")
+@mock.patch(
+    "argparse.ArgumentParser.parse_args",
+    return_value=argparse.Namespace(**VALIDATE_COLUMN_CONFIG),
+)
 def test_successful_column_validation_with_mocked_run_validation(
     mock_args, mock_build, mock_run
 ):
     main.main()
 
 
-@mock.patch(
-    "argparse.ArgumentParser.parse_args",
-    return_value=argparse.Namespace(**BROKEN_VALIDATE_COLUMN_CONFIG_MISSING_COMMAND),
-)
+@mock.patch("data_validation.__main__.run_validation")
 @mock.patch(
     "data_validation.__main__.build_config_managers_from_args",
     return_value=[
@@ -363,7 +360,10 @@ def test_successful_column_validation_with_mocked_run_validation(
         )
     ],
 )
-@mock.patch("data_validation.__main__.run_validation")
+@mock.patch(
+    "argparse.ArgumentParser.parse_args",
+    return_value=argparse.Namespace(**BROKEN_VALIDATE_COLUMN_CONFIG_MISSING_COMMAND),
+)
 def test_throws_for_malformed_input_config_missing_command(
     mock_args, mock_build, mock_run
 ):
@@ -372,10 +372,7 @@ def test_throws_for_malformed_input_config_missing_command(
     assert e_info.value.args[0] == "'Namespace' object has no attribute 'command'"
 
 
-@mock.patch(
-    "argparse.ArgumentParser.parse_args",
-    return_value=argparse.Namespace(**BROKEN_VALIDATE_COLUMN_CONFIG_INCORRECT_COMMAND),
-)
+@mock.patch("data_validation.__main__.run_validation")
 @mock.patch(
     "data_validation.__main__.build_config_managers_from_args",
     return_value=[
@@ -387,7 +384,10 @@ def test_throws_for_malformed_input_config_missing_command(
         )
     ],
 )
-@mock.patch("data_validation.__main__.run_validation")
+@mock.patch(
+    "argparse.ArgumentParser.parse_args",
+    return_value=argparse.Namespace(**BROKEN_VALIDATE_COLUMN_CONFIG_INCORRECT_COMMAND),
+)
 def test_throws_for_malformed_input_config_incorrect_command(
     mock_args, mock_build, mock_run
 ):
@@ -399,10 +399,7 @@ def test_throws_for_malformed_input_config_incorrect_command(
     )
 
 
-@mock.patch(
-    "argparse.ArgumentParser.parse_args",
-    return_value=argparse.Namespace(**VALIDATE_ROW_CONFIG),
-)
+@mock.patch("data_validation.__main__.run_validation")
 @mock.patch(
     "data_validation.__main__.build_config_managers_from_args",
     return_value=[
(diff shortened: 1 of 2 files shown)


tests/unit/test__main.py
```diff
@@ -35,47 +35,115 @@
...
-    "kube_completions": True,
 }
-CONFIG_RUNNER_ARGS_2 = {
+CONFIG_RUNNER_ARGS_2 = BASE_RUNNER_ARGS | {
+    "config_dir": "gs://pso-kokoro-resources/resources/test/unit/test__main/3validations",
+}
+CONFIG_RUNNER_ARGS_3 = BASE_RUNNER_ARGS | {
+    "config_dir": "gs://pso-kokoro-resources/resources/test/unit/test__main/4partitions",
+}
+CONFIG_RUNNER_ARGS_4 = CONFIG_RUNNER_ARGS_3 | {
```
nick-redfearn: Ive made sure not to change the values of `CONFIG_RUNNER_ARGS_1` + `CONFIG_RUNNER_ARGS_2` + `CONFIG_RUNNER_ARGS_3` + `CONFIG_RUNNER_ARGS_4`, Ive used merge syntax to reduce repeated rows.

(I can revert this if its preferred)

sundar-mudupalli-work:
Nick,

Thank you for contributing to the DVT code base. I am not clear specifically the issue you are trying to address. Our preferred approach is to first create an issue that clearly states the problem. Using that description, we should be able to write a test case that demonstrates the problem. We often do that and use that as a means of testing fixes. 

Today, when an exception is raised, a non zero status code is returned. Of course the exception trace is also included. For a tool like DVT which is used primarily by programmers and database admins, returning a non zero status along with the exception trace is sufficient, I think. 

Sundar Mudupalli

nick-redfearn (commented):
Thanks for looking @sundar-mudupalli-work 

I've realised since your review that I can achieve what I need using something like this...
```
python -m data_validation configs run -c ./config.yaml || echo "Take some action here"
```

As such, Ive reverted my `__main__.py` changes.

I'm updating the scope of the PR to "chore: adding unit test coverage to `__main__.py`"

helensilva14:
Hi @nick-redfearn! Unit tests failed in our Cloud Build execution, could you please take a look in the error below? Tks! 

```
__________________ ERROR collecting tests/unit/test__main.py ___________________
tests/unit/test__main.py:46: in <module>
    CONFIG_RUNNER_ARGS_1 = BASE_RUNNER_ARGS | {
E   TypeError: unsupported operand type(s) for |: 'dict' and 'dict'
```

nick-redfearn:
Thanks @helensilva14 - I think its because Python 3.8 doesnt support the | operand type, oops

Ive pushed this commit, I think it should be fine now -> https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1329/commits/15b09907da4942de06aa5e3dc7475f1d5f199bc6

tests/unit/test__main.py
nick-redfearn: Reverted here -> https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1329/commits/15b09907da4942de06aa5e3dc7475f1d5f199bc6

helensilva14:
/gcbrun

All good @nick-redfearn, thanks for the timely fix! Starting build again and let's see if all will pass!

tests/unit/test__main.py
```diff
@@ -207,3 +327,194 @@ def test_config_runner_4(mock_args, mock_build, mock_run, caplog):
...
+    "data_validation.__main__.build_config_managers_from_args",
+    return_value=[
+        config_manager.ConfigManager(
+            VALIDATE_COLUMN_CONFIG, MockIbisClient(), MockIbisClient(), verbose=False
+        )
+    ],
+)
+@mock.patch("data_validation.__main__.run_validation")
+def test_successful_column_validation_with_mocked_run_validation(
+    mock_args, mock_build, mock_run
```
nj1973, line 346: Either these arguments in are in the wrong order of the `mock.patch` calls are in the wrong order. I don't suppose it matters in practice as long as they do what they do. But I guess the names should match up.

tests/unit/test__main.py
```diff
@@ -207,3 +327,194 @@ def test_config_runner_4(mock_args, mock_build, mock_run, caplog):
...
+            BROKEN_VALIDATE_COLUMN_CONFIG_MISSING_COMMAND,
+            MockIbisClient(),
+            MockIbisClient(),
+            verbose=False,
+        )
+    ],
+)
+@mock.patch("data_validation.__main__.run_validation")
+def test_throws_for_malformed_input_config_missing_command(
+    mock_args, mock_build, mock_run
```
nj1973, line 368: Arg order.

tests/unit/test__main.py
```diff
@@ -207,3 +327,194 @@ def test_config_runner_4(mock_args, mock_build, mock_run, caplog):
...
+            BROKEN_VALIDATE_COLUMN_CONFIG_INCORRECT_COMMAND,
+            MockIbisClient(),
+            MockIbisClient(),
+            verbose=False,
+        )
+    ],
+)
+@mock.patch("data_validation.__main__.run_validation")
+def test_throws_for_malformed_input_config_incorrect_command(
+    mock_args, mock_build, mock_run
```
nj1973, line 392: Arg order.
I won't add any more of these because it's probably annoying. I'm sure you can find the rest yourself :-)

nj1973 (changes requested):
Broadly speaking, looks good. Just a minor comment regarding mock order.

tests/unit/test__main.py
nick-redfearn: @nj1973 
I dont see whats wrong, can you suggest what order the `mock_args, mock_build, mock_run` should be in. Im happy to make the change throughout the script

nj1973: I can't find it in the pytest docs but args are in reverse order to the `@mock.patch` calls.

I wasn't aware of this myself but noticed some other tests in the same file had them reversed, for example:
```
@mock.patch("data_validation.__main__.run_validations")
@mock.patch(
    "data_validation.__main__.build_config_managers_from_yaml",
    return_value=["config dict from one file"],
)
@mock.patch(
    "argparse.ArgumentParser.parse_args",
    return_value=argparse.Namespace(**CONFIG_RUNNER_ARGS_4),
)
def test_config_runner_4(mock_args, mock_build, mock_run, caplog):
```

So I added a breakpoint in your new test and had a look at the contents of the parameters. `mock_args` contained `@mock.patch("data_validation.__main__.run_validation")`.

It doesn't really matter in practice but I thought we should be consistent.

nick-redfearn: @nj1973 Thanks for explaining, I wasn't aware of this detail.

Ill make a fix now 👍

nick-redfearn: @nj1973 Fixed here -> https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1329/commits/796783564a28997b77cd0f4d689c73892139314a

tests/unit/test__main.py
nick-redfearn: Fixed here -> https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1329/commits/796783564a28997b77cd0f4d689c73892139314a

tests/unit/test__main.py
nick-redfearn: Fixed here -> https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1329/commits/796783564a28997b77cd0f4d689c73892139314a

nj1973 (approved):
LGTM, thanks again for contributing.

PR #1332, opened 2024-11-17
fix: catch InvalidArgument when checking if config vals are secrets

dmedora:
In #1285, SecretManager's exception catching scope was reduced from BaseException to just NotFound. This introduces a bug where if the value passed contains invalid chars for a secret name, eg. IP address for the 'host' config parameter, SecretManager throws InvalidArgument instead of NotFound. This PR catches InvalidArgument as well.

Sample error:
```
google.api_core.exceptions.InvalidArgument: 400 The provided Secret ID [projects/myproject/secrets/10.128.0.27/versions/latest] does not match the expected format [projects/*/secrets/*/versions/*]
```

### DIFF
--- a/data_validation/secret_manager.py
+++ b/data_validation/secret_manager.py
@@ -53,6 +53,6 @@ class GCPSecretManager:
             # Return the decoded payload.
             payload = response.payload.data.decode("UTF-8")
             return payload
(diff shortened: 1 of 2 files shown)


nj1973 (approved):
LGTM. Thanks for fixing.

PR #1339, opened 2024-11-21
fix: Prevent Oracle LOB columns from causing exceptions in Column and Row validation

### LINKED ISSUE #1335: Prevent column validation exceptions caused by Oracle CLOB/NCLOB columns used for JSON
**Test case**

Oracle:
```
CREATE TABLE dvt_test.tab_json
(
  id VARCHAR2(10) NOT NULL PRIMARY KEY
, col_json1 CLOB
, col_json2 CLOB
);
ALTER TABLE dvt_test.tab_json ADD CONSTRAINT tab_json_chk1 CHECK (col_json1 IS JSON) ENABLE;
ALTER TABLE dvt_test.tab_json ADD CONSTRAINT tab_json_chk2 CHECK (col_json2 IS JSON) ENABLE;
INSERT INTO dvt_test.tab_json VALUES (1,'{"a": 1}','{"a": 1}');
COMMIT;
```

PostgreSQL:
```
CREATE TABLE dvt_test.tab_json
(
  id varchar(10) NOT NULL PRIMARY KEY
, col_json1 json
, col_json2 jsonb
);
INSERT INTO dvt_test.tab_json VALUES (1,'{"a": 1}','{"a": 1}');
```

**Column validation**
DVT command Oracle vs PostgreSQL:
```
$ data-validation validate column -sc ora_local -tc pg_local -tbls="dvt_test.tab_json" --count '*' --sum '*' --min '*' --max '*' --avg '*' --std '*' --filter-status fail
...
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedFunction) function length(json) does not exist
LINE 2: ..._json2 AS col_json2, length(t1.id) AS length__id, length(t1....
                                                             ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

The problem is that Oracle stores JSON data in a CLOB.

**Row validation**
DVT command Oracle vs PostgreSQL:
```
$ data-validation validate row -sc ora_local -tc pg_local -tbls="dvt_test.tab_json" -pk=id --hash=col_json1,col_json2
...
  File "/usr/local/google/home/neiljohnson/github/professional-services-data-validator/.venv/lib/python3.11/site-packages/sqlalchemy/engine/default.py", line 736, in do_execute
cx_Oracle.DatabaseError: ORA-00932: inconsistent datatypes: expected CLOB got CHAR
```

nj1973:
Oracle LOB columns are pretty restrictive to use, for example:
```
SQL> select count(col_clob) from pso_data_validator.dvt_ora2pg_types;
select count(col_clob) from pso_data_validator.dvt_ora2pg_types
             *
ERROR at line 1:
ORA-00932: inconsistent datatypes: expected - got CLOB
```

In DVT we lose the true data type and work with Ibis canonical types, for example a CLOB column becomes `dt.String`, a BLOB column becomes `dt.Binary`. This PR adds code to fetch the raw data types for Oracle tables/custom queries and excludes BLOB, CLOB and NCLOB from Column and Row (hash/concat) validations.

I only added the `raw_metadata()` for Oracle because that is the only engines we currently need it for. If we run into similar issues for other engines where we need to differentiate between specific raw data types then we can extend this technique further.

Ideally we would be able to add expressions to truly validate these types. I see that as a feature request rather than an exception avoidance issue which is what this one is.


data_validation/config_manager.py
```diff
@@ -944,6 +946,15 @@ def require_pre_agg_calc_field(
                         f"Skipping {agg_type} on {column} due to data type: {column_type}"
                     )
                 continue
+            elif agg_type == "count" and self._is_oracle_lob(column):
```
helensilva14, line 952:
```suggestion
            # Oracle LOB columns are invalid for use with SQL COUNT function.
            elif agg_type == "count" and self._is_oracle_lob(column):
```

data_validation/config_manager.py
```diff
@@ -856,13 +864,7 @@ def require_pre_agg_calc_field(
             ]:
                 return True
             elif column_type in ["binary", "!binary"]:
-                if agg_type == "count":
-                    # Oracle BLOB is invalid for use with SQL COUNT function.
-                    return bool(
-                        self.source_client.name == "oracle"
-                        or self.target_client.name == "oracle"
-                    )
-                else:
```
helensilva14, line 867: With this block change, what will happen when the case is to count binary columns from other engines rather than Oracle? Because this part at least was able to return `False` in those scenarios. Just wondering 🤔

data_validation/config_manager.py
```diff
@@ -1008,6 +1019,66 @@ def _get_comparison_max_col_length(self) -> int:
...
+                )
+                self._target_raw_data_types = {
+                    _.casefold(): raw_data_types[_] for _ in raw_data_types
+                }
+            else:
+                self._target_raw_data_types = {}
+        return self._target_raw_data_types
+
+    def _is_oracle_lob(self, casefold_column_name: str) -> bool:
+        return bool(
```
helensilva14, line 1069:
```suggestion
         """Checks if a column in either the source or target database is an Oracle LOB (Large Object) data type. 
 
        Args:
          casefold_column_name: The case-insensitive name of the column to check."""
        return bool(
```

I don't know why the suggestion block is formatting the suggested documentation string with code syntax, but you can get the idea 😅

data_validation/config_manager.py
```diff
@@ -1008,6 +1019,66 @@ def _get_comparison_max_col_length(self) -> int:
...
+            .endswith("LOB")
+            or self._get_target_raw_data_types()
+            .get(casefold_column_name, "")
+            .endswith("LOB")
+        )
+
+    def _exclude_oracle_lob_columns(self, casefold_column_names: list):
+        """Remove LOB columns from validation to avoid ORA-00932 errors
+
+        Not using set() - set() because we want to retain the order of the list."""
```
helensilva14: Is this comment regarding the data type of `casefold_column_names` argument? If so, it'd be good to make it clearer if possible. Also the current text gives the impression of a subtraction operation between two sets.

third_party/ibis/ibis_oracle/__init__.py
```diff
@@ -163,3 +163,25 @@ def list_primary_key_columns(self, database: str, table: str) -> list:
                 list_pk_col_sql, parameters=(database.upper(), table.upper())
             )
             return [_[0] for _ in result.cursor.fetchall()]
+
+    def raw_metadata(
```
helensilva14:
```suggestion
    def raw_column_metadata(
```
A more descriptive name to differentiate from `def _metadata(self, query)` .

Probably it'd be good to add a quick documentation as well on how this method works. Not clear to me how `"SELECT * FROM {source} t0 WHERE ROWNUM <= 1"` is able to return the column and its data type 🤔

data_validation/config_manager.py
```diff
@@ -944,6 +946,16 @@ def require_pre_agg_calc_field(
                         f"Skipping {agg_type} on {column} due to data type: {column_type}"
                     )
                 continue
+            # Oracle LOB columns are invalid for use with SQL COUNT function.
+            elif agg_type == "count" and self._is_oracle_lob(column):
+                # TODO Eventually we need to allow COUNT(Oracle LOB) by using a CASE expression:
+                #      COUNT(CASE WHEN clob_col IS NULL THEN NULL ELSE 1 END)
```
sundar-mudupalli-work, line 954: Why skip this column when you can change the [query builder here ](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/11d96cef5cd8692b4d0537d45981ee639b995c35/data_validation/query_builder/query_builder.py#L38), using the [ibis.cases method](https://ibis-project.org/reference/expression-generic#ibis.expr.types.generic.Value.cases). Isn't that a simpler and better solution ?

Thanks.

Sundar Mudupalli

data_validation/config_manager.py
nj1973: I've tried to make it clearer.

data_validation/config_manager.py
nj1973: I don't think I can skip it in the query builder because I don't have data type information for both the source and target. I believe ConfigManager is where I have all of the information required to exclude a column from both source and target queries.

data_validation/config_manager.py
```diff
@@ -1008,6 +1020,70 @@ def _get_comparison_max_col_length(self) -> int:
...
+            .get(casefold_column_name, "")
+            .endswith("LOB")
+            or self._get_target_raw_data_types()
+            .get(casefold_column_name, "")
+            .endswith("LOB")
+        )
+
+    def _exclude_oracle_lob_columns(self, casefold_column_names: list):
+        """Remove LOB columns from validation to avoid ORA-00932 errors."""
+        # Not using set() minus set() below because we want to retain the order of the list."""
```
```suggestion
```
At the end I think we can remove this line, what do you think?

data_validation/config_manager.py
nj1973: As far as I'm aware count over binary data is fine in most cases, this is true for Oracle RAW too. It's just Oracle LOBs that are the issue, they are difficult to work with.
I added tests for most engines to double check. Only the DB2 test failed which I suspect is a lack of binary support in DVT rather than a DB2 issue. I created an issue and disabled the test.

third_party/ibis/ibis_oracle/__init__.py
nj1973: Done, thanks

data_validation/config_manager.py
helensilva14: Awesome, thank you!

helensilva14:
All good from my side and only @sundar-mudupalli-work's comment is left to be resolved.

/gcbrun

helensilva14:
@nj1973 - CLA check is failing because of my former Google email not being found anymore, sorry about that. I will need to go through the commits I did with the corp email and remove it from the authors

data_validation/config_manager.py
helensilva14: Hi! Based on your discussion during the sync call, is this comment going to be resolved as-is or there will be code changes related to it?

helensilva14 (approved):

sundar-mudupalli-work:
Neil,

This PR is a bit confusing. I thought it was ensuring that exceptions were not thrown when CLOBs and BLOBs are row and column validated. However I find that column validation with clobs and blobs works well. Column validations correctly count the number of non null columns I believe.

This PR removes (count) column validations of LOB and BLOB columns. I believe min, max etc are also supported. These should likely stay as they are today. 

The issue is with row validations, which throw an error. It would be nice to fix this and put out an info message. This PR fixes this and only puts out an INFO message if the verbose option is selected. We also have another case where an INFO message is output (when verbose option is selected) when a column is removed. In both cases, I believe an INFO message should be output - not just when verbose option is selected. I also opened another Issue to [remove the verbose option entirely](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1363).

My general philosophy on validations (please feel free to disagree) is that the customer is trusting DVT to tell them when two tables are different. If we are excluding items from the validation - we should tell the customer explicitly that these are being excluded. We should not quietly exclude things and only disclose information when the verbose option is used.

Thank you for working on this.

Sundar Mudupalli

nj1973:
> I find that column validation with clobs and blobs works well. Column validations correctly count the number of non null columns I believe.

I retested the test case in the issue (https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1335) and I still get the same exception:

```
sqlalchemy.exc.DatabaseError: (cx_Oracle.DatabaseError) ORA-00932: inconsistent datatypes: expected - got CLOB
```

BLOB columns will not have thrown an exception in the past because they are skilled due to this code (which I have replaced in this PR):
```
                if agg_type == "count":
                    # Oracle BLOB is invalid for use with SQL COUNT function.
                    return bool(
                        self.source_client.name == "oracle"
                        or self.target_client.name == "oracle"
                    )
                else:
```

nj1973:
Closing this PR because it is based on a misunderstanding of current behaviour. I'll start over and maybe cherry pick any parts of these changes I need for second attempt.

PR #1340, opened 2024-11-22
feat: Support customer defined api endpoint for BigQuery and Spanner

### LINKED ISSUE #1337: Require option to override BigQuery endpoint (bigquery.googleapis.com)
Some customers do not expose the default BigQuery endpoint of "bigquery.googleapis.com" internally and instead have a private endpoint.

This StackOverflow discussion shows how it might be achieved in client code: https://stackoverflow.com/questions/67093510/google-cloud-sdk-how-to-call-google-apis-or-urls-via-internal-gateway-url

```
from google.cloud import bigquery
from google.api_core.client_options import ClientOptions
client_options = ClientOptions(api_endpoint="https://myNewEndpoint") 
client = bigquery.Client(project="myProject",client_options=client_options)
...
```

I hoped there would be an override available via gcloud but I can't see bigquery listed in the output from:

```
gcloud config list api_endpoint_overrides/ --all
```

nj1973:
This PR add a new attribute to BigQuery and Spanner connection files: api_endpoint.

This is because some customers do not expose the default API endpoints, for example "bigquery.googleapis.com"` internally and instead have a private endpoint. Both the BigQuery and Spanner client support provision of a customer endpoint.

There was a further complication regarding the BigQuery Request Handler. Our command line shorthand does not really cater for extra attributes. I figured it made much more sense to allow a defined connection to be used in the result handler in place of the project. This future proofs us for the addition of more BigQuery related attributes.

### DIFF
--- a/data_validation/clients.py
+++ b/data_validation/clients.py
@@ -288,10 +288,10 @@ def get_data_client(connection_config):
             consts.GOOGLE_SERVICE_ACCOUNT_KEY_PATH
         )
         if key_path:
-            decrypted_connection_config["credentials"] = (
-                google.oauth2.service_account.Credentials.from_service_account_file(
-                    key_path
-                )
+            decrypted_connection_config[
+                "credentials"
+            ] = google.oauth2.service_account.Credentials.from_service_account_file(
+                key_path
             )
 
     if source_type not in CLIENT_LOOKUP:
--- a/data_validation/clients.py
+++ b/data_validation/clients.py
@@ -278,10 +278,10 @@ def get_data_client(connection_config):
             consts.GOOGLE_SERVICE_ACCOUNT_KEY_PATH
         )
         if key_path:
-            decrypted_connection_config[
-                "credentials"
-            ] = google.oauth2.service_account.Credentials.from_service_account_file(
-                key_path
+            decrypted_connection_config["credentials"] = (
+                google.oauth2.service_account.Credentials.from_service_account_file(
+                    key_path
+                )
             )
 
     if source_type not in CLIENT_LOOKUP:
--- a/data_validation/clients.py
+++ b/data_validation/clients.py
@@ -22,6 +22,7 @@ import google.oauth2.service_account
 import ibis
 import pandas
 from google.cloud import bigquery
+from google.api_core import client_options
 
 from data_validation import client_info, consts, exceptions
 from data_validation.secret_manager import SecretManagerBuilder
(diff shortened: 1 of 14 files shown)


data_validation/result_handlers/bigquery.py
```diff
@@ -62,16 +64,19 @@ def get_handler_for_project(
                 Explicit credentials to use in case default credentials
                 aren't working properly.
             status_list (list): provided status to filter the results with
+            api_endpoint (str): provided status to filter the results with
```
helensilva14:
```suggestion
            api_endpoint (str): BigQuery API endpoint (e.g. https://mybq.p.googleapis.com)
```

data_validation/clients.py
```diff
@@ -90,16 +91,26 @@ def get_client_call(*args, **kwargs):
     db2_connect = _raise_missing_client_error("pip install ibm_db_sa")
 
 
-def get_bigquery_client(project_id, dataset_id="", credentials=None):
+def get_google_bigquery_client(project_id, credentials=None, api_endpoint=None):
```
helensilva14:
```suggestion
def get_google_bigquery_client(project_id: str, credentials=None, api_endpoint: str =None):
```

data_validation/clients.py
```diff
@@ -90,16 +91,26 @@ def get_client_call(*args, **kwargs):
...
+    return bigquery.Client(
         project=project_id,
         client_info=info,
         credentials=credentials,
         default_query_job_config=job_config,
+        client_options=options,
+    )
+
+
+def get_bigquery_client(project_id, dataset_id="", credentials=None, api_endpoint=None):
```
helensilva14:
```suggestion
def get_bigquery_client(project_id: str, dataset_id="", credentials=None, api_endpoint: str=None):
```

tests/system/data_sources/common_functions.py
```diff
@@ -181,15 +183,17 @@ def schema_validation_test(
         f"-tc={tc}",
         f"-tbls={tables}",
         f"--exclusion-columns={exclusion_columns}",
-        "--filter-status=fail",
+        f"--filter-status={filter_status}" if filter_status else None,
         f"--allow-list={allow_list}" if allow_list else None,
         f"--allow-list-file={allow_list_file}" if allow_list_file else None,
+        (f"--bq-result-handler={bq_result_handler}" if bq_result_handler else None),
```
helensilva14: Curious and for my understanding: why adding parenthesis just for this new line?

helensilva14:
One last thing: did the customer tested this branch and everything worked? It's the only point I'd still check, because regarding the code it's making sense.

sundar-mudupalli-work:
> One last thing: did the customer tested this branch and everything worked? It's the only point I'd still check, because regarding the code it's making sense.

+1 to that. Since we don't have a test in the test suite for this - we need confirmation from the customer or we would need to set up a PSC in our project - have not looked into how complicated that is and if it affects our current settings.

Thanks.

Sundar Mudupalli

tests/system/data_sources/common_functions.py
nj1973: Good spot, I don't know if I added them or if my IDE added them. I've now removed them.

nj1973:
> One last thing: did the customer tested this branch and everything worked? It's the only point I'd still check, because regarding the code it's making sense.

Yes, the customer has tested the feature branch is now unblocked to start evaluating DVT.

nj1973:
@helensilva14 , I accepted one of your code suggestions but it was from your personal account which does not have Contributor License Agreement. Are you able to action that when you get time please? This PR will remain blocked until we resolve that.

helensilva14:
> @helensilva14 , I accepted one of your code suggestions but it was from your personal account which does not have Contributor License Agreement. Are you able to action that when you get time please? This PR will remain blocked until we resolve that.

Done! Everything is passing now and I'll take one last check to approve!

helensilva14 (approved):
LGTM!

PR #1343, opened 2024-11-22
chore: Add Python 3.11 to test coverage

### LINKED ISSUE #1321: test: Validate all integration tests with Python 3.11
Adding 3.11 and 3.12 to the Python versions list on `noxfile.py` will need to be addressed in a separate issue because first we need to make sure all clients work properly for each one of the versions. So far, I was able to get errors from Oracle and Hive with 3.12. And since we randomly choose a Python version for each system, it's not something that we can have certainty right now on when it will be selected or not.

_Originally posted by @helensilva14 in https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1303#issuecomment-2462420861_

nj1973:
This PR adds Python 3.11 to unit test coverage and into the assorted Pythons we pick from for integration tests.

To prepare for this I hardcoded all tests to 3.11 and they passed so I'm confident we can start testing on 3.11.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -27,7 +27,7 @@ import random
 import nox
 
 # Python version used for linting.
-DEFAULT_PYTHON_VERSION = "3.10"
+DEFAULT_PYTHON_VERSION = "3.11"
 
 # Python versions used for testing. TODO: Add 3.11 and 3.12 when issue-1321 is complete.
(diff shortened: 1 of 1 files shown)


helensilva14 (approved):
Awesome, LGTM!

Issue #1330, opened 2024-11-12
PostgreSQL XML data type not supported


nj1973:
Test case from a customer validating Oracle vs PostgreSQL.

PostgreSQL:
```
CREATE TABLE dvt_test.tab_xml
( id numeric(1000,0) NOT NULL PRIMARY KEY
, batch_id numeric(1000,0) NOT NULL
, xml_data xml);
INSERT INTO dvt_test.tab_xml VALUES (1,1,'<?xml version="1.0"?>
<Tests>
  <Test Id="1">
    <Name>Test 1</Name>
    <Command>test1.sh</Command>
  </Test>
  <Test Id="2">
    <Name>Test 2</Name>
    <Command>test2.sh</Command>
  </Test>
</Tests>');
```

Schema validation:
```
$ data-validation -v validate schema --source-conn pg --target-conn pg --tables-list "dvt_test.tab_xml"
...
    cursor.execute(statement, parameters)
psycopg2.errors.SyntaxError: syntax error at or near "dvt_test"
LINE 1: ...EMPORARY VIEW d8c496e55b324f97acba672aa8610c88 AS dvt_test.t...
```

Column validation:
```
$ data-validation -v validate column --source-conn pg --target-conn pg --tables-list "dvt_test.tab_xml"
...
    cursor.execute(statement, parameters)
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.SyntaxError) syntax error at or near "dvt_test"
LINE 1: ...PORARY VIEW "26d25695f1eb44ea9fbec0b11ac6190c" AS dvt_test.t...
                                                             ^
[SQL: CREATE TEMPORARY VIEW "26d25695f1eb44ea9fbec0b11ac6190c" AS dvt_test.tab_xml]
(Background on this error at: https://sqlalche.me/e/14/f405)
```

Row validation:
```
$ data-validation -v validate row --source-conn pg --target-conn pg --tables-list "dvt_test.tab_xml" -pk=id --hash="*"
...
    cursor.execute(statement, parameters)
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.SyntaxError) syntax error at or near "dvt_test"
LINE 1: ...PORARY VIEW "52ee1651e62a4392965f1569fe1a5792" AS dvt_test.t...
                                                             ^
[SQL: CREATE TEMPORARY VIEW "52ee1651e62a4392965f1569fe1a5792" AS dvt_test.tab_xml]
[...]

PR #1353, opened 2024-11-26
fix: 1261 pg custom query without creating view

### LINKED ISSUE #1261: Error validating Postgres using Custom Query "permission denied to create temporary tables"
Customer is trying to run below command for performing row validation between Postgres and Teradata using Custom query : 

data-validation -v validate custom-query row \
--source-conn pg_qa \
--target-conn tera_qa \
--source-query-file Z0005_PG.sql \
--target-query-file Z0005_TD.sql \
--primary-keys ID \
-comp-fields agreementId,decisionStatus,rejectionReason,approvalExpiryDate,decisionDate,etlLogId

They're getting error as : 
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.InsufficientPrivilege) **permission denied to create temporary tables in database "fcealfa_ods"**

[SQL: CREATE TEMPORARY VIEW "9f9ee039688446228a7cd00db332f08c" AS SELECT  id,  agreementId,  decisionStatus,  rejectionReason,  approvalExpiryDate,  decisionDate,  etlLogId FROM public.EodAgreementCreditDecision]

Seems while validating between Postgres and TD, DVT creates a temporary table on Postgres DB to fetch schema. 

Since Postgres is Source for Customer, it's difficult for them to get create table permissions there. 
Requesting a solution for the same.

sundar-mudupalli-work:
Hi,

The Postgres Custom Query validation used to create a temporary view. This fixes #1261, by finding the types from the custom query itself. Change only to one file that is specific for Postgres.

Sundar Mudupalli


third_party/ibis/ibis_postgres/client.py
```diff
@@ -72,22 +72,23 @@ def connect(dbapi_connection, connection_record):
 
 
 def _metadata(self, query: str) -> sch.Schema:
```
nj1973, line 74: I think you need a "SELECT" check here similar to what happens in the Oracle client:

https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/7bfdfa14c87f9014cd700bc23a8b19312cd2a4e5/third_party/ibis/ibis_oracle/__init__.py#L139

Otherwise we can get an exception here when we drop in for a simple table, not a custom query:
```
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.SyntaxError) syntax error at or near ")"
LINE 1: select * from (dvt_test.tab_xml) t0 limit 0
```

I used the test table from issue https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1330 to generate this example.

third_party/ibis/ibis_postgres/client.py
```diff
@@ -72,22 +72,23 @@ def connect(dbapi_connection, connection_record):
...
-        con.exec_driver_sql(f"DROP VIEW IF EXISTS {name}")
+        cur = con.exec_driver_sql(f"select * from ({query}) t0 limit 0")
+        qry_cols = [
+            f"('{column.name}'::text, {column.type_code},"
+            + f"{column.table_oid if column.table_oid else 'NULL'}::int,"
+            + f"{column.table_column if column.table_column else 'NULL'}::int, {idx})"
+            for idx, column in enumerate(cur.cursor.description)
+        ]
+        type_info = con.exec_driver_sql(
+            f"""select name, CASE WHEN t0.attrelid is NULL
```
nj1973: A bit of a nitpick but I personally don't like to see a mixture of SQL keywords in upper and lower case. If it were me I'd either upper case tokens like "SELECT", "FROM UNNEST", etc or lower case the "CASE", "THEN", "ELSE" tokens. Just to be consistent.

nj1973 (changes requested):
I wonder if we should replace the `test_custom_query_validation_core_types` test with one on `dvt_pg_types` to test with more data types?

third_party/ibis/ibis_postgres/client.py
helensilva14: +1 here, uppercasing all SQL keywords will make this method consistent with other SQL strings/expressions we have in other files.

third_party/ibis/ibis_postgres/client.py
```diff
@@ -72,22 +72,23 @@ def connect(dbapi_connection, connection_record):
...
-    WHERE attrelid = CAST(:raw_name AS regclass)
-    AND attnum > 0
-    AND NOT attisdropped
-    ORDER BY attnum"""
     with self.begin() as con:
-        con.exec_driver_sql(f"CREATE TEMPORARY VIEW {name} AS {query}")
-        type_info = con.execute(sa.text(type_info_sql).bindparams(raw_name=raw_name))
-        yield from ((col, _get_type(typestr)) for col, typestr in type_info)
-        con.exec_driver_sql(f"DROP VIEW IF EXISTS {name}")
+        cur = con.exec_driver_sql(f"select * from ({query}) t0 limit 0")
```
helensilva14:
```suggestion
        cur = con.exec_driver_sql(f"SELECT * FROM ({query}) t0 LIMIT 0")
```

third_party/ibis/ibis_postgres/client.py
```diff
@@ -72,22 +72,23 @@ def connect(dbapi_connection, connection_record):
...
+            + f"{column.table_column if column.table_column else 'NULL'}::int, {idx})"
+            for idx, column in enumerate(cur.cursor.description)
+        ]
+        type_info = con.exec_driver_sql(
+            f"""select name, CASE WHEN t0.attrelid is NULL
+                                THEN format_type(t0.type_code, NULL)
+                                ELSE format_type(t1.atttypid, t1.atttypmod) END as type
+                    from unnest(array[{','.join(qry_cols)}])
+                    as t0(name text, type_code int, attrelid int, attnum int, col_ord int)
+                    left join pg_attribute t1 using (attrelid, attnum) order by col_ord"""
```
helensilva14:
```suggestion
                                ELSE format_type(t1.atttypid, t1.atttypmod) END AS type
                    FROM UNNEST(array[{','.join(qry_cols)}])
                    AS t0(name text, type_code int, attrelid int, attnum int, col_ord int)
                    LEFT JOIN pg_attribute t1 USING (attrelid, attnum) ORDER BY col_ord"""
```

helensilva14 (commented):
Definitely a complex challenge to solve, well done Sundar!

third_party/ibis/ibis_postgres/client.py
sundar-mudupalli-work: Neil,
I am able to reproduce the issue - it looks like whenever you have a type that we don't recognize, then we throw an execption - even if that specific column is not part of the validation. I think this bug existed before the changes I made. We are off tomorrow - I will quickly check any response you provide.

Thanks.

Sundar Mudupalli

nj1973: My point is that we are generating SQL with a syntax error. We don't need to do that, we can easily create valid SQL using the same check as someone put into the Oracle equivalent code.

Perhaps that then still generates a later exception due to data type support and that would be a different issue, for the XML issue in this case, but my opinion is that this method should generate valid SQL.

third_party/ibis/ibis_postgres/client.py
sundar-mudupalli-work: Thank you Helen - fixed the ones I saw.

third_party/ibis/ibis_postgres/client.py
sundar-mudupalli-work: Ok, I realize now under what conditions it will raise this error - and believe have made it more clear - it will result in NotImplementedError when types like xml are used that don't have equivalent Ibis types.

Sundar Mudupalli

sundar-mudupalli-work:
> I wonder if we should replace the `test_custom_query_validation_core_types` test with one on `dvt_pg_types` to test with more data types?

The big difference is only in the time data types. Ibis support many more of the Postgres types - MAC Address, Polygon etc - we have not defined how to validate those. We may need to figure that out and include those types would be my suggestion.

Sundar Mudupalli

nj1973:
> Ibis support many more of the Postgres types - MAC Address, Polygon etc - we have not defined how to validate those

Agreed, I wasn't suggesting we look into any new data types. It just crossed my mind that we could switch the test from one existing table to another already existing table. Anyway - let's leave it alone for now.

nj1973 (approved):
LGTM

helensilva14 (approved):

PR #1358, opened 2024-11-28
docs: Refine connection secret support section

### LINKED ISSUE #1356: Incorrect wording regarding secrets in connections documentation
The lines below indicate that, when using secrets, all connection attributes must be secrets.

https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/7bfdfa14c87f9014cd700bc23a8b19312cd2a4e5/docs/connections.md?plain=1#L17-L18

This is not the case, we can mix and match secrets and string tokens. All string tokens will be checked to see if they refer to secrets but if they do not, the string token is used as normal.

While doing this we should also include a clear example that the URL can also be passed as a secret, for example:
```
data-validation connections add \
    --secret-manager-type GCP \
    --secret-manager-project-id my-project \
    --connection-name ora_by_url Oracle \
    --url="DVT-URL-SECRET"
```

nj1973:
Just some clearer wording and another example.


docs/connections.md
```diff
@@ -14,15 +14,24 @@ To do so simply add the GCS path to the environment i.e
 ## Using GCP Secret Manager
 DVT supports [Google Cloud Secret Manager](https://cloud.google.com/secret-manager) for storing and referencing secrets in your connection configuration.
 
-If the secret-manager flags are present, the remaining connection flags should reference secret names instead of the secret itself. For example,
-the following BigQuery connection references a secret with name 'dvt-project-id' stored in project MY-PROJECT.
+If the secret-manager flags are present, any of the remaining connection flags can reference secret names instead of the secret itself.
```
helensilva14, line 17: I think we should add an example mixing secret names and hardcoded values   to make it clear that's a possibility. Not sure if the text change only can clarify that

nj1973: Yes, that's a good idea. I added another example. Look okay?

helensilva14 (approved):
LGTM!

Issue #1366, opened 2024-12-05
Enhance random rows feature to support UUID primary keys


nj1973:
I'm spinning this off from https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1345 in order to manage the scope of the work. We do need this support but first we should get the basics in place.

Enable test `test_row_validation_uuid_rr_oracle_to_postgres` in `test_oracle.py` to produce the problem.

PR #1367, opened 2024-12-05
feat: Add UUID support

### LINKED ISSUE #1345: PostgreSQL uuid support
Hit this on an Oracle to PostgreSQL project.

**Test tabes**
Oracle:
```
create table dvt_test.tab_uuid_id
( id raw(16) primary key
, col_uuid raw(16)
, status varchar2(10));
insert into dvt_test.tab_uuid_id values
(hextoraw('387BDC3B218443B28EC23AC791C5B0F1')
,hextoraw('387BDC3B218443B28EC23AC791C5B0F1')
,'A');
insert into dvt_test.tab_uuid_id values
(hextoraw('397BDC3B218443B28EC23AC791C5B0F1')
,hextoraw('397BDC3B218443B28EC23AC791C5B0F1')
,'B');
commit;
```

PostgreSQL:
```
create table dvt_test.tab_uuid_id
( id uuid primary key
, col_uuid uuid
, status varchar(10));
insert into dvt_test.tab_uuid_id values
(uuid('387bdc3b218443b28ec23ac791c5b0f1')
,uuid('387bdc3b218443b28ec23ac791c5b0f1')
,'A');
insert into dvt_test.tab_uuid_id values
(uuid('397bdc3b218443b28ec23ac791c5b0f1')
,uuid('397bdc3b218443b28ec23ac791c5b0f1')
,'B');
```

**Column validation**
Exception:
```
data-validation validate column -sc ora -tc pg -tbls="dvt_test.tab_uuid_id" --min="*" --count="*" --sum="*"
...
ibis.common.exceptions.IbisTypeError: argument passes none of the following rules: value(<class 'ibis.expr.datatypes.core.Binary'>,)
```

**Row validation**
The primary key values do not match due to non-matching cast output.
```
data-validation validate row -sc ora -tc pg -tbls="dvt_test.tab_uuid_id" --hash="*"
...
╒═══════════════════╤═══════════════════╤══════════════════════╤══════════════════════╤══════════════════════════════════════════════════════════════════╤══════════════════════════════════════════════════════════════════╤══════════════════╤═════════════════════╤══════════════════════════════════════╕
│ validation_name   │ validation_type   │ source_table_name    │ source_column_name   │ source_agg_value                                                 │ target_agg_value                                                 │   pct_difference │ validation_status   │ run_id                               │
╞═══════════════════╪═══════════════════╪══════════════════════╪══════════════════════╪══════════════════════════════════════════════════════════════════╪══════════════════════════════════════════════════════════════════╪══════════════════╪═════════════════════╪══════════════════════════════════════╡
│ hash__all         │ Row               │ dvt_test.tab_uuid_id │ hash__all            │ 0fcbf19b675c5b0efe16a1659a0111f4a36b1f9e4de3f30e58fd72135f1041af │ nan                                                              │              nan │ fail                │ 8e6a1d33-108a-4863-9ee1-7276f9e8cc6c │
├───────────────────┼───────────────────┼──────────────────────┼──────────────────────┼──────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────┼──────────────────┼─────────────────────┼──────────────────────────────────────┤
│ hash__all         │ Row               │ dvt_test.tab_uuid_id │ hash__all            │ 1af7b36b0f40fe29dc1b3efda6abb4b1bacc4bafc6aebaf33877840461b2113f │ nan                                                              │              nan │ fail                │ 8e6a1d33-108a-4863-9ee1-7276f9e8cc6c │
├───────────────────┼───────────────────┼──────────────────────┼──────────────────────┼──────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────┼──────────────────┼─────────────────────┼──────────────────────────────────────┤
│ hash__all         │ Row               │ dvt_test.tab_uuid_id │ nan                  │ nan                                                              │ ea533081c287693400dde05a9e58b09f65eb8117743c152eb322357a7132b518 │              nan │ fail                │ 8e6a1d33-108a-4863-9ee1-7276f9e8cc6c │
├───────────────────┼───────────────────┼──────────────────────┼──────────────────────┼──────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────┼──────────────────┼─────────────────────┼──────────────────────────────────────┤
│ hash__all         │ Row               │ dvt_test.tab_uuid_id │ nan                  │ nan                                                              │ a2942152e788779b4e7d434aa4d564d7ce5c4bb31e1f82909e85717aae909869 │              nan │ fail                │ 8e6a1d33-108a-4863-9ee1-7276f9e8cc6c │
╘═══════════════════╧═══════════════════╧══════════════════════╧══════════════════════╧══════════════════════════════════════════════════════════════════╧══════════════════════════════════════════════════════════════════╧══════════════════╧═════════════════════╧══════════════════════════════════════╛
```

**Random row validation**
```
data-validation validate row -sc ora -tc pg -tbls="dvt_test.tab_uuid_id" --hash="*" --use-random-row
...
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedFunction) operator does not exist: uuid = bytea
LINE 4: WHERE t6.id IN (decode('387bdc3b218443b28ec23ac791c5b0f1', '...
                    ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
```

nj1973:
This PR adds support for UUID type columns in cross engine validations. We already supported UUIDs but not for comparisons because some engines don't have a UUID type so we needed to be able to compare a UUID in binary or string columns with pure UUIDs.

It does not add tests for Snowflake or Teradata because they don't have a UUID type and their likely comparison target is BigQuery which also doesn't have a UUID type.

The code changes rely on introducing a cast to "uuid_string" and then intercepting it inside the cast method to do the right thing, we do this because we need to cast all UUIDs to a common pattern and some engines don't have a UUID type so we can't do the right thing inside the engine specific casts. We do something similar for booleans which also do not have universal support.

This PR does not add random row support, I instead created https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1366 for that in order to get the basics in place first.


tests/resources/bigquery_test_tables.sql
```diff
@@ -602,3 +602,14 @@ CREATE TABLE `pso_data_validator`.`dvt_bool`
 ) OPTIONS (description='Integration test table used to test boolean data type, especially in non-boolean columns.');
 INSERT INTO `pso_data_validator`.`dvt_bool` VALUES (1,true,true,true,true);
 INSERT INTO `pso_data_validator`.`dvt_bool` VALUES (2,false,false,false,false);
+
+-- BigQuery stores UUIDs in STRINGs, see GENERATE_UUID().
```
helensilva14:
```suggestion
-- BigQuery stores UUIDs in STRING data type according to its utility function called GENERATE_UUID().
```

data_validation/config_manager.py
```diff
@@ -586,12 +586,12 @@ def add_rstrip_to_comp_fields(self, comparison_fields: List[str]) -> List[str]:
 
             if (
                 source_ibis_type.is_string() or target_ibis_type.is_string()
-            ) and self._comp_field_cast(
-                # Do not add rstrip if the column is a bool hiding in a string.
+            ) and not self._comp_field_cast(
+                # Do not add rstrip if the column is a bool or UUID hiding in a string.
```
helensilva14, line 590: In the PR's description you said Teradata wasn't considered, but by looking at the change you did to this condition I realized the message logged inside of it contains a Teradata description hardcoded. Just an observation, I'm not sure where exactly this part is used by the new changes.

data_validation/config_manager.py
```diff
@@ -617,16 +617,47 @@ def _comp_field_cast(
     ) -> str:
         # We check with .get() below because sometimes field is a computed name
```
```suggestion
        # We check below if the field exists because sometimes it is a computed name
```

data_validation/query_builder/query_builder.py
```diff
@@ -220,6 +220,8 @@ def compile(self, ibis_table):
         comparison_field = ibis_table[self.field_name]
         alias = self.alias or self.field_name
         if self.cast:
+            # TODO Not sure what to do here yet.
+            # if self.cast != consts.CONFIG_CAST_UUID_STRING:
```
helensilva14: Is this comment still needed, such as is it still gonna be addressed on this PR?

data_validation/config_manager.py
```diff
@@ -763,15 +800,15 @@ def append_pre_agg_calc_field(
...
                 cast_type = "timestamp"
                 pre_calculated_config = self.build_and_append_pre_agg_calc_config(
                     source_column,
                     target_column,
                     calc_func,
                     column_position,
-                    cast_type,
-                    depth,
+                    cast_type=cast_type,
+                    depth=depth,
```
helensilva14: Why these changes were needed? Not clear to me.

third_party/ibis/ibis_addon/api.py
```diff
@@ -67,10 +74,13 @@ def same_type(from_type, to_type) -> bool:
     return op.to_expr()
 
 
-def force_cast(self, target_type: dt.DataType) -> Value:
+def force_cast(self, target_type: str) -> Value:
```
helensilva14, line 95: Wondering if changing data type for `target_type` argument didn't cause any side effects on other parts this method is used?

helensilva14 (commented):
Awesome job Neil! My comments are mostly questions to better understand the changes

third_party/ibis/ibis_addon/api.py
sundar-mudupalli-work: I think we should cast it into the standard form which includes hyphens - humans are going to read it and we want it to be understandable.

data_validation/config_manager.py
nj1973: Those arguments are named and not positional and therefore, in my opinion, calling code should reference them by name. The change was not "needed" but IMO that is how they should be used. Because I was referencing them in my other changes I also modified the call here.

third_party/ibis/ibis_addon/api.py
nj1973: Regarding changing "dt.DataType" to "str". That is just a type hint and I did it because when debugging my changes I noticed that we pass a string data type in to this function and not an Ibis data type. I was just correcting the hint to match what is being done in the code.

data_validation/query_builder/query_builder.py
nj1973: Great spot, that should have been removed. Thanks

data_validation/config_manager.py
nj1973: Good suggestion, thanks

data_validation/config_manager.py
nj1973: `self._comp_field_cast` will only return a value for a bool or uuid column. So my change above is simply updating the comment to include both data types. We could have a PostgreSQL uuid to Teradata "uuid-inside-a-string" validation. We don't need to follow the rstrip code path in those cases.

nj1973:
@sundar-mudupalli-work , I have modified the code to include hyphens in the formatted output.

data_validation/config_manager.py
helensilva14: Okay, I got the context now and understood this path isn't used for UUID and bool use cases. Regardless I do think the logging message is misleading since it can happen to other engines besides Teradata, right? For now, might be worth to put a TODO code comment as a flag for the message to be reviewed in the future as it's not something critical, what do you think?

data_validation/config_manager.py
helensilva14: I had to search a little bit to confirm my initial thought and according to the link below it looks like if parameters are not declared in the function with the special characters that enforce positional-only or keyword-only behavior, either approach can be used with no problems. So it doesn't matter at the end as both ways work, anyway it was good to recap this Python detail!

https://docs.python.org/3/tutorial/controlflow.html#positional-or-keyword-arguments

third_party/ibis/ibis_addon/api.py
helensilva14: > Regarding changing "dt.DataType" to "str". That is just a type hint and I did it because when debugging my changes I noticed that we pass a string data type in to this function and not an Ibis data type. I was just correcting the hint to match what is being done in the code.

Ahh okay, no worries! Forgot about it being just a type hint and not an enforcement, all good!

data_validation/config_manager.py
nj1973: I also thought it could apply to other engines but just took a look and turns out it is Teradata only:
```
    # As per #1190, add rstrip for Teradata string comparison fields
    if (
        config_manager.source_client.name == "teradata"
        or config_manager.target_client.name == "teradata"
    ):
        comparison_fields = config_manager.add_rstrip_to_comp_fields(comparison_fields)
```

helensilva14: Ahhh that makes sense, all good here then, thanks much for confirming Neil! Resolving this topic as well

helensilva14 (approved):
LGTM!

sundar-mudupalli-work (approved):
LGTM.

PR #1387, opened 2025-01-06
fix: Cast of Oracle Timestamp to Date removes the time component

### LINKED ISSUE #1378: Compare Oracle  date to PG date
Hi,
is there a clever way to compare Oracle date with PostgreSQL date columns? Is there a way to (silently) use trunc(date) in Oracle to compare with the date column in PG?
Otherwise rows will always be marked as different

nj1973:
When running a column validation with group by columns of a Timestamp column DVT casts the column to Date. This is fine for all engines except Oracle. Oracle date includes a time component so we need to additionally remove it, via the TRUNC function.

This PR adds the above change and some integration tests.


sundar-mudupalli-work (approved):
Neil - Happy New Year - welcome back.

Great work.

Sundar Mudupalli

Issue #1396, opened 2025-01-14
validate column: Pandas exception in combiner for min/max of dates/timestamps


nj1973:
If a table contains dates/timestamps with a wide gap from the present day then min/max validations throw an exception:
```
    raise ValueError("Trying to coerce float values to integers")
ValueError: Trying to coerce float values to integers
```

**Test tables**
Oracle:
```
CREATE TABLE dvt_test.dvt_tricky_dates (
  id            NUMBER(5) NOT NULL PRIMARY KEY
, col_dt_low    DATE
, col_dt_epoch  DATE
, col_dt_high   DATE
, col_ts_low    TIMESTAMP(0)
, col_ts_epoch  TIMESTAMP(0)
, col_ts_high   TIMESTAMP(0));
INSERT INTO dvt_test.dvt_tricky_dates VALUES
(1,DATE'1000-01-01',DATE'1970-01-01',DATE'9999-12-31'
,TIMESTAMP'1000-01-01 00:00:00',TIMESTAMP'1970-01-01 00:00:00',TIMESTAMP'9999-12-31 23:59:59');
COMMIT;
```

PostgreSQL:
```
CREATE TABLE dvt_test.dvt_tricky_dates (
  id            integer NOT NULL PRIMARY KEY
, col_dt_low    date
, col_dt_epoch  date
, col_dt_high   date
, col_ts_low    timestamp(0)
, col_ts_epoch  timestamp(0)
, col_ts_high   timestamp(0));
INSERT INTO dvt_test.dvt_tricky_dates VALUES
(1,DATE'1000-01-01',DATE'1970-01-01',DATE'9999-12-31'
,TIMESTAMP'1000-01-01 00:00:00',TIMESTAMP'1970-01-01 00:00:00',TIMESTAMP'9999-12-31 23:59:59');
```

**Test commands**

col_dt_low:
```
data-validation -v validate column -sc ora_local -tc pg_local  --tables-list=dvt_test.dvt_tricky_dates \
 --min="col_dt_low" --max="col_dt_low" 
Traceback (most recent call last):
  File "/professional-services-data-validator/.venv/bin/data-validation", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/professional-services-data-validator/data_validation/__main__.py", line 694, in main
    validate(args)
  File "/professional-services-data-validator/data_validation/__main__.py", line 670, in validate
    run(args)
[...]

PR #1397, opened 2025-01-14
feat: Cater for ExtractEpochSeconds overflowing integer

### LINKED ISSUE #1390: validate column: PostgreSQL exception with --wildcard-include-timestamp and old timestamps
PostgreSQL test table (note the old date used):
```
CREATE TABLE dvt_test.tab_old_ts
( id         int primary key
, col_old_ts timestamp(0));
INSERT INTO dvt_test.tab_old_ts VALUES (1,'1900-01-01'::date);
```

Test command:
```
data-validation -v validate column -sc pg -tc pg \
 --tables-list=dvt_test.tab_old_ts --sum="col_old_ts" --wildcard-include-timestamp
...
sqlalchemy.exc.DataError: (psycopg2.errors.NumericValueOutOfRange) integer out of range

[SQL: SELECT count(*) AS count, sum(t0.epoch_seconds__col_old_ts) AS sum__epoch_seconds__col_old_ts
FROM (SELECT t1.id AS id, t1.col_old_ts AS col_old_ts, CAST(EXTRACT(epoch FROM t1.col_old_ts) AS INTEGER) AS epoch_seconds__col_old_ts
FROM dvt_test.tab_old_ts AS t1) AS t0]
(Background on this error at: https://sqlalche.me/e/14/9h9h)
```

PostgreSQL test:
```
postgres=# select CAST(EXTRACT(epoch FROM date'1901-01-01') AS integer);
ERROR:  integer out of range
postgres=# select CAST(EXTRACT(epoch FROM date'1905-01-01') AS integer);
   extract
-------------
 -2051222400
```

The number of seconds between 1901 and the Unix epoch overflows a 32 bit integer.

The same would be true of dates far beyond the Unix epoch, for example 9999-12-31, which is sometimes used in applications as a placeholder for an open ended date range.

nj1973:
This PR fixes `ExtractEpochSeconds` for PostgreSQL, Teradata and SQL Server.

It also adds tests for large ranges of values for `ExtractEpochSeconds` to have to deal with.


third_party/ibis/ibis_addon/operations.py
sundar-mudupalli-work, line 1: Thank you for moving stuff out of ibis_addon. Every little bit helps.

tests/system/data_sources/test_mysql.py
sundar-mudupalli-work, line 1: Would have loved to have some test data with the max MySQL Timestamp - i.e.  3001-01-19 03:14:07.999999 - one way I suggest is to put it in its own row and filter it out using --fliter for those engines that it does not make sense I am wondering - might we have an issue like [1396](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1396), where the combiner has a fit.

Thank you.

Sundar Mudupalli

sundar-mudupalli-work (dismissed):
Could you please add tests for MySQL max timestamp.

tests/system/data_sources/test_mysql.py
nj1973: Thanks, you were absolutely right that MySQL falls into the same issue as other engines for min/max. I made a mistake when excluding the low/high columns, the limitation only applies to SUM.

I've rectified this mistake. Good spot.

renzokuken (approved):
LGTM. Thanks for re-organizing the addon library as well

nehanene15 (approved):

PR #1398, opened 2025-01-15
feat: add connections cli commands

jwerner08:
Hello!

I've been using this tool quite a bit lately and its helped me a lot, so I figured it would be a good idea to spread some love and give back to this great tool. 🫡

Essentially there are 2 new feature with this PR: 
1. The ability to perform a `connections delete -c CONN_NAME` command.
2. The ability to perform a `connections describe -c CONN_NAME` command.

The intention of adding these is to make managing connection config files easier natively within this tools CLI.

I also made more smaller updates that are formatting related: 
1. `connecitons list` now outputs with a bit more prefacing information and separated in blocks to make it easier for humans.
2. I linted the `connection.md` file with my markdown linter and added code block language types to make it easier to read in a code editor.

All of the features I've added should have unit tests in all the correct places. I tried my best to keep the styling consistent within each file. If I missed anything I'm happy to update...

Thanks for this awesome tool, and hopefully my contributions are well taken! 😸


data_validation/cli_tools.py
```diff
@@ -1085,13 +1106,29 @@ def store_connection(connection_name, conn):
...
+
+
 def list_connections():
     """List all saved connections."""
     mgr = state_manager.StateManager()
     connections = mgr.list_connections()
     for conn_name in connections:
         source_type = mgr.get_connection_config(conn_name).get("source_type")
-        logging.info(f"Connection Name: {conn_name} : {source_type}")
+        print(f"Connection Name: {conn_name}")
```
jwerner08, line 1121: I chose `print` because the output seemed more so for immediate user interaction and feedback rather than for logging

tests/unit/test_cli_tools.py
```diff
@@ -159,21 +176,46 @@ def test_get_connection_config_from_args():
     assert conn["project_id"] == "example-project"
 
 
-def test_create_and_list_connections(caplog, fs):
+def test_connections_comands(caplog, fs):
```
jwerner08: I tried splitting `connections` commands into their own test functions, but there's some reliance on storing configs for testing downstream `connections` commands, and I think I was running into an issue with doing that pattern - pyfakefs seems to create "fresh" environments/temp directories per test function? 

Happy to take suggestions, or just repeat config storage code to separate tests for `connections` commands if owners wish :)

nj1973:
Hi @jwerner08, thanks very much for contributing. I'll review this PR either later this week or early next.

data_validation/cli_tools.py
nj1973: I agree with that decision.

data_validation/state_manager.py
```diff
@@ -72,6 +82,18 @@ def get_connection_config(self, name: str) -> Dict[str, str]:
 
         return json.loads(conn_str)
 
+    def describe_connection(self, connection_name: str, output_format: str):
```
nj1973: We are a bit hit and miss when it comes to type hints but I try to add them when adding new code so think it would be good to include one here. `-> Union[dict, str]` perhaps?

docs/connections.md
```diff
@@ -47,13 +52,29 @@ data-validation connections add \
...
+data-validation connections delete -c CONN_NAME
+```
+
+## Describe an existing connection
+
+```sh
+data-validation connections describe -c CONN_NAME
+```
+
 The data validation tool supports the following connection types.
```
nj1973, line 78: I think this section is a continuation of the "List supporting connection types" section. Perhaps the two new sections should move above "List supporting connection types".

While we are at it can we change "supporting" to "supported" in the above header, I know it's not part of your change but we might as well fix that while we are here.

docs/connections.md
```diff
@@ -65,35 +86,37 @@ The data validation tool supports the following connection types.
 * [Postgres](#postgres)
 * [MySQL](#mysql)
 * [Redshift](#redshift)
-* [FileSystem](#filesystem-csv-parquet-orc-json-only)
-* [Impala](#Impala)
-* [Hive](#Hive)
-* [DB2](#DB2)
-* [AlloyDB](#AlloyDB)
+* [FileSystem](#filesystem-csv-orc-parquet-or-json-only)
```
nj1973, line 89: Thanks for fixing these 👍

docs/connections.md
```diff
@@ -65,35 +86,37 @@ The data validation tool supports the following connection types.
...
+```sh
 data-validation connections add -c CONN_NAME <connection type> -h
 ```
 
 Below are the connection parameters for each database.
 
 ## Raw
-```
+
+```sh
```
nj1973: Some of these are screen output and not shell commands. For example in this snippet we are getting syntax highlighting for "for", "<", ">" add "|" which we weren't previously.

Setting the language is better for real shell snippets, for example the "Using an Oracle wallet" section, but I don't think the help output examples should have the language defined.

tests/unit/test_cli_tools.py
nj1973: I agree with your train of thought. I think I've had the same thoughts about `test_create_and_list_connections` in the past. It would be nice to have individual tests for each step in the lift cycle but then there's a danger of us using "connections add" as a pre-req to "list" and if we break "connections add" then both tests fail and it's just as unclear as having multiple commands in a single test.

In the past I must have decided there were more important things to worry about so I'm happy to continue to ignore it :-)

However, if it is bugging you, then perhaps mocking a connection file into `fs` that you can then list is the way to do it without leaking tests into each other. Examples here: https://pytest-pyfakefs.readthedocs.io/en/latest/usage.html

data_validation/state_manager.py
```diff
@@ -72,6 +82,18 @@ def get_connection_config(self, name: str) -> Dict[str, str]:
...
+        """Describe an existing connection by returning connection details in yaml format.
```
nj1973, line 88: I have a nagging doubt about whether adding `describe` is a good idea. It probably is but I'd like to sleep on it.

My train of thought is that a super secure installation might do the following:

- Install DVT in GKE or Cloud Run using a service account with access to certain GCP services
- Place connection files in Cloud Storage that only the service account can access
- Give end users access to submit jobs to the service

This prevents the end users from knowing the credentials used for validations or from using those credentials outside of the secure service.

If we add a describe option then we've given the users an API get see things they maybe shouldn't see 🤔 

Having said that, while typing this comment I've realised that, in the super secure case, you would want to block all connection actions, including `add`, `list` and `delete` so the entry point would need to force `data-validation validate` rather than only `data-validation`. And probably would be using secrets too.

So probably this comment is irrelevant. But I'd still like to seek a second opinion. Thanks for your patience.

nj1973 (changes requested):
This is great work, thanks for your input.

I've requested a few minor changes and want to have a think about my `connections describe` comment.

data_validation/state_manager.py
jwerner08: Ah, Apologies - I missed that return type hint.

data_validation/state_manager.py
jwerner08: Additional context for `describe` and `delete` commands: 

My goal with adding `describe` and `delete` commands was to speed up local development, where I often had to troubleshoot and clean up connection files. Specifically, I found myself frequently having to:

Check inside the config directory to figure out which connection details I'd messed up, and then `cat {conn_details_filename}` and `rm -rf {conn_details_filename}` to remove those files.

Using a `describe` and `delete` command removes the need for that manual back-and-forth. But I also get that in a locked-down environment, you’d likely disable or hide this command behind some level of privilege.

Hopefully this makes sense, and I'm happy to wait for confirmation that the pattern I'm adding is ok.

docs/connections.md
jwerner08: Happy to fix this! 😸

docs/connections.md
jwerner08: Ah, I totally missed that in some of these code blocks - I'll update.

tests/unit/test_cli_tools.py
jwerner08: I'll do some due diligence in trying this approach.

docs/connections.md
jwerner08: This should be resolved!

tests/unit/test_cli_tools.py
jwerner08: Ended up figuring this out for cli tools connection command unit tests. :) Leaving the other tests (state manager, etc.) as is for now, but feel free to ping me if anything needs to change with this test or others!

data_validation/state_manager.py
jwerner08: This should be resolved!

docs/connections.md
jwerner08: This should be resolved!

data_validation/state_manager.py
nj1973: Having slept on this I've decided this is valid and therefore my train of thought can be ignored:

"in the super secure case, you would want to block all connection actions, including `add`, `list` and `delete` so the entry point would need to force `data-validation validate` rather than only `data-validation`. And probably would be using secrets too."

tests/unit/test_cli_tools.py
nj1973: Looks good. Thanks.

nj1973 (approved):
LGTM, thanks for contributing.

PR #1402, opened 2025-01-17
docs: Add Snowflake key-pair example

### LINKED ISSUE #1385: Need to add Snowflake connection using DVT command with RSA key.
i want setup a connection of Snowflake DB using RSA key as don't want to use password for security purpose. 

image

Can anyone suggest how can i setup the snowflake connection?
As i did not find the RSA key option in DVT connection guide.

nj1973:
Add Snowflake key-pair example to connections doc


helensilva14 (approved):
LGTM!

PR #1415, opened 2025-01-29
fix: Row validations fail for values with trailing newline

### LINKED ISSUE #1411: concat and hash row validations fail for values with trailing newline
A hash comparison of rows from two tables between Oracle 19c and Postgres 15 failed with Data Validator 7.1
We have used the concat comparison to identify the root cause: a VARCHAR2(140) column called "rmtinf" contains ASCII characters followed by a single trailing newline both in Oracle and Postgres. 

The concat value created by Data Validation for Oracle contains the trailing newline but the concat value for Postgres does not.

Oracle:      rtrim(t2.ifnull__cast__rmtinf)
Postgres: rtrim(t2.ifnull__cast__rmtinf, :rtrim_1)

[err.txt](https://github.com/user-attachments/files/18573568/err.txt)
[out.txt](https://github.com/user-attachments/files/18573566/out.txt)
[validate-concat-rmtinf.txt](https://github.com/user-attachments/files/18573567/validate-concat-rmtinf.txt)

nj1973:
This PR merged changes to align RStrip behaviour across as many SQL engines as possible. The engines changed here are:

- Oracle
- DB2
- Snowflake

BigQuery, PostgreSQL, Spanner and Teradata already behaved correctly.

MySQL and Hive rstrip() functions do not support characters other than space and therefore are not included in this PR.

SQL Server has mixed support for customizing the characters to remove and I've created a specific issue for this: https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1419

I've added integration tests for all systems above.


nehanene15 (approved):
Looks great!

To address one of your questions in the comments... I believe we originally added the rstrip to account for the teradatasql bug that adds extra padding when returning results. For validations that don't include TD as a source or target we probably don't need the rstrip as a sanitization step before hashing but it's good that this PR now fixes the whitespace bug 😄

tests/resources/oracle_test_tables.sql
```diff
@@ -764,7 +764,22 @@ CREATE TABLE pso_data_validator.dvt_tricky_dates (
 , col_ts_low    TIMESTAMP(0)
 , col_ts_epoch  TIMESTAMP(0)
 , col_ts_high   TIMESTAMP(0));
+COMMENT ON TABLE pso_data_validator.dvt_tricky_dates IS 'Integration test table used to test potentially difficult Timestamps.';
 INSERT INTO pso_data_validator.dvt_tricky_dates VALUES
 (1,DATE'1000-01-01',DATE'1970-01-01',DATE'9999-12-31'
 ,TIMESTAMP'1000-01-01 00:00:00',TIMESTAMP'1970-01-01 00:00:00',TIMESTAMP'9999-12-31 23:59:59');
 COMMIT;
+
+DROP TABLE pso_data_validator.dvt_tricky_strings;
```
helensilva14, line 773:
```suggestion
DROP TABLE IF EXISTS `pso_data_validator.dvt_tricky_strings;
```

tests/resources/snowflake_test_tables.sql
```diff
@@ -627,6 +627,21 @@ CREATE TABLE PSO_DATA_VALIDATOR.PUBLIC.DVT_TRICKY_DATES (
 , col_ts_low    TIMESTAMP(0)
 , col_ts_epoch  TIMESTAMP(0)
 , col_ts_high   TIMESTAMP(0));
+COMMENT ON TABLE PSO_DATA_VALIDATOR.PUBLIC.dvt_tricky_dates IS 'Integration test table used to test potentially difficult Timestamps.';
 INSERT INTO PSO_DATA_VALIDATOR.PUBLIC.DVT_TRICKY_DATES VALUES
 (1,DATE'1000-01-01',DATE'1970-01-01',DATE'9999-12-31'
 ,TIMESTAMP'1000-01-01 00:00:00',TIMESTAMP'1970-01-01 00:00:00',TIMESTAMP'9999-12-31 23:59:59');
+
+DROP TABLE PSO_DATA_VALIDATOR.PUBLIC.dvt_tricky_strings;
```
helensilva14:
```suggestion
DROP TABLE IF EXISTS PSO_DATA_VALIDATOR.PUBLIC.dvt_tricky_strings;
```

tests/resources/sqlserver_test_tables.sql
```diff
@@ -588,3 +588,15 @@ CREATE TABLE pso_data_validator.dvt_tricky_dates (
 INSERT INTO pso_data_validator.dvt_tricky_dates VALUES
 (1,'1000-01-01','1970-01-01','9999-12-31'
 ,'1000-01-01 00:00:00','1970-01-01 00:00:00','9999-12-31 23:59:59');
+
+DROP TABLE pso_data_validator.dvt_tricky_strings;
```
helensilva14:
```suggestion
DROP TABLE IF EXISTS pso_data_validator.dvt_tricky_strings;
```

tests/resources/teradata_test_tables.sql
```diff
@@ -695,3 +695,16 @@ INSERT INTO udf.dvt_tricky_dates VALUES
 ,TIMESTAMP'1000-01-01 00:00:00',TIMESTAMP'1970-01-01 00:00:00',TIMESTAMP'9999-12-31 23:59:59+00:00');
 -- col_ts_high value above forced to UTC based on article below, but we still get wrong answer from the test:
 --   https://support.teradata.com/knowledge?id=kb_article_view&sys_kb_id=0e81918ac36da9103eb2d88f05013138
+
+DROP TABLE udf.dvt_tricky_strings;
```
helensilva14, line 699:
```suggestion
DROP TABLE IF EXISTS udf.dvt_tricky_strings;
```

tests/system/data_sources/test_sql_server.py
```diff
@@ -520,6 +520,22 @@ def test_row_validation_tricky_dates_to_bigquery():
...
 
 
+@mock.patch(
+    "data_validation.state_manager.StateManager.get_connection_config",
+    new=mock_get_connection_config,
+)
+def test_row_validation_tricky_strings_to_bigquery():
+    """Test with string values containing special characters."""
+    pytest.skip(
+        "Skipping test_row_validation_tricky_dates_to_bigquery because the version of SQL Server we have does not support rtrim of all whitespace."
```
helensilva14, line 530:
```suggestion
        "Skipping test_row_validation_tricky_dates_to_bigquery because the version of SQL Server we have does not support rtrim of all whitespaces."
```

helensilva14 (changes requested):

sundar-mudupalli-work (approved):
Nick,

Thank you for the work on this. I understand this is holding up a customer - so will approve this.

I agree with Neha that we are overusing RTRIM/Rstrip and distorting validation results. I have opened [a separate issue](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1422) requesting this be fixed.

Thanks again for the hard work and the test cases!!

Sundar Mudupallli

tests/resources/oracle_test_tables.sql
nj1973: `DROP TABLE IF EXISTS` is not valid syntax in Oracle

tests/resources/teradata_test_tables.sql
nj1973: `DROP TABLE IF EXISTS` is not valid syntax in Teradata

tests/system/data_sources/test_sql_server.py
nj1973: The original statement is more correct. The plural of "whitespace" is also "whitespace". Like "sheep" and "sheep" - confusing I know :-)

helensilva14 (approved):
Regarding the `IF EXISTS` part, I definitely was unsure of it about Teradata and for Oracle I did some search before and I found it available, but I checked again and it's specific to version 23 apparently, so my bad!

I'll look further on proper ways to do this type of checking for TD and Oracle in the draft PR for the DDLs: #1416

Issue #31, opened 2020-06-18
CLI Support for Parallel Validation Execution


dhercher:
In the CLI __main__.py we currently run validations in series.  Since we will often be waiting on query responses this should be a configurable paramter to allow parallel execution.

Add:
- Num Parallel Workers Config to the arg parse
- Add logic to use multi-processing in the execution

nehanene15:
Closed with #654

PR #1417, opened 2025-01-29
feat: Print summary information after row validation

### LINKED ISSUE #1277: Print summary information after row validation
Hi,

When Row Validation is completed the result dataframe is sent to the output and to BigQuery. This is fine for row validations where the number of rows are small. When validating 10,000 to millions of rows, there is. a lot of wasted output - especially when most rows are "successfully" validated. We have the option `--filter-status=fail` so that only failures will be output - which is great.

However, we don't provide a summary - for e.g. how many rows were in the source and target, how many rows were in the source, but not in target and vice versa, how many rows were present in both source and target, but did not successfully validate. Providing this information will help summarize results of large scale validations that involve millions and billions of rows.

This can be done by using the `configuration_json` column which is meant for "JSON representation of the validation metadata". While this was initially meant for the configuration, we could use this column for the validation summary, particularly for row validation. I suggest a JSON object similar to the following
```
{
    "source_validation": {
        "table_name": "aaaa",
        "total_no_of_rows": 1234,
        "no_of_rows_successfully_validated": 123,
        "no_of_rows_present_not_successful": 12,
        "no_of_rows_not_present_in_other": 23
    },
    "target_validation": {
        "table_name": "aaaa",
        "total_no_of_rows": 1234,
        "no_of_rows_successfully_validated": 123,
        "no_of_rows_present_not_successful": 12,
        "no_of_rows_not_present_in_other": 23
    }
}
```

### LINKED ISSUE #1414: Report total number of validated rows
We must disable the output of success results for row validations in order to save disk space.
We would still like to compare the total number of rows validated with the rows in a table in order to calculate a validation coverage ratio.
Could data validation tool report the total number of validated rows - e.g. with a log message with log level INFO?

helensilva14:
## Description of changes

- `data_validation/__main__.py`: removed TODO comment for an issue already closed (#31)
- `data_validation/combiner.py`: implemented `_get_summary` function to log a summary report only of row validation results, including statistics on rows present in source but not in target, and vice versa; added type checking and type hinting for all functions' parameters in the file; added a call to `_get_summary` on function `generate_report` before returning its result; replaced some hardcoded strings to either existing or new constant variables.
- `data_validation/consts.py`: added new constant variables; in some arrays, replaced hardcoded strings for the equivalent constant variables.
- `data_validation/data_validation.py`: replaced two constant variables.
- `data_validation/result_handlers/bigquery.py`: updated code comment
- `tests/unit/test_combiner.py`: implemented unit test for `_get_summary` function

## Issues to be closed

Closes #1277
Closes #1414 

## Checklist
- [x] I have added unit and integration tests relevant to my change as needed
- [x] I have updated any relevant documentation to reflect my changes


helensilva14:
Print of what this version was able to save on BQ (very messy, but it's a start)

image

data_validation/combiner.py
```diff
@@ -24,8 +24,10 @@
 import logging
 import ibis
 import ibis.expr.datatypes as dt
+import ibis.expr.types as types
 
-from data_validation import consts
+from data_validation import consts, metadata
+from pandas import DataFrame
```
nj1973: If this is imported just for type hints then we should do it inside a TYPE_CHECKING condition. Search for TYPE_CHECKING in the repo for an example.

data_validation/combiner.py
```diff
@@ -24,8 +24,10 @@
 import logging
 import ibis
 import ibis.expr.datatypes as dt
+import ibis.expr.types as types
```
nj1973: If this is imported just for type hints then we should do it inside a TYPE_CHECKING condition. Search for TYPE_CHECKING in the repo for an example.

data_validation/combiner.py
```diff
@@ -392,16 +394,30 @@ def _join_pivots(source, target, differences, join_on_fields):
...
 
+    # maybe check how to restrict logic below for row validations only?
     joined = joined[
         joined,
         ibis.literal(run_metadata.run_id).name("run_id"),
         ibis.literal(run_metadata.labels).name("labels"),
         ibis.literal(run_metadata.start_time).name("start_time"),
         ibis.literal(run_metadata.end_time).name("end_time"),
+        # problem: current always prints table names for the amount of rows, but it saves on BQ, what is already something
+        ibis.literal(str({
```
nj1973: Need to consider JSON and CSV output too.

I'm not a fan of adding this metadata to the result data frame and think we should discuss other options.

nj1973 (dismissed):
Just a few minor comments and one that I think we should not go down this route and discuss other options.

data_validation/combiner.py
helensilva14: Thank you, gonna do that! It's the first time I tried to add type hinting to follow your good practice and I wasn't sure if it was correct 😅

sundar-mudupalli-work (changes requested):
Helen,

This is really close. What the customer is asking for ROW validation
1. Number of rows in the Source table - this is basically length of source_df
2. Number of rows with validation success - this you have calculated use that
3. Number of rows present in source and target, but failed validation - this is basically all failures where source_agg_value is not NULL and target_agg_value is not NULL
4. Number of rows present in source, but not in target - this is basically all failures where source_agg_value is not NULL and target_agg_value is NULL

Similarly for the target table. 

I think these are additional conditions that you can run on the Pandas table. Please try it out and if you run into difficulty, we can work on it together.

Thanks.

Sundar Mudupalli

sundar-mudupalli-work (approved):
Helen,

Thank you for completing this - very helpful - Neil can take it to his customer.
Also thank you for using good coding practices by including typing - I call that the gift that you gave us that keeps on. giving - in the form of clarity when making other changes.

Thanks.

Sundar Mudupalli

data_validation/combiner.py
```diff
@@ -25,17 +25,28 @@
 import ibis
 import ibis.expr.datatypes as dt
 
+from typing import TYPE_CHECKING
 from data_validation import consts
+from data_validation.consts import (
```
nj1973: We've already import consts on the line above. I think it is better to use consts.VALIDATION_STATUS and not import VALIDATION_STATUS directly.

data_validation/combiner.py
```diff
@@ -405,3 +431,45 @@ def _add_metadata(joined, run_metadata):
     ]
 
     return joined
+
+
+def get_summary(joined_df: "DataFrame", source_df: "DataFrame", target_df: "DataFrame"):
```
nj1973: Tiny nitpick, feel free to ignore.

As `get_summary` is private to this module I would prefix the name an underscore,  `_get_summary`

data_validation/combiner.py
```diff
@@ -405,3 +431,45 @@ def _add_metadata(joined, run_metadata):
...
+        rows_present_in_target_not_in_source = (
+            joined_df["source_agg_value"].isnull()
+            & joined_df["target_agg_value"].notnull()
+        )
+        rows_present_in_both_source_and_target = (
+            joined_df["source_agg_value"].notnull()
+            & joined_df["target_agg_value"].notnull()
+        )
+
+        logging.info(
```
nj1973: I wonder if we need to hide this message behind `-v` or at a lover log level.

We have a number of customers using CSV output and, because we don't yet support a file result handler, this new log message will appear in the CSV file.
```
data-validation validate row  --source-conn ora_local --target-conn ora_local --tables-list dvt_test.issue_1277a=dvt_test.issue_1277b --hash="*" --format=csv
02/11/2025 09:38:46 AM-INFO: {'total_source_rows': 9, 'total_target_rows': 9, 'total_rows_validated': 10, 'total_rows_success_validation_status': 7, 'total_rows_fail_validation_status': 3, 'failed_rows_present_in_source_not_in_target': 1, 'failed_rows_present_in_target_not_in_source': 1, 'failed_rows_present_in_both_source_and_target': 1}
validation_name,validation_type,aggregation_type,source_table_name,source_column_name,source_agg_value,target_table_name,target_column_name,target_agg_value,group_by_columns,primary_keys,num_random_rows,difference,pct_difference,pct_threshold,validation_status,run_id,labels,start_time,end_time
hash__all,Row,,dvt_test.issue_1277a,hash__all,a73fcf339640929207281fb8e038884806e2eb0840f2245694dbba1d5cc89e65,dvt_test.issue_1277b,hash__all,a73fcf339640929207281fb8e038884806e2eb0840f2245694dbba1d5cc89e65,"{""id"": ""1""}",{id},,,,0.0,success,1e6e40e1-c897-4ed9-8803-522f4540b452,[],2025-02-11 09:38:44.644402+00:00,2025-02-11 09:38:45.180847+00:00
hash__all,Row,,dvt_test.issue_1277a,hash__all,b2239ae33d9bc5614aeb2e439e5032b1fe77c32f7080d3211c1d03f1b17ffcf6,dvt_test.issue_1277b,hash__all,b2239ae33d9bc5614aeb2e439e5032b1fe77c32f7080d3211c1d03f1b17ffcf6,"{""id"": ""2""}",{id},,,,0.0,success,1e6e40e1-c897-4ed9-8803-522f4540b452,[],2025-02-11 09:38:44.644402+00:00,2025-02-11 09:38:45.180847+00:00
...
```

So we will break existing setups. I did wonder if we should use a custom logger but that appears to be a bad practice.

What do you think?

data_validation/combiner.py
```diff
@@ -405,3 +431,45 @@ def _add_metadata(joined, run_metadata):
...
+            {
```
nj1973: I wonder if this dict should also include the start/end time seeing as we have them in the metadata object?

I know we include these in some output but not all, for example screen output does not show them:
```
02/11/2025 09:31:08 AM-INFO: {'total_source_rows': 9, 'total_target_rows': 9, 'total_rows_validated': 10, 'total_rows_success_validation_status': 7, 'total_rows_fail_validation_status': 3, 'failed_rows_present_in_source_not_in_target': 1, 'failed_rows_present_in_target_not_in_source': 1, 'failed_rows_present_in_both_source_and_target': 1}
╒═══════════════════╤═══════════════════╤══════════════════════╤══════════════════════╤══════════════════════════════════════════════════════════════════╤══════════════════════════════════════════════════════════════════╤══════════════════╤═════════════════════╤══════════════════════════════════════╕
│ validation_name   │ validation_type   │ source_table_name    │ source_column_name   │ source_agg_value                                                 │ target_agg_value                                                 │   pct_difference │ validation_status   │ run_id                               │
╞═══════════════════╪═══════════════════╪══════════════════════╪══════════════════════╪══════════════════════════════════════════════════════════════════╪══════════════════════════════════════════════════════════════════╪══════════════════╪═════════════════════╪══════════════════════════════════════╡
│ hash__all         │ Row               │ dvt_test.issue_1277a │ hash__all            │ a73fcf339640929207281fb8e038884806e2eb0840f2245694dbba1d5cc89e65 │ a73fcf339640929207281fb8e038884806e2eb0840f2245694dbba1d5cc89e65 │                  │ success             │ 1d087ff3-fe27-453b-9f1d-75f621ad84a6 │
├───────────────────┼───────────────────┼──────────────────────┼──────────────────────┼──────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────┼──────────────────┼─────────────────────┼──────────────────────────────────────┤
│ hash__all         │ Row               │ dvt_test.issue_1277a │ hash__all            │ b2239ae33d9bc5614aeb2e439e5032b1fe77c32f7080d3211c1d03f1b17ffcf6 │ b2239ae33d9bc5614aeb2e439e5032b1fe77c32f7080d3211c1d03f1b17ffcf6 │                  │ success             │ 1d087ff3-fe27-453b-9f1d-75f621ad84a6 │
...
```

tests/unit/test_combiner.py
```diff
@@ -17,9 +17,15 @@
 import pandas
 import pandas.testing
 import pytest
+import logging
 
 from freezegun import freeze_time
 from data_validation import metadata, consts
+from data_validation.consts import (
```
nj1973: consts is imported above. I don't think we should import long lists of constants, better to use `consts.MY_CONSTANT`.

nj1973 (changes requested):

data_validation/combiner.py
nj1973: Actually, my comment is incorrect and I'll resolve it. The logger output goes to stderr and therefore does not interfere with CSV output. Apologies.

data_validation/combiner.py
nj1973: Also should it contain the run id? If there are multiple run ids in a single execution then do we need to be able to tie the stats back to the results?

nj1973:
I tested this code on a 10m row validation with no failures:
```
$ time data-validation validate row -sc=ora23 -tc=pg -tbls=dvt_test.tab_vol_10m -pk=id --hash='*' --format=csv --filter-status=fail >/tmp/dvt_test.tab_vol_10m.txt
02/11/2025 11:42:02 AM-INFO: {'total_source_rows': 10000000, 'total_target_rows': 10000000, 'total_rows_validated': 10000000, 'total_rows_success_validation_status': 10000000, 'total_rows_fail_validation_status': 0, 'failed_rows_present_in_source_not_in_target': 0, 'failed_rows_present_in_target_not_in_source': 0, 'failed_rows_present_in_both_source_and_target': 0}

real	5m17.948s
user	4m12.019s
sys	0m43.285s
```

I then commented out the `get_summary()` call and ran my test again:
```
time data-validation validate row -sc=ora23 -tc=pg -tbls=dvt_test.tab_vol_10m -pk=id --hash='*' --format=csv --filter-status=fail >/tmp/dvt_test.tab_vol_10m.txt

real	3m32.898s
user	2m44.205s
sys	0m25.949s
```

That's a 50% increase in elapsed time for the metrics. I don't know which data is causing that but I think we need to be careful not to add a lot of overhead.

data_validation/combiner.py
helensilva14: Not sure about adding the times, the end time might not be necessary since the log output will already have a timestamp, but start time might make sense.

Agreed on putting the run_id there!

helensilva14:
@nj1973  Thanks a lot for the performance test! I refactored it after using Gemini for insights and it might be significantly better now, could you please check the elapsed time again for your large table? Already changed the minor items pointed out, only pending the topic about adding more fields to the dict summary.

data_validation/combiner.py
```diff
@@ -405,3 +427,55 @@ def _add_metadata(joined, run_metadata):
...
+            & result_df["target_agg_value"].notnull()
+        )
+        present_in_both_tables = (
+            result_df["source_agg_value"].notnull()
+            & result_df["target_agg_value"].notnull()
+        )
+
+        logging.info(
+            {
+                "validation_run_id": result_df.loc[0, "run_id"],
```
nj1973: This is just a suggestion, you can ignore it.

If it were me I would pass in the metadata object and get run_id, start_time and end_time from there instead. It just logically seems like that is their "home" and they just happen to be duplicated (many times over) to results dataframe.

data_validation/combiner.py
```diff
@@ -405,3 +427,55 @@ def _add_metadata(joined, run_metadata):
...
+                "validation_start_time": result_df.loc[0, "start_time"].strftime(
```
nj1973: Do we need the "validation_" prefix for run_id, start_time and end_time? I think using the same name as used in other output would make sense.

Perhaps adding new constants for all keys in this dict would be a good idea too?

data_validation/combiner.py
```diff
@@ -405,3 +427,55 @@ def _add_metadata(joined, run_metadata):
     ]
 
     return joined
+
+
+def _get_summary(
+    result_df: "DataFrame", source_df: "DataFrame", target_df: "DataFrame"
+):
+    """Logs a summary report/stats of row validation results."""
```
nj1973, line 442: Seeing as there are some complex calculations in here I wonder if this whole section should be inside a try/except block with a logger call if an exception is raised?

That way any issues caused by this nice-to-have metadata are not fatal for the overall validation.

nj1973 (changes requested):
I've made a couple of small recommendations. But overall it's looking good.

I tested with and without the new code and elapsed time was unaffected, this is great.

I created a data mismatch and added an extra target row to a 1 million row table and saw good output in the log:
```
{'validation_run_id': 'cb882e3a-bec2-48c3-8cb8-37976754e290',
 'validation_start_time': '2025-02-18 13:02:24 UTC',
 'validation_end_time': '2025-02-18 13:02:36 UTC',
 'total_source_rows': 1000000,
 'total_target_rows': 1000001,
 'total_rows_validated': 1000001,
 'total_rows_success_validation_status': 999999,
 'total_rows_fail_validation_status': 2,
 'failed_rows_present_in_source_not_in_target': 0,
 'failed_rows_present_in_target_not_in_source': 1,
 'failed_rows_present_in_both_source_and_target': 1}
```

One question, have you tested the code works well for comparison fields where rows/validations are not the same thing?

helensilva14:
> One question, have you tested the code works well for comparison fields where rows/validations are not the same thing?

@nj1973 Yes, I tested with both `--hash="*"` and `--comparison-fields="*"` on the test tables I had which some rows were different between source and target.

nj1973 (approved):
LGTM, great work Helen.

PR #1418, opened 2025-01-30
fix: Change get_filters to not hang due to long filter strings

### LINKED ISSUE #1410: Filters option processing hanging in regex processing
The processing of the options for the command below hangs (or is at least very slow, I let it run for 4 minutes with no progress):
```
data-validation -v validate row  --source-conn ora --target-conn pg \
--tables-list pso_data_validator.dvt_core_types \
--primary-keys id --hash "*" \
--filters "COL_DATETIME between to_timestamp('1900-01-31 00.00.00.000','YYYY-MM-DD HH24.MI.SS.FF') and to_timestamp('2024-12-31 00.00.00.000','YYYY-MM-DD HH24.MI.SS.FF')":"COL_DATETIME between to_timestamp('1900-01-31 00.00.00.000','yyyy-mm-dd hh24.mi.ss.ms') and to_timestamp('2024-12-31 00.00.00.000','yyyy-mm-dd hh24.mi.ss.ms')"
^CTraceback (most recent call last):
...
  File "/usr/lib/python3.11/argparse.py", line 2551, in _get_value
    result = type_func(arg_string)
             ^^^^^^^^^^^^^^^^^^^^^
  File "/professional-services-data-validator/data_validation/cli_tools.py", line 1211, in get_filters
    if result := re.fullmatch(single_filter, filter_value):
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/re/__init__.py", line 171, in fullmatch
    return _compile(pattern, flags).fullmatch(string)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyboardInterrupt
```

The regex that is hanging is the `re.fullmatch` call here:
```
    single_filter = r"([^':]*('[^']*')*)*"
    double_filter = (
        r"(?P<source>" + single_filter + r"):(?P<target>" + single_filter + r")"
    )
    filter_config = []
    if result := re.fullmatch(single_filter, filter_value):
```

nj1973:
We had an issue in the regex used to split filters that was hanging for a long time on long filters.

I created a new utility function dedicated to splitting strings in separators that are not in single quotes and used that function to split `--filters` strings. We no longer get the timeouts.

This PR adds a new Python package for testing, `pytest-timeout`. I used this to write failing tests before I fixed the problem.


helensilva14 (approved):
LGTM, great catch!

nehanene15 (approved):
LGTM! Cool usage of pytest-timeout!

sundar-mudupalli-work (approved):
Hi,

I wrote the original code and the filter expression(s) and not sure why they were performing poorly. These are straightforward regular expressions - no lookahead, so in theory should only touch each character once and supposed to be super efficient.

I would have tried re.compile first to see if that fixed the issue. Perhaps you tried that.

However, since you have done all this work, I will not ask for a change.

Thank you.

Sundar Mudupalli

PR #1431, opened 2025-02-09
fix: Convert Snowflake connect_args to a dict before passing to Ibis

### LINKED ISSUE #1401: Unable to create Snowflake connection file via `connections add` if we specify config_args attribute
If I try to add a Snowflake connection with `connect-args` attribute it fails:
```
data-validation connections add -c snowflake_p8 Snowflake \
 --user=dvtuserp8 --password="" \
 --account="my-acc" \
 --database="my-db" \
 --connect-args='{"private_key_file": "/home/nobody/dvt/snowflake/rsa_key.p8", "private_key_file_pwd": "my-pass"}'
...
 in do_connect
    connect_args.setdefault("converter_class", _SnowFlakeConverter)
    ^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'setdefault'
```

The problem is that `--connect-args` remains a string value and is not converted to a dict.

We need to cater for this in `cli_tools.get_connection_config_from_args()`. If we manually create the connection file then things work from there on.

While fixing the issue we should probably add some help to the connections doc regarding connecting to Snowflake using a key file.

nj1973:
This PR fixes an issue where we didn't convert Snowflake connect_args attribute to a dict before passing to Ibis.

Existing Teradata code did this but did the conversion in patched Ibis code. I moved the conversion up a level so we do it in DVT code before passing to Ibis.

Also added some tests.

### DIFF
--- a/data_validation/util.py
+++ b/data_validation/util.py
@@ -12,9 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import json
 import logging
 import re
 import time
+
+from data_validation import exceptions
 
 
 def timed_call(log_txt, fn, *args, **kwargs):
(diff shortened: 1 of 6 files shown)


helensilva14 (approved):
LGTM!

PR #1446, opened 2025-02-26
test: Add datetime primary key tests

### LINKED ISSUE #1444: Missing tests with datetime primary key
We are missing tests with datetime primary key

nj1973:
Adding row validation tests for datetime primary key columns.

I have a customer who has some of these and want to be sure the basics work.


helensilva14 (approved):
LGTM, awesome work!

PR #1450, opened 2025-03-03
feat: Add --format option to raw query command

### LINKED ISSUE #1447: data-validation query output contains Python escapes and list/tuple brackets
We have a customer who is using `data-validation query` to retrieve values from the same connection descriptor they will also use to run a validation as part of an automation script. The output contains Python escape characters and list indicators which means values need to be post processed.

Test case:
```
$ Q="SELECT '{\"keyA\": \"value A\", \"keyB\": 12345, \"keyC\": 67890}' FROM dual"
$ echo ${Q}
SELECT '{"keyA": "value A", "keyB": 12345, "keyC": 67890}' FROM dual
$ data-validation query --conn ora_local --query="${Q}"
['(\'{"keyA": "value A", "keyB": 12345, "keyC": 67890}\',)']
$
```

Notice the extra leading characters `['(\'` and trailing characters `\',)']`.

Ideally these would not be there. Or perhaps there could be an option for raw output without any Python extras.

nj1973:
We have a customer who plans to use `data-validation query` to generate other DVT commands as part of their automation scripts. The attraction of doing this is that the automation can piggy back on the same connection configuration as DVT itself without needing a second secret management technique.

Unfortunately the `data-validation query` command does a simple print of a Python structure and therefore adds list and tuple brackets to the output and escapes quotes.

This PR adds a new `--format` option to `data-validation query`. The default output remains the same but the user (i.e. this specific customer) can add `--format=minimal` to their commands which will simplify the output enabling them to capture the output in variables and pass it to other scripts.


data_validation/raw_query.py
```diff
@@ -30,18 +35,42 @@ def run_raw_query_against_connection(args) -> list:
...
+                                 in the output.
     """
 
-    def row_to_str(row) -> str:
+    def row_to_tuple(row) -> tuple:
         """This prevents SQLAlchemy string truncation inside Row() objects by first converting them to a tuple."""
         if isinstance(row, Row):
-            return str(tuple(row))
+            return tuple(row)
+        elif isinstance(row, list):
```
nehanene15: nit: Could combine the conditional into `isinstance(row, (list, Row))` since the result is the same

nehanene15 (approved):

data_validation/raw_query.py
nj1973: Yep, that's a good "nit". Thanks

PR #1463, opened 2025-03-06
feat: Enhanced row validation summary logging (support for custom-query row, change to JSON compliant log output)

### LINKED ISSUE #1462: Enhancements request to combiner summary logged message
For issue https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1277 we added an INFO level summary message describing what was validated.

For example:
```
03/06/2025 01:50:06 PM-INFO: {'run_id': 'a3c775e4-b0d0-40f0-a97e-fb8d6bd4ed22', 'start_time': '2025-03-06 13:50:03 UTC', 'end_time': '2025-03-06 13:50:06 UTC', 'total_source_rows': 3, 'total_target_rows': 3, 'total_rows_validated': 3, 'total_rows_success_validation_status': 3, 'total_rows_fail_validation_status': 0, 'failed_rows_present_in_source_not_in_target': 0, 'failed_rows_present_in_target_not_in_source': 0, 'failed_rows_present_in_both_source_and_target': 0}
```

One of the customers that was interested in this summary information has pointed out a couple of details we missed:

1. The message is not there for custom-query row validations, only pure row validations.
2. As the message it there for users to consume it really ought to be pure JSON and not a Python "almost-JSON" value. Logging a Python object makes it harder to scrape and use the values.

helensilva14:
## Description of changes

- Adding support for custom-query row validations, not restricting anymore to only pure row validations.
- Changing the log object to be JSON compliant (dates following ISO format)
- Updates to related unit tests

## Issues to be closed

Closes #1462

## Checklist

### DIFF
--- a/data_validation/combiner.py
+++ b/data_validation/combiner.py
@@ -27,7 +27,6 @@ import ibis.expr.datatypes as dt
 
 from typing import TYPE_CHECKING
 from data_validation import consts
-from data_validation.util import datetime_to_iso_format
 
 if TYPE_CHECKING:
     from pandas import DataFrame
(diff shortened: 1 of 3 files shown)


data_validation/combiner.py
```diff
@@ -446,55 +447,63 @@ def _get_summary(
...
-        if (
-            not result_df.empty
-            and result_df.loc[0, consts.VALIDATION_TYPE] == consts.ROW_VALIDATION
-        ):
-            # Vectorized calculations for all counts.
-            success_condition = (
-                result_df[consts.VALIDATION_STATUS] == consts.VALIDATION_STATUS_SUCCESS
-            )
-            fail_condition = ~success_condition  # Invert success for fail condition.
+        if not result_df.empty:
```
nj1973: Here I would be tempted to flip the condition and return out. For example:
```
if result_df.empty:
    return
```

That means you won't need to indent the body any further than it already was.

data_validation/util.py
```diff
@@ -62,3 +64,53 @@ def dvt_config_string_to_dict(config_string: str) -> dict:
...
+    elif isinstance(obj, int):
+        if obj > 1000000000000:  # milliseconds
+            return datetime.fromtimestamp(
+                obj / 1000, tz=datetime.timezone.utc
+            ).isoformat()
+        else:  # seconds
+            return datetime.fromtimestamp(obj, tz=datetime.timezone.utc).isoformat()
+    else:
+        raise TypeError(
+            f"Object of type {obj.__class__.__name__} is not JSON serializable"
```
nj1973: I don't think we mean "JSON serializable" here?

data_validation/util.py
```diff
@@ -62,3 +64,53 @@ def dvt_config_string_to_dict(config_string: str) -> dict:
...
+        '2023-03-01T00:00:00+00:00'
+
+        >>> datetime_to_iso("invalid")
+        Traceback (most recent call last):
+            ...
+        TypeError: Object of type str is not JSON serializable
+    """
+    if isinstance(obj, datetime):
+        return obj.isoformat()
+    elif isinstance(obj, PandasTimestamp):
```
nj1973: Seeing as the start and end times can only be datetime values (we control where they get set and it's always a datetime as far as I can see) it might be simpler to just use `.isoformat()` in place of calling `datetime_to_iso_format()`? I understand that you've tried to plan for future data types but perhaps we just cater for `PandasTimestamp` and `int` start/end times if we ever change how we set them in the first place and not add this unused code?

nj1973 (commented):
A few minor comments

data_validation/combiner.py
helensilva14: Great catch, agreed!

data_validation/util.py
helensilva14: Makes sense! I tried to go for `.isoformat()` at first but then I started getting `TypeError: Object of type int64 is not JSON serializable`  and I ended up thinking the JSON serialization was maybe changing the data type behind the scenes, so that's why I came up with the method.

Much later on I finally realized the problem was actually in the other columns, so I added the `int` conversion to solve it, but I forgot to review the actual need of the new method. I tested it out and indeed it isn't needed anymore, thank you!

nj1973 (approved):
LGTM, thanks

Polkaponika (commented):
2025-07-19#Log.txt

PR #1465, opened 2025-03-10
fix: Prevent RecursionError in combiner by slicing Dataframes by validation count

### LINKED ISSUE #1430: validate column fails with RecursionError: maximum recursion depth exceeded in comparison
Data validation tool 7.2 fails to execute column validations for two tables with 100 / 139 columns with error "RecursionError: maximum recursion depth exceeded in comparison".

Please find attached the config files, a DDL scripts and the log files

[paymtort-minmax-by-day-v.log](https://github.com/user-attachments/files/18729024/paymtort-minmax-by-day-v.log)

[paymtord-minmax-by-day.log](https://github.com/user-attachments/files/18708593/paymtord-minmax-by-day.log)

[sql.zip](https://github.com/user-attachments/files/18708632/sql.zip)
[yamls.zip](https://github.com/user-attachments/files/18708540/yamls.zip)

nj1973:
## Description of changes
The issue behind this PR is `RecursionError` thrown in the validation combiner code at this line:

https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/66cff07d51f607452f56b19c888ad6a749a31de4/data_validation/combiner.py#L84

It appears that the code in `combiner._calculate_differences()` converts each validation (Dataframe column) to a subquery and then unions them all together. It is this operation that causes the `RecursionError`. I created a unit test to reproduce this as was getting the exception at 146 validations/columns which, if you were to run min, max, sum and count on a table, you would encounter at only 36 input columns.

I had four attempts to resolve this.

**1. Refactor the code**
I spent quite a long time trying to understand the code in `combiner._calculate_differences()` looking for a different way to get the same result. I couldn't find a better approach and wholesale reworking of that section might just be too risky. Perhaps someone with a better handle on Ibis might be more successful.

**2. Increase the recursion limit**
The recursion limit can be viewed/defined in Python by these functions:
```
sys.getrecursionlimit()
sys.setrecursionlimit(n)
```

I've seen a few SO posts stating that the default limit of 1000 is pretty conservative but no guidance on what a safe higher limit might be. I tested myself with assorted numbers of columns and settings for `recursionlimit`, and for each test I captured process memory.
```
Columns RecLimit Combiner Mem Delta Mb
------- -------- ---------------------
    100     1000 32MB
    150     1000 RecursionError
    150     2000 40MB
    200     2000 48MB
    250     2000 55MB
    300     2000 64MB
    350     2000 RecursionError
    350     2500 72MB
    400     2500 RecursionError
    400     3000 79MB
    450     3000 86MB
    500     3000 RecursionError
```

If we assume that most schemas have a table with 100 columns then `recursionlimit=3000` would be a good step in the right direction. But 100 columns is not extreme. If we decided to go down this route perhaps we should add an option to allow users to control the option if they need to, with a default of 2000 or 3000 to increase the default limit.

**3. Use multiple config managers**
We did this for row validations exceeding limits on source and target engines. It results in DVT issuing breaking the overall validation up into multiple queries with their own run ids. We could use that technique here to avoid this issue.

I don't like this option because this problem is in the combiner, we have successfully pulled data from both source and target. Each extra query we using increases the resources required in the source and target by around 100% which should be avoided if we can.

**4. Process the results in pieces**
The changes in this PR slice up the incoming source/target Dataframes by column, process them independently to prevent `RecursionError` and then concatenate the results together again. I think this protects us no matter how many validations/columns are in the results.

But it has generated quite a bit of change.

**Summary**

My thoughts on the options:

1. Refactor the code: This would be the best option if possible. I don't know if it is possible and it is potentially the riskiest option.
2. Increase the recursion limit: This is the simplest option but is not a fix, it only increases the threshold so we don't see the exception as often. Any side effects of increasing `recursionlimit` are unknown although I suspect increasing to 3000 would be fine.
3. Use multiple config managers: I don't like this option and did not explore it any further than thinking it through.
4. Process the results in pieces: This is the approach I have taken in this PR. It feels like a bit of a hack but does protect us more completely from `RecursionError`

I suspect options 2 and 4 are the only valid options.

Even though I have put so much effort into option 4, after typing all of this out I think I am leaning towards throwing this code away and going with option 2 instead. Therefore the main purpose of this PR is to get a review of the research and some opinions on 1, 2 and 4.

## Issues to be closed

Closes #1430

## Checklist


tests/system/data_sources/test_oracle.py
```diff
@@ -769,17 +769,46 @@ def test_find_views_and_tables():
...
     """
     row_validation_many_columns_test(expected_config_managers=4)
 
 
+@mock.patch(
+    "data_validation.state_manager.StateManager.get_connection_config",
+    new=mock_get_connection_config,
+)
+def test_row_validation_comp_fields_many_columns():
+    """Test row validation on a table with reserved words in column names."""
```
helensilva14:
```suggestion
    """dvt_many_cols row validation using comparison fields"""
```

tests/system/data_sources/test_oracle.py
```diff
@@ -769,17 +769,46 @@ def test_find_views_and_tables():
...
+    df = row_validation_test(
+        tables="pso_data_validator.dvt_many_cols",
+        tc="mock-conn",
+        comp_fields="*",
+        filter_status=None,
+    )
+    # There should be result per column per row = 399 for this table.
```
helensilva14:
```suggestion
    # There should be a result per column per row = 399 for this table.
```

data_validation/config_manager.py
```diff
@@ -138,10 +138,6 @@ def case_insensitive_match(self):
         """Return if the validation should perform a case insensitive match."""
         return self._config.get(consts.CONFIG_CASE_INSENSITIVE_MATCH) or False
 
-    def process_in_memory(self):
```
helensilva14, line 188: I wasn't aware of this method, removing it is a nice code refactor, great catch!

data_validation/combiner.py
```diff
@@ -34,27 +36,106 @@
...
 )
 
 
 def generate_report(
-    client,
     run_metadata: "RunMetadata",
-    source: "IbisTable",
-    target: "IbisTable",
+    source_df: pandas.DataFrame,
+    target_df: pandas.DataFrame,
```
helensilva14:
```suggestion
    source_df: "DataFrame",
    target_df: "DataFrame",
```
Shouldn't we use the same type hint approach as used for the other params? `DataFrame` is already imported on L34 specifically in the `TYPE_CHECKING` block

data_validation/combiner.py
```diff
@@ -34,27 +36,106 @@
...
     join_on_fields=(),
     is_value_comparison=False,
     verbose=False,
-):
+) -> pandas.DataFrame:
```
helensilva14:
```suggestion
) -> "DataFrame":
```

data_validation/combiner.py
```diff
@@ -34,27 +36,106 @@
...
+    source_df: pandas.DataFrame,
+    target_df: pandas.DataFrame,
+    join_on_fields=(),
+    is_value_comparison=False,
+    verbose=False,
+) -> pandas.DataFrame:
     """Combine results into a report.
 
     Args:
         client (ibis.client.Client): Ibis client used to combine results.
```
helensilva14:
```suggestion
```
Since `client` isn't a param anymore.

data_validation/combiner.py
```diff
@@ -34,27 +36,106 @@
...
+# This constant is a threshold at which we slice up the input Dataframes
+# and stitch them back together again after Ibis processing.
+COMBINER_COLUMN_SLICE_WIDTH = 120
+
 COMBINER_GET_SUMMARY_EXC_TEXT = (
     "Error while generating summary report of row validation results"
 )
 
 
 def generate_report(
```
helensilva14, line 49:
```suggestion
def get_validation_report(
```
I really liked the approach you took on having this wrapper around the main logic. But I think we should create specific names for each of them to let it very clear what which one is intended to be used for, instead of relying only on function documentation.

data_validation/combiner.py
```diff
@@ -34,27 +36,106 @@
...
+        pandas.DataFrame:
+            A pandas DataFrame with the results of the validation in the same
+            schema as the report table.
+    """
+    _check_schema_names(source_df, target_df)
+
+    join_on_fields = tuple(join_on_fields)
+
+    validation_columns = run_metadata.validations.keys()
+    slice_thresholds = [
```
helensilva14, line 85:
```suggestion
    # Explanation of the slice_thresholds list comprehension:
    # 1. len(validation_columns): Gets the total number of validation columns.
    # 2. len(...) / COMBINER_COLUMN_SLICE_WIDTH: Divides the total columns by the slice width.
    # 3. int(...): Converts the result to an integer (rounding down).
    # 4. int(...) + 1: Adds 1 to ensure all columns are included in the last slice.
    # 5. range(...): Creates a sequence of numbers from 0 to the calculated number of slices.
    # 6. _ * COMBINER_COLUMN_SLICE_WIDTH: Multiplies each number by the slice width to get its thresholds.
    slice_thresholds = [
```

Adding comments to make this block easier to comprehend, very elegant btw!

data_validation/combiner.py
```diff
@@ -34,27 +36,106 @@
...
+        result_df.target_table_name.fillna(
+            first.get_table_name(consts.RESULT_TYPE_TARGET), inplace=True
+        )
+
+    _get_summary(run_metadata, result_df, source_df, target_df)
+
+    return result_df
+
+
+def _generate_report(
```
helensilva14:
```suggestion
def _generate_report_item(
```
Just an idea for a new name, you can use something else you feel is more appropriate.

data_validation/combiner.py
```diff
@@ -34,27 +36,106 @@
...
+    run_metadata: "RunMetadata",
+    source_df: pandas.DataFrame,
+    target_df: pandas.DataFrame,
+    join_on_fields=(),
+    is_value_comparison=False,
+    verbose=False,
+) -> pandas.DataFrame:
```
helensilva14: Same comment about the type hints as stated for the previous function.

data_validation/combiner.py
```diff
@@ -86,18 +166,18 @@ def generate_report(
...
         target, join_on_fields, run_metadata.validations, consts.RESULT_TYPE_TARGET
     )
-    target_df = client.execute(target_pivot)
+    target_pivot_df = client.execute(target_pivot)
 
     con = ibis.pandas.connect(
         {
-            consts.RESULT_TYPE_SOURCE: source_df,
+            consts.RESULT_TYPE_SOURCE: source_pivot_df,
             "differences": differences_df,
```
helensilva14: At this point I think we should create a const variable for the `"differences"` string too.

tests/unit/test_combiner.py
```diff
@@ -55,49 +54,86 @@ def module_under_test():
...
+        start_time=datetime.datetime(1998, 9, 4, 7, 30, 1),
+        end_time=None,
+        labels=[],
+        run_id="test-run",
+    )
+
+    report = module_under_test.generate_report(
+        run_metadata,
+        input_df,
+        input_df,
```
helensilva14:
```suggestion
        input_df, # mock for source_df
        input_df, # mock for target_df
```

helensilva14 (commented):
Thank you for the very detailed description of all your attempts! In my opinion everything you did here for option 4 is way more valuable than option 2 and we should move forward with it, so I added some comments/suggestions!

data_validation/combiner.py
nj1973: You are right to flag this inconsistency. Because we need to import pandas for `pandas.concat` we now have the package imported and can use that instead of having a second import for typing only. So I was thinking to replace all `"DataFrame"` with `pandas.DataFrame` but then forgot to come back to it.

I don't suppose it really matters if we have a second import really. I'll leave this comment open for now and come back to it.

data_validation/combiner.py
nj1973: Agreed

data_validation/combiner.py
nj1973: I left this name the same as before I started making changes but did rename the sub-function.

Is that any better or would still recommend renaming the top level function?

helensilva14: Yep, changing only the sub-function name already makes a difference to distinguish them, seems great to me! I'll resolve this comment.

helensilva14:
@nj1973 build failed after 13m, could you please take a look? thanks!

helensilva14 (approved):

data_validation/combiner.py
```diff
@@ -34,27 +36,110 @@
...
+    # Process the input Dataframes in slices of columns to avoid "RecursionError"s.
+    for slice_start in slice_thresholds:
+        columns_in_vertical_slice = list(validation_columns)[
+            slice_start : slice_start + COMBINER_COLUMN_SLICE_WIDTH
+        ]
+        # Ensure any join columns are included in the column slice.
+        columns_in_vertical_slice.extend(
+            set(join_on_fields) - set(columns_in_vertical_slice)
+        )
+        # Ensure join_on_fields columns are in the slice.
```
sundar-mudupalli-work: Repeated comment - delete this perhaps

data_validation/combiner.py
```diff
@@ -34,27 +36,110 @@
...
+            target_df[columns_in_vertical_slice],
+            join_on_fields=join_on_fields,
+            is_value_comparison=is_value_comparison,
+            verbose=verbose,
+        )
+        if result_df is None:
+            result_df = interim_result_df
+        else:
+            result_df = pandas.concat([result_df, interim_result_df])
+
```
sundar-mudupalli-work, line 112: Perhaps a pythonic way of doing the same
```
result_df = pandas.connect([result_df, interim_result_df]) if result_df else interim_result_df
```

data_validation/data_validation.py
```diff
@@ -329,67 +320,48 @@ def _execute_validation(self, validation_builder, process_in_memory=True):
             )
         )
 
-        if process_in_memory:
```
sundar-mudupalli-work, line 332: It appears that we have made `process_in_memory` as always true. I thought the other code was dead already, so this is likely not a change.

sundar-mudupalli-work (approved):
Overall LGTM. It may make sense to make some minor changes suggested. Have posed a question to MC Hilton for his input on `process_in_memory` flag.

data_validation/data_validation.py
renzokuken: the origin of this flag was from when we wanted to swap between a local execution backend and bigquery for the combiner. Probably no harm in removing but might want to keep if alternative backends are added in the future

data_validation/combiner.py
nj1973: Good spot, thanks

data_validation/combiner.py
nj1973: I've decided to leave this. I specifically need `is None` in there because a Dataframe cannot be used in truthy logic and I prefer the readability in this specific case. Usually I would agree though. Thanks

data_validation/data_validation.py
nj1973: It's interesting to know why the variable was there, thanks.

I think it's better to reinstate the variable if/when we decide to add the feature in the future and not maintain the (admittedly small amount of code) code until that point.

Thanks for checking it.

PR #1469, opened 2025-03-11
feat: Add raw_column_metadata to Oracle, Teradata and PostgreSQL Backend objects

### LINKED ISSUE #1468: Require framework for validations to access raw metadata
There are a few DVT issues that require us to differentiate between column baed on their raw data type and not just their Ibis database:

- https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1256
- https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1338
- https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1364
- https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1368

This issue is to put the foundation in place so that we can get the raw system data type information when required.

nj1973:
## Description of changes
Adds methods to Oracle, Teradata and PostgreSQL backends to retrieve raw data type information, plus tests.
Also adds methods to ConfigManager to retrieve these types and store them in state.

## Issues to be closed

Closes #1468

## Checklist

### DIFF
--- a/data_validation/config_manager.py
+++ b/data_validation/config_manager.py
@@ -16,7 +16,7 @@ import copy
 import logging
 import string
 import random
-from typing import TYPE_CHECKING, Dict, Iterable, List, Optional, Union, Tuple
+from typing import TYPE_CHECKING, Dict, List, Optional, Union, Tuple
 
 import google.oauth2.service_account
 import ibis.expr.datatypes as dt
--- a/data_validation/config_manager.py
+++ b/data_validation/config_manager.py
@@ -429,6 +429,7 @@ class ConfigManager(object):
             self._source_ibis_table = clients.get_ibis_table(
                 self.source_client, self.source_schema, self.source_table
             )
+        return self._source_ibis_table
 
     def get_source_ibis_table_from_query(self):
         """Return IbisTable from source."""
--- a/data_validation/config_manager.py
+++ b/data_validation/config_manager.py
@@ -16,7 +16,7 @@ import copy
 import logging
 import string
 import random
-from typing import TYPE_CHECKING, Optional, Union, List, Dict
+from typing import TYPE_CHECKING, Dict, Iterable, List, Optional, Union, Tuple
 
 import google.oauth2.service_account
 import ibis.expr.datatypes as dt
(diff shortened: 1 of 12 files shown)


tests/unit/ibis_oracle/__init__.py
sundar-mudupalli-work, line 1: Why do we have this file?

tests/unit/ibis_postgres/__init__.py
sundar-mudupalli-work, line 1: Why do we have this file?

tests/unit/ibis_teradata/__init__.py
sundar-mudupalli-work, line 1: Why do we have this file?

data_validation/config_manager.py
sundar-mudupalli-work, line 1: Neil,

Overall I don't see many issues. It all looks good. One question I have is that configManager is set up to have very few top level properties and everything is stored in the `_config` property. Is that good practice? Should `_source_raw_data_types` and `_target_raw_data_types` be stored in the `_config` property? Not sure.

tests/unit/ibis_oracle/__init__.py
nj1973: My understanding is that we need these files for pytest to import the tests. I tried removing them and get an error.

The reason is described here: https://stackoverflow.com/questions/53918088/import-file-mismatch-in-pytest/76820000#76820000

sundar-mudupalli-work (approved):
Neil,

This is awesome and such a quick turnaround. Thank you so much.
Perhaps we should change the way we use pytest to importlib mode - something to look at in the future.

Sundar Mudupalli

data_validation/config_manager.py
nj1973: It feels right to be to store them in state as private attributes of the class. It's just about caching to avoid calling the Backend method multiple times. For example if we call `get_source_raw_data_types()` two times during a validation we don't want to call `source_client.raw_column_metadata()` twice, therefore I cache the values in state in private variables and the second call would simple return the private attribute.

No-one should interact with `_source_raw_data_types` and `_target_raw_data_types` directly, hence the leading underscore.

nj1973: We could always remove these methods and variables for this PR if you prefer and you can add in what feels right when working on the char issue?

I added them to ConfigManager because that's where I thought I needed them for the CLOB/JSON work but I might be wrong. Happy to remove them if preferred.

Issue #1470, opened 2025-03-13
Fixed length character columns as primary keys not handled correctly in SQL Server, Hive, Impala and RedShift


sundar-mudupalli-work:
**Describe the bug**
Fixed length character columns as part of primary keys are not handled correctly in Hive, Impala and Redshift

**Why is this bug happening**
DVT correctly handles fixed length character columns in many database as stated in [PR 1472](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1472). 

The methods `raw_column_metadata` and `is_char_type_padded` are not yet implemented for Microsoft SQL Server, Hive, Impala and Redshift, therefore fixed length character columns that are primary keys are not trimmed.

helensilva14:
SQL Server part might be solved by PR #1530

nj1973:
Removed Db2 from the description because it is being fixed as part of https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1638

PR #1472, opened 2025-03-14
fix: Only strip primary keys when they are fixed chars, not for varchars

### LINKED ISSUE #1368: Inclusion of tsp flag leads to incorrect DVT output results
Overview: the -tsp flag was implemented as an optional flag due to a bug in teradatasql that adds extra whitespaces to CHAR data types, resulting in incorrect DVT results.

Example: ‘ABC       ’  CHAR(10) in TD and “ABC‘ in BQ will become identical when DVT executes with -tsp

We understand this raises a risk of duplicate primary keys in the instance where a table contains two different CHAR or VARCHAR PKs, which only differ by trailing spaces.

Observation: -tsp flag applied to CHAR/VARCHAR data results in incorrect DVT output results, even after accounting for the expected duplicate rows. We see identical hashes flagged as "fail" and different hashes flagged as "success"

### LINKED ISSUE #1422: DVT should only use RTRIM/Rstrip where it is needed
Hi,

DVT assumes every string is a variable length string for comparison. In order to make this work, DVT trims trailing blanks of every string, even strings that DVT generates by casting numerics as strings.

From a database perspective, trailing blanks are semantically insignificant only in fixed length character strings. Trailing blanks *are semantically significant* in variable length character strings. By obscuring this distinction, DVT is doing validation incorrectly. I made the situation worse by recommending that we not only trim blanks, but also all whitespace characters.

As outlined in [this doc on string handling](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/issue-1368-fixed-length-pks/docs/internal/strings.md), trailing blanks must be trimmed only from fixed length strings. Fortunately, there is database functionality to do this correctly in most databases - see doc for details.

I suggest we remove all RTRIM and RSTRIP except where explicitly needed. When numerics are cast as strings, they should not generate strings with trailing blanks.

Thanks.

Sundar Mudupalli

### LINKED ISSUE #1471: Random rows does not work when primary key is a fixed char column
**Describe the bug**
Random rows does not work when primary key is a fixed char column.

**What is the root cause of the problem?**
Random rows derives keys by doing a query. When that happens and the primary key is a fixed char column, the random keys include trailing blanks. With the [fix for 1368](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1368), keys with trailing blanks are not acceptable.

The correct fix is for Random rows to use `get_...__raw_data_types` function in the config manager, see if the specific column needs to be trimmed or not.

Sundar Mudupalli

sundar-mudupalli-work:
## Description of changes
DVT handles fixed length character strings as primary keys without the use of `-tsp` option - for the following databases:
1. Oracle
2. Postgres
3. Teradata

Following backends do not pad characters so no change needed to work correctly:
5. BigQuery
6. Spanner
7. Snowflake
8. MySQL (stripped on retrieval automatically)

The following backends have not implemented `raw_column_metadata` and `is_char_type_padded` methods - DB2, Microsoft SQL Server, Hive, Impala and Redshift. See [issue 1470](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1470).

If the primary key on Random rows is a fixed length character type, we strip the value before using it. Therefore this fixes #1471.

This fixes #1368 

This partially fixes #1422 - because it only addresses fixed length character type for primary keys. It does not address the comparison of a column with fixed length in Oracle to a String column in BigQuery. In that situation, the Oracle column must be stripped, while the BigQuery  column must not be stripped.

This change also deprecates the `-tsp` option - it is removed from the documentation and users get a warning.

## Issues to be closed
Closes #1368
Closes #1471  

## Checklist


data_validation/validation_builder.py
```diff
@@ -87,6 +87,41 @@ def _construct_isin_filter(
...
+        # If we don't have metadata don't trim.
+        if not raw_column_metadata:
+            return False
+        match client:
+            # The following backends don't pad blanks - either don't support fixed char or
+            # trim spaces before using.
+            case "bigquery" | "spanner" | "snowflake" | "mysql":
+                return False
+            case "teradata":
+                return raw_column_metadata[column_name][0] == "CHAR"
```
nj1973: I don't think we should have these system specific data type names listed here.

We could add a method to each Backend called `is_fixed_length_char(raw_type_name: str): -> bool` and replace all of this with a single `return client.is_fixed_length_char(raw_column_metadata[column_name][0])`.

BigQuery/Spanner/etc client would just return False and not check the data type.

data_validation/validation_builder.py
```diff
@@ -87,6 +87,41 @@ def _construct_isin_filter(
...
+        raw_column_metadata: dict,
+        column_name: str,
+    ) -> bool:
+        """Returns true if the string column provided is a fixed length char column.
+        Since some databases automatically pad fixed length chars when storing them and returning the value, these
+        will need to be trimmed before comparison. Some databases - e.g. BQ do not support
+        fixed length character strings
+        """
+        # If we don't have metadata don't trim.
+        if not raw_column_metadata:
```
nj1973: I wonder if we should also test that the column name is in `raw_column_metadata` here? Something like:
```
if not raw_column_metadata or column_name not in raw_column_metadata:
```

docs/internal/strings.md
```diff
@@ -0,0 +1,50 @@
...
+
+For databases where the encoding can only be set at the database or server level (Oracle, DB2 and Postgres) we need to test with a database with a different encoding (e.g. ISO-8859-1), store Latin characters in it and compare it against the same strings stored in BQ. A separate [issue](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1423) has been opened.
+
+For Hive, MySQl and MS SQL the encoding can be set at the table level, create a table with ISO-8859-1 encoding, store Latin characters in it and compare against the same strings stored in BQ. A separate [issue](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1424) has been opened.
+
+DVT correctly handles Teradata strings encoded in ISO-8859 (Latin) encoding [PR 1226](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1226)
+
+## Handling NCHAR and NVARCHAR data types in MS SQL and Oracle
```
nj1973: Should we use "SQL Server" instead of "MS SQL". We all know what "MS SQL" means but I don't think it is an official product name.

docs/internal/strings.md
```diff
@@ -0,0 +1,50 @@
...
+
+Customer has opened an [issue](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1406) that tables with NCHAR and NVARCHAR cannot be validated in Oracle. It is possible a similar issue also exists in SQL Server. By definition, NCHAR and NVARCHAR use Unicode characters and in theory is a superset of the encoding used for CHAR and VARCHAR. In practice, current versions of Oracle and SQL Server use UTF8 encoding for CHAR and VARCHAR by default. Customers upgrading from earlier versions may be using a more limited encoding such as ASCII7 or Latin-1.
+
+Since the character set for NCHAR and NVARCHAR is guaranteed to be a superset of the character set for CHAR and VARCHAR, DVT can convert all string fields to NVARCHAR. Please note comments earlier in this document that fixed length strings for Oracle and SQL Server must be blank trimmed prior to processing. Once converted to variable length character strings in Unicode, DVT can perform the comparison with other character strings represented in UTF-8.
```
nj1973: I'm not sure we should include this statement. It is preempting a change we may or may not make in the future.

tests/resources/bigquery_test_tables.sql
```diff
@@ -141,6 +140,31 @@ INSERT INTO `pso_data_validator`.`dvt_char_id` VALUES
 ('DVT4  ', 'Row 4  	  '),
 ('DVT5  ', 'Row 5');
 
+# BigQuery does not have a fixed char data type and blanks at the end of strings are not removed by the database.
+# Other databases have a fixed char data type and these tables are to allow the comparison of those columns with BQ and each other
+CREATE OR REPLACE TABLE `pso_data_validator`.`dvt_varchar_id`
```
nj1973, line 181: We already have test table `dvt_string_id`. Do we either not need `dvt_varchar_id ` or need to remove `dvt_string_id`?

tests/resources/mysql_test_tables.sql
```diff
@@ -134,6 +134,30 @@ INSERT INTO `pso_data_validator`.`dvt_datetime_id` VALUES
...
+,   other_data  varchar(100)
+) OPTIONS (description='Integration test table used to test variable length string, trailing blanks are significant');
+INSERT INTO `pso_data_validator`.`dvt_varchar_id` VALUES
+('DVT-key-1', 'Row 1'),
+('DVT-key-2', 'Row 2'),
+('DVT-key-3', 'Row 3'),
+('DVT-key-4 ', 'Row 4'),
+('DVT-key-5', 'Row 5');
+
+DROP TABLE IF EXISTS `pso_data_validator`.`dvt_fixed_char_id`
```
nj1973, line 182: We already have `dvt_char_id` for several systems, this new table is a duplicate.

tests/resources/oracle_test_tables.sql
```diff
@@ -258,6 +258,32 @@ INSERT INTO pso_data_validator.dvt_char_id VALUES ('DVT4', 'Row 4');
 INSERT INTO pso_data_validator.dvt_char_id VALUES ('DVT5', 'Row 5');
 COMMIT;
 
+DROP TABLE pso_data_validator.dvt_fixed_char_id;
```
nj1973, line 283: Same as for BigQuery, the new tables appear to be duplicates of existing tables.

nj1973 (changes requested):
A few suggestions

docs/internal/strings.md
sundar-mudupalli-work: Soften this a bit. It is a design document, so wanted to layout one way to address the problem.

tests/resources/bigquery_test_tables.sql
sundar-mudupalli-work: For me the tables `dvt_string_id` and `dvt_char_id` were confusing - not clear about the purpose of each. So I created new tables with names that are pretty clear. Also the data in those tables assumed you were stripping string primary keys. The data in the `dvt_varchar_id` includes a row with a primary key with a trailing blank. This would have been stripped in the earlier design. Now it will not.

Yes - maybe in a few months, we can remove the tables `dvt_string_id` and `dvt_char_id` from our databases.

Thank you.

Sundar Mudupalli

data_validation/validation_builder.py
sundar-mudupalli-work: This is a very helpful comment. I have adopted your suggestion. Please review.

tests/resources/bigquery_test_tables.sql
nj1973: You've removed the tests that use `dvt_string_id` so I assumed the new table was a replacement for the old one. If they serve different purposes should the old tests be left in place? If the table is redundant we can remove it from the source code but not actually drop it until your changes reach `develop`.

helensilva14:
Hi team! I was wondering if issue #1422 could also be closed by this PR?

data_validation/data_validation.py
```diff
@@ -162,8 +162,20 @@ def _add_random_row_filter(self):
                 **{source_pk_column: query[source_pk_column].cast("string")}
             )
 
-        if self.config_manager.trim_string_pks():
-            query = query.mutate(**{source_pk_column: query[source_pk_column].rstrip()})
+        # If the primary key is a padded string, then update the query to ensure the string
+        # is rstripped in the results.
+        if self.config_manager.source_table:  # Not a custom query
```
nj1973: Why can this not also work for custom query?

data_validation/partition_builder.py
```diff
@@ -195,14 +195,21 @@ def _get_partition_key_filters(self) -> List[List[List[str]]]:
             window1 = ibis.window(order_by=source_pks)
             row_number = (ibis.row_number().over(window1) + 1).name(consts.DVT_POS_COL)
 
-            if config_manager.trim_string_pks():
+            # If any of the keys are padded strings, we need to rstrip them so the values come out right
+            if config_manager.source_table:  # Not a custom query
```
nj1973: Repeat comment re table vs custom query

data_validation/validation_builder.py
```diff
@@ -15,6 +15,7 @@
 import logging
 from typing import TYPE_CHECKING
 
+import ibis.backends.base
```
nj1973: I don't think we need this.

We already import `ibis` for type checking, see `_get_calc_type()` for usage.

Or if you prefer the way you've done it here then we should make the same change to `_get_calc_type()`.

Either way, imports for type checking only should probably be moved within the `TYPE_CHECKING` condition.

data_validation/validation_builder.py
```diff
@@ -284,14 +290,32 @@ def add_primary_key(self, primary_key):
         # grab calc field metadata
         alias = primary_key.get(consts.CONFIG_FIELD_ALIAS)
         cast = primary_key.get(consts.CONFIG_CAST)
-        trim = self.config_manager.trim_string_pks()
-        # check if valid calc field and return correct object
+        config_manager = self.config_manager
+        trim = False
+        if config_manager.source_table:  # Not a custom query
```
nj1973: Repeat comment re table vs custom query

tests/resources/bigquery_test_tables.sql
```diff
@@ -176,28 +176,30 @@ INSERT INTO `pso_data_validator`.`dvt_binary` VALUES
...
-) OPTIONS (description='Integration test table used to test CHAR pk matching.');
-INSERT INTO `pso_data_validator`.`dvt_char_id` VALUES
-('DVT1  ', 'Row 1	  '),
-('DVT2  ', 'Row 2  	'),
-('DVT3  ', 'Row 3  '),
-('DVT4  ', 'Row 4  	  '),
-('DVT5  ', 'Row 5');
+) OPTIONS (description='Integration test table used to test fixed char primary key - trailing blanks are not significant');
+INSERT INTO `pso_data_validator`.`dvt_fixed_char_id` VALUES
+('DVT1', 'Row 1	  '),
```
nj1973, line 198: Do we need the trailing spaces in the id column here to prove the rtrim works?

data_validation/validation_builder.py
```diff
@@ -284,14 +290,32 @@ def add_primary_key(self, primary_key):
         # grab calc field metadata
         alias = primary_key.get(consts.CONFIG_FIELD_ALIAS)
         cast = primary_key.get(consts.CONFIG_CAST)
-        trim = self.config_manager.trim_string_pks()
-        # check if valid calc field and return correct object
+        config_manager = self.config_manager
+        trim = False
```
nj1973: If I were migrating data from PostgreSQL CHAR to BigQuery STRING I have two options:

1) Migrate the trailing spaces too
2) Trim off the trailing spaces

The logic here will only trim the trailing spaces from PostgreSQL data, not from BigQuery. Should `trim=True` for source and target if **either** one is a padder string? That way we cater for both scenarios 1 and 2 above.

tests/system/data_sources/test_oracle.py
```diff
@@ -631,21 +631,28 @@ def test_row_validation_binary_pk_to_bigquery():
     "data_validation.state_manager.StateManager.get_connection_config",
     new=mock_get_connection_config,
 )
-def test_row_validation_string_pk_to_bigquery():
-    """Test string primary key join columns"""
+def test_row_validation_fixed_char_pk_to_bigquery():
+    """Oracle to BigQuery dvt_fixed_char_id row validation.
+    This is testing CHAR primary key join columns.
+    Random rows don't play well with fixed char pk columns see https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1471
```
nj1973: I thought this PR fixed 1471 and therefore we can allow `use_randow_row` to be `True`?

tests/system/data_sources/test_postgres.py
```diff
@@ -862,10 +862,23 @@ def test_row_validation_binary_pk_to_bigquery():
     "data_validation.state_manager.StateManager.get_connection_config",
     new=mock_get_connection_config,
 )
-def test_row_validation_char_pk_to_bigquery():
-    """Test padded char primary key join columns"""
+def test_row_validation_fixed_char_pk_to_bigquery():
+    """Test datetime primary key join columns"""
+    # TODO Remove use_randow_row option below when issue-1471 is actioned.
```
nj1973: As above, I thought this PR fixed 1471?

tests/system/data_sources/test_teradata.py
```diff
@@ -602,24 +602,28 @@ def test_row_validation_binary_pk_to_bigquery():
     "data_validation.state_manager.StateManager.get_connection_config",
     new=mock_get_connection_config,
 )
-def test_row_validation_string_pk_to_bigquery():
-    """Test string primary key join columns"""
+def test_row_validation_fixed_char_pk_to_bigquery():
+    """Oracle to BigQuery dvt_fixed_char_id row validation.
+    This is testing CHAR primary key join columns.
+    Random rows don't play well with fixed char pk columns see https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1471
```
nj1973: 1471 comment again.

tests/system/data_sources/test_teradata.py
```diff
@@ -602,24 +602,28 @@ def test_row_validation_binary_pk_to_bigquery():
...
+def test_row_validation_varchar_pk_to_bigquery():
+    """Oracle to BigQuery dvt_varchar_id row validation.
+    This is testing CHAR primary key join columns.
+    Includes random row filter test.
+    """
     id_column_row_validation_test(
-        "udf.dvt_char_id=pso_data_validator.dvt_char_id",
-        use_randow_row=False,
-        # We need to trim padded string PKs due to a Teradata client "quirk".
-        trim_string_pks=True,
```
nj1973, line 622: Should we remove `trim_string_pks` arg from the source function in `common_functions.py`?

third_party/ibis/ibis_db2/__init__.py
```diff
@@ -101,3 +101,31 @@ def list_primary_key_columns(self, database: str, table: str) -> list:
                 list_pk_col_sql, parameters=(database.upper(), table.upper())
             )
             return [_[0] for _ in result.cursor.fetchall()]
+
+    def raw_column_metadata(
+        self, database: str = None, table: str = None, query: str = None
+    ) -> Iterable[Tuple]:
+        """Define this method to allow DVT to test if backend specific transformations may be needed for comparison.
+        Partner method to _get_schema_using_query that retains raw data type information instead of converting
```
nj1973: There is no `_get_schema_using_query` in this module?

third_party/ibis/ibis_db2/__init__.py
```diff
@@ -101,3 +101,31 @@ def list_primary_key_columns(self, database: str, table: str) -> list:
...
+                                           and COLNAME='{column_name.upper()}'
+        This works for tables, but does not allow us to find the underlying SQL data type for queries. Cursor.description
+        provides the underlying "compatible" SQL data types, so a fixed char and varchar column will have the same underlying
+        SQL data types.
+
+        Returns:
+            list: A list of tuples containing the standard 7 DB API fields:
+                  https://peps.python.org/pep-0249/#description
+        """
+        return ()
```
nj1973, line 120: Should this method do the same as the `_metadata` method but with `_get_type` so we can create a list with raw data types?

We do the same in the Oracle equivalent method.

third_party/ibis/ibis_mssql/__init__.py
```diff
@@ -109,3 +109,30 @@ def list_primary_key_columns(self, database: str, table: str) -> list:
...
+                                            where TABLE_NAME='{table_name}'
+                                            and COLUMN_NAME='{column_name}'
+        This works for tables, but does not allow us to find the underlying SQL data type for queries. For that we need to
+        wait until this issue is resolved in https://github.com/mkleehammer/pyodbc/issues/167
+
+        Returns:
+            list: A list of tuples containing the standard 7 DB API fields:
+                  https://peps.python.org/pep-0249/#description
+        """
+        return ()
```
nj1973, line 128: Should this method do the same as the `_metadata` method but with `_type_from_result_set_info` so we can create a list with raw data types?

We do the same in the Oracle equivalent method.

third_party/ibis/ibis_mssql/__init__.py
```diff
@@ -109,3 +109,30 @@ def list_primary_key_columns(self, database: str, table: str) -> list:
         with self.begin() as con:
             result = con.exec_driver_sql(list_pk_col_sql, parameters=(database, table))
             return [_[0] for _ in result.cursor.fetchall()]
+
+    def raw_column_metadata(
+        self, database: str = None, table: str = None, query: str = None
+    ) -> Iterable[Tuple]:
```
nj1973: We have some undefined type hints here that need fixing.

third_party/ibis/ibis_mssql/__init__.py
```diff
@@ -109,3 +109,30 @@ def list_primary_key_columns(self, database: str, table: str) -> list:
...
+        This works for tables, but does not allow us to find the underlying SQL data type for queries. For that we need to
+        wait until this issue is resolved in https://github.com/mkleehammer/pyodbc/issues/167
+
+        Returns:
+            list: A list of tuples containing the standard 7 DB API fields:
+                  https://peps.python.org/pep-0249/#description
+        """
+        return ()
+
+    def is_char_type_padded(self, char_type: Tuple) -> bool:
```
nj1973, line 130: `Tuple` is an undefined type hints.

tests/resources/sqlserver_test_tables.sql
```diff
@@ -174,6 +174,30 @@ INSERT INTO pso_data_validator.dvt_binary VALUES (CAST('DVT-key-3' AS binary), 3
...
+INSERT INTO pso_data_validator.dvt_fixed_char_id VALUES ('DVT4', 'Row 4  	  ');
+INSERT INTO pso_data_validator.dvt_fixed_char_id VALUES ('DVT5', 'Row 5');
+
+DROP TABLE pso_data_validator.dvt_varchar_id;
+CREATE TABLE pso_data_validator.dvt_varchar_id
+(   id          VARCHAR(15) NOT NULL PRIMARY KEY
+,   other_data  VARCHAR(100)
+);
+execute sp_addextendedproperty 'Comment', 'Integration test table used to test varchar pk matching. Trailing blanks are significant', 'SCHEMA', 'pso_data_validator', 'table', 'dvt_varchar_id';
+INSERT INTO pso_data_validator.dvt_varchar_id VALUES ('DVT-key-1', 'Row 1');
```
nj1973, line 195: We are missing tests for this table.

tests/resources/mysql_test_tables.sql
```diff
@@ -176,6 +167,30 @@ INSERT INTO `pso_data_validator`.`dvt_datetime_id` VALUES
...
+CREATE TABLE `pso_data_validator`.`dvt_fixed_char_id`
```
nj1973, line 183: We are missing tests for this table.

tests/resources/mysql_test_tables.sql
```diff
@@ -176,6 +167,30 @@ INSERT INTO `pso_data_validator`.`dvt_datetime_id` VALUES
 ('2020-01-01 12:00:00', 'Row 1'), ('2020-02-01 12:00:00', 'Row 2'), ('2020-03-01 12:00:00', 'Row 3'),
 ('2020-04-01 12:00:00', 'Row 4'), ('2020-05-01 12:00:00', 'Row 5');
 
+DROP TABLE IF EXISTS `pso_data_validator`.`dvt_varchar_id`
```
nj1973, line 170: We are missing tests for this table.

tests/resources/sqlserver_test_tables.sql
```diff
@@ -174,6 +174,30 @@ INSERT INTO pso_data_validator.dvt_binary VALUES (CAST('DVT-key-3' AS binary), 3
 INSERT INTO pso_data_validator.dvt_binary VALUES (CAST('DVT-key-4' AS binary), 4, 'Row 4');
 INSERT INTO pso_data_validator.dvt_binary VALUES (CAST('DVT-key-5' AS binary), 5, 'Row 5');
 
+DROP TABLE pso_data_validator.dvt_fixed_char_id;
+CREATE TABLE pso_data_validator.dvt_fixed_char_id
```
nj1973, line 178: We are missing tests for this table.

docs/internal/strings.md
```diff
@@ -0,0 +1,50 @@
...
+
+DVT (and Python) default to using variable length character strings encoded in UTF8. When comparisons are made, DVT generates SQL for the source and target databases to convert their strings to UTF8. Depending on the particular data types (fixed vs variable length), the SQL uses cast and trim operators to ensure the output UTF8 bytes are equal when the source and target strings are the same.
+
+## Data types support and behavior by database
+
+| Database | Primary String data types | Secondary (encoding) data types | Comments |
+| --- | --- | --- | --- |
+| AlloyDB | See Postgres | See Postgres | See Postgres |
+| BigQuery | STRING | None | All stored in UTF8 |
+| Db2 | CHAR, VARCHAR | | CHAR datatype is padded; Encoding specified per database |
```
nj1973, line 17: Do we need test tables and tests for this Db2 datatype?

nj1973 (changes requested):
Mostly questions but a few recommendations too.

data_validation/data_validation.py
sundar-mudupalli-work: Neil,

Code has been updated so it assumes that the backend will be able to provide `raw_column_metadata` whether it is a custom query or a table. How to accomplish this is different for different backends. Some backends, the python driver alone is sufficient. In others, we have to execute stored procedures or other means.

Thank you - I learned something today.

Sundar Mudupalli

data_validation/validation_builder.py
sundar-mudupalli-work: Thank you.

data_validation/validation_builder.py
sundar-mudupalli-work: This is correct. Trailing spaces have no semantic significance in Postgres (fixed length) CHAR type, they are semantically significant in BigQuery (variable length) STRING type. In our comparisons, we trim all fixed length strings  and compare the result with variable length strings.

tests/system/data_sources/test_teradata.py
sundar-mudupalli-work: Thank you. Done

tests/system/data_sources/test_postgres.py
sundar-mudupalli-work: Fixed. Thank you

tests/system/data_sources/test_teradata.py
sundar-mudupalli-work: Fixed. Thank you.

tests/resources/bigquery_test_tables.sql
sundar-mudupalli-work: No. Trailing spaces are semantically significant in the STRING data type in BigQuery. We do not trim the id columns in Big Query, only in databases where trailing spaces are not semantically significant. We have a row with the id `DVT-key-4 ` (with a trailing space) in the `dvt_varchar_id` table test this specifically.

data_validation/validation_builder.py
sundar-mudupalli-work: see https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1472#discussion_r2112641101

tests/resources/mysql_test_tables.sql
sundar-mudupalli-work: Done. Thank you.

tests/resources/mysql_test_tables.sql
sundar-mudupalli-work: Done. Thank you.

third_party/ibis/ibis_mssql/__init__.py
sundar-mudupalli-work: Fixed. Thank you

tests/resources/sqlserver_test_tables.sql
sundar-mudupalli-work: I have not completed the work for SQL Server and marked that as part of [issue 1470](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1470). I want to get a release out now - have lot of smaller, simpler changes rather than one big change.

tests/resources/sqlserver_test_tables.sql
sundar-mudupalli-work: Explained earlier.

docs/internal/strings.md
sundar-mudupalli-work: Yes, we should - that is part of [issue 1470](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1470) now

tests/system/data_sources/test_oracle.py
sundar-mudupalli-work: Yes, this is fixed and part of the test suite. Required some additional work for Oracle - you will see that in the code, please provide your review.

third_party/ibis/ibis_db2/__init__.py
sundar-mudupalli-work: Postponed to resolution during [issue 1470](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1470)

third_party/ibis/ibis_db2/__init__.py
sundar-mudupalli-work: Yes. Postponed to [issue 1470](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1470)

third_party/ibis/ibis_mssql/__init__.py
sundar-mudupalli-work: [Issue 1470](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1470)

third_party/ibis/ibis_mssql/__init__.py
sundar-mudupalli-work: Fixed.

tests/system/data_sources/common_functions.py
```diff
@@ -442,14 +441,63 @@ def id_column_row_validation_test(
...
+        "--primary-keys=id",
+        col_option,
+    ]
+    cli_arg_list = [_ for _ in cli_arg_list if _]
+    args = parser.parse_args(cli_arg_list)
+    df = run_test_from_cli_args(args)
+    id_type_test_assertions(df)
+
+
+def fixed_char_varchar_test(
```
nj1973, line 480: I think we've discussed this before so this is just me having an opinion and not a request. Feel free to ignore it.

In my opinion we gain nothing by bundling 4 tests into a single test but we do lose a bit of clarity. If the test fails, how do we know if it relates to fixed or variable length data types or table centric vs custom query code? Like I said, not a big deal but just my thoughts.

third_party/ibis/ibis_mssql/__init__.py
```diff
@@ -110,6 +110,28 @@ def list_primary_key_columns(self, database: str, table: str) -> list:
             result = con.exec_driver_sql(list_pk_col_sql, parameters=(database, table))
             return [_[0] for _ in result.cursor.fetchall()]
 
+    def raw_column_metadata_not_implemented(
+        self, database: str = None, table: str = None, query: str = None
+    ) -> List[Tuple]:
```
nj1973, line 115: As far as I can see `List` is still undefined in this module.

nj1973 (approved):
Just a couple of optional comments.

Good work. Thanks.

third_party/ibis/ibis_mssql/__init__.py
sundar-mudupalli-work: Thank you. Fixed.

tests/system/data_sources/common_functions.py
sundar-mudupalli-work: Neil,

I am not sure I understood it the first time, now I do. I was thinking I was saving a few lines by not repeating the mock.patch, not realizing I made failure detection more complicated. I have reversed it now for this PR.

I have opened an issue that we should look at our tests and use fixtures by module/package for database connections. When that is done most of the mock.patch for get_connection_config will be unnecessary. I have opened an [Issue here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1527)

Sundar Mudupalli

PR #1488, opened 2025-03-31
fix: NaT handling in string_to_epoch now matches datetime64 cast output

### LINKED ISSUE #1473: TypeError exception from column validation: NAType has no len()
Specific set of circumstances to cause exception:
```
  File "/./python3.10/site-packages/third_party/ibis/ibis_addon/operations.py", line 537, in string_to_epoch
    parsed_ts = dateutil.parser.isoparse(ts).astimezone(dateutil.tz.UTC)
  File "/./python3.10/site-packages/dateutil/parser/isoparser.py", line 37, in func
    return f(self, str_in, *args, **kwargs)
  File "/./python3.10/site-packages/dateutil/parser/isoparser.py", line 134, in isoparse
    components, pos = self._parse_isodate(dt_str)
  File "/./python3.10/site-packages/dateutil/parser/isoparser.py", line 208, in _parse_isodate
    return self._parse_isodate_common(dt_str)
  File "/./python3.10/site-packages/dateutil/parser/isoparser.py", line 213, in _parse_isodate_common
    len_str = len(dt_str)
TypeError: object of type 'NAType' has no len()
```

This error happens in a where all of the following conditions are satisfied at the same time:

- the validation includes the option --grouped_columns
- the validation includes the option --wildcard-include-timestamp
- the table being validated includes at least one timestamp column that is nullable and that does not appear in the `--grouped-columns` option, so that its values are being aggregated
- the table includes at least one group of rows, where some aggregation functions return results for other columns, but where the value of the timestamp column is null for all the rows in the group.

In that case the aggregated timestamp value of the nullable timestamp column can be null and DVT is not able to
derive an epoch seconds value.

nj1973:
## Description of changes
When comparing min/max of datetimes DVT changes them to epoch seconds. There were a couple of flaws here that are fixed by this PR:

1. NaT would throw an exception when it appeared in the same Series as a date that overflows datetime64 (i.e. an old date or a far future date)
2. The NaT value for datetime64 was being cast to int32 which caused another overflow. It is now cast to int64.

## Issues to be closed

Closes #1473

## Checklist


helensilva14 (approved):
Well done!

PR #1499, opened 2025-04-09
fix: Issue casting SQL Server DECIMAL with scale > 0 to string

helensilva14:
## Description of changes

- Adds condition to function `sa_cast_mssql` on `third_party/ibis/ibis_addon/operations.py` to correctly handle decimal casting to string
- Updates some SQL Server system tests, specially the one where the issue was mentioned
- Lints `tests/unit/test_combiner.py`

## Issues to be closed
_Note: Before submitting a pull request, please open an issue for discussion if you are not associated with Google._

Closes #1111

## Checklist

- [x] I have followed the [`CONTRIBUTING` Guide](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/CONTRIBUTING.md).
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have updated any relevant documentation to reflect my changes, if applicable
- [x] I have added unit and/or integration tests relevant to my change as needed
- [x] I have already checked locally that all unit tests and linting are passing (use the `tests/local_check.sh` script)
- [x] I have manually executed end-to-end testing (E2E) with the affected databases/engines

PR #1521, opened 2025-05-22
fix: Fix `find-tables` execution for SQL Server to properly filter by schema

### LINKED ISSUE #1079: validate row: Failing hash validation on SQL Server decimal(38,30) column
As part of issue https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/956 I added a new integration test that tests large decimal primary keys and included large decimal columns in the hash comparison too.

On SQL Server the new test is failing.

When working on this issue you can search for "issue-1079" in `tests/system/data_sources/test_sql_server.py` and add the column as the `TODO` comment describes.

### LINKED ISSUE #1198: find-tables is incorrect for SQL Server
On SQL Server the code in clients.py->list_tables() is not returning correct results.

This is the current code:
```
def list_tables(client, schema_name):
    """Return a list of tables in the DB schema."""
    if client.name in ["db2", "mssql", "redshift", "snowflake"]:
        return client.list_tables()
    return client.list_tables(database=schema_name)
```

I tested this with `schema_name=pso_data_validator` and reviewed the list of tables:

```
(Pdb) p client.list_tables()
['entries']
(Pdb) p schema_name
'pso_data_validator'
(Pdb) p client.list_tables(database=schema_name)
['dvt_binary', 'dvt_core_types', 'dvt_large_decimals', 'dvt_mssql2pg_types', 'dvt_null_not_null', 'dvt_pangrams']
```

Note that calling `client.list_tables()` with no schema is giving me a list of table names in the `dbo` schema. But calling it with the correct schema gives the correct results.

```
1> SELECT table_name FROM information_schema.tables WHERE table_schema = 'dbo';
2> go
table_name
--------------------------------------------------------------------------------------------------------------------------------
entries

1> SELECT table_name FROM information_schema.tables WHERE table_schema = 'pso_data_validator';
2> go
table_name
--------------------------------------------------------------------------------------------------------------------------------
dvt_null_not_null
dvt_mssql2pg_types
tblIssue990
dvt_large_decimals
dvt_binary
dvt_pangrams
dvt_core_types
```

I think we need to remove `"mssql"` from the list of clients referenced in `list_tables()`.

When this has been resolved we can enable the `test_find_tables()` test in `test_sql_server.py`.

See also: https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1043

helensilva14:
## Description of changes

- Fix DDL and DML for `test_generate_partitions_v2` table on SQL Server
- Changes on `tests/system/data_sources/test_sql_server.py`:
   - Unskip `test_find_tables`
   - Add `col_dec_38_30` to hash string in `test_column_validation_large_decimals_to_bigquery` and `test_row_validation_large_decimals_to_bigquery` as issue #1079 ended up being solved by PR #1499 
   - Add allow list types for `test_schema_validation_core_types_to_bigquery` and `test_schema_validation_not_null_vs_nullable`
- Add functions on `third_party/ibis/ibis_mssql/__init__.py` to fix `find_tables` execution 

## Issues to be closed

Closes #1198
Closes #1079 

## Checklist

### DIFF
--- a/data_validation/schema_validation.py
+++ b/data_validation/schema_validation.py
@@ -237,8 +237,10 @@ def split_allow_list_str(allow_list_str: str) -> list:
     # I've not moved this patter to a compiled constant because it should only
     # happen once per command and I felt splitting the pattern into variables
     # aided readability.
-    nullable_pattern = r"!?"
-    precision_scale_pattern = r"(?:\((?:[0-9 ,\-]+|'UTC')\))?"
+    nullable_pattern = r"!?"  # Matches an optional '!'
+    # Matches precision/scale like (N), (N,M), ('UTC'), or (N, 'UTC')
+    # Crucially, [0-9 ,-]+ allows for digits, spaces, commas, and hyphens (for ranges like "1-18").
+    precision_scale_pattern = r"(?:\((?:[0-9 ,-]+(?:,[ ]*'UTC')?|'UTC')\))?"
     data_type_pattern = nullable_pattern + r"[a-z0-9 ]+" + precision_scale_pattern
     csv_split_pattern = data_type_pattern + r":" + data_type_pattern
     data_type_pairs = [
--- a/third_party/ibis/ibis_mssql/__init__.py
+++ b/third_party/ibis/ibis_mssql/__init__.py
@@ -15,6 +15,9 @@ from __future__ import annotations
 
 from typing import Literal
 
+import pandas
+import warnings
+
 import sqlalchemy as sa
 from ibis.backends.base.sql.alchemy import BaseAlchemyBackend
 from ibis.backends.mssql.compiler import MsSqlCompiler
@@ -30,6 +33,7 @@ class Backend(BaseAlchemyBackend):
     name = "mssql"
     compiler = MsSqlCompiler
     supports_create_or_replace = False
+    LOCK_SQL = "WITH (ROWLOCK)"
 
     _sqlglot_dialect = "tsql"
 
@@ -81,6 +85,7 @@ class Backend(BaseAlchemyBackend):
             with dbapi_connection.cursor() as cur:
                 cur.execute("SET DATEFIRST 1")
 
+        self.client = engine
         return super().do_connect(engine)
 
     def _metadata(self, query):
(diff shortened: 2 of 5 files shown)


third_party/ibis/ibis_mssql/__init__.py
```diff
@@ -109,3 +114,74 @@ def list_primary_key_columns(self, database: str, table: str) -> list:
         with self.begin() as con:
             result = con.exec_driver_sql(list_pk_col_sql, parameters=(database, table))
             return [_[0] for _ in result.cursor.fetchall()]
+
+    LIST_DATABASE_SQL = """
+        SELECT schema_name FROM information_schema.schemata
+        WHERE schema_name LIKE '%{schema_like}%'
```
nj1973: If possible I think we should use `?` bind variables for SQL parameters, like in `list_primary_key_columns`

This protects against SQL injection.

third_party/ibis/ibis_mssql/__init__.py
```diff
@@ -109,3 +114,74 @@ def list_primary_key_columns(self, database: str, table: str) -> list:
         with self.begin() as con:
             result = con.exec_driver_sql(list_pk_col_sql, parameters=(database, table))
             return [_[0] for _ in result.cursor.fetchall()]
+
+    LIST_DATABASE_SQL = """
```
nj1973: I'm interested in why we're declaring `LIST_DATABASE_SQL` as a class level attribute and not either a constant or a local variable inside `list_databases()`?

I'm not saying this is wrong, just wondered why.

third_party/ibis/ibis_mssql/__init__.py
```diff
@@ -109,3 +114,74 @@ def list_primary_key_columns(self, database: str, table: str) -> list:
...
+
+        list_database_sql = self.LIST_DATABASE_SQL.format(schema_like=schema_like)
+        databases_df = self._execute(list_database_sql, results=True)
+
+        return list(databases_df.schema_name.str.rstrip())
+
+    LIST_TABLE_SQL = """
+        SELECT table_name
+        FROM information_schema.tables
+        WHERE table_schema LIKE '%{schema_like}%'
```
nj1973: Same bind variable comment as above.

third_party/ibis/ibis_mssql/__init__.py
```diff
@@ -109,3 +114,74 @@ def list_primary_key_columns(self, database: str, table: str) -> list:
...
+            schema_like=schema, table_like=table, type_like=type_like
+        )
+        tables_df = self._execute(list_table_sql, results=True)
+        return list(tables_df.table_name.str.rstrip())
+
+    def dvt_list_tables(self, like=None, database=None) -> list:
+        """Duplicate of list_tables() but only returning tables in the output."""
+        return self.list_tables(like=like, schema=database, type_like="BASE TABLE")
+
+    def _execute(self, sql, results=False, params=None):
```
nj1973: Why do we need a new execute function? Existing code uses something like:
```
        with self.begin() as con:
            result = con.exec_driver_sql(list_pk_col_sql, parameters=(database, table))
```

Interested in why we need a new method?

third_party/ibis/ibis_mssql/__init__.py
```diff
@@ -30,6 +33,7 @@ class Backend(BaseAlchemyBackend):
     name = "mssql"
     compiler = MsSqlCompiler
     supports_create_or_replace = False
+    LOCK_SQL = "WITH (ROWLOCK)"
```
nj1973: Does this mean we are locking tables when we select from them?

I don't understand SQL Server very well. Is it that readers lock tables and this manages it in some way?

nj1973 (changes requested):

third_party/ibis/ibis_mssql/__init__.py
helensilva14: The only reason behind all your comments actually is the fact I followed the exact same approach already implemented for Teradata. For each comment I'll bring the snippet I followed:

https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/e75816b043cb63e9706e89c22d61776b2f821081/third_party/ibis/ibis_teradata/__init__.py#L76-L87

third_party/ibis/ibis_mssql/__init__.py
helensilva14: I didn't notice we already had a code in the file to be able to execute the SELECT, I just created an equivalent to the Teradata function:

https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/e75816b043cb63e9706e89c22d61776b2f821081/third_party/ibis/ibis_teradata/__init__.py#L179-L191

third_party/ibis/ibis_mssql/__init__.py
helensilva14: Honestly I'm not sure if this is really needed but I assumed it was a safeguard worth to bring an equivalent from the Teradata code. My understanding was this will make sure only the "final" state of the rows are read instead of getting temporary or uncommitted changes.

https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/e75816b043cb63e9706e89c22d61776b2f821081/third_party/ibis/ibis_teradata/__init__.py#L37

helensilva14: Bringing here the documentation reference to help us understand it better: https://learn.microsoft.com/en-us/sql/t-sql/queries/hints-transact-sql-table?view=sql-server-ver17#rowlock

third_party/ibis/ibis_mssql/__init__.py
helensilva14: Changing it to be a local variable inside the function

helensilva14:
@nj1973 - please, besides the code review I need you to tell me what failed in the buid after 18 min so I can know what to fix, thank you! 🙏

helensilva14:
I found out these tests are failing, please let me know if the same happened in the build or if it was something else:

```
============================== short test summary info =================================
FAILED tests/system/data_sources/test_sql_server.py::test_schema_validation_core_types_to_bigquery - AssertionError
FAILED tests/system/data_sources/test_sql_server.py::test_schema_validation_not_null_vs_nullable - AssertionError: Column: col_nn, status: fail != success
```

third_party/ibis/ibis_mssql/__init__.py
nj1973: Thanks for the context.

For this particular comment I was not saying it needed action, just trying to understand.

Anyway, I think being consistent with other methods in the same module is probably a good idea. Plus Teradata is slightly different because it is not subclassed from the same backend as some other engines, like SQL Server/Oracle/PostgreSQL.

data_validation/schema_validation.py
```diff
@@ -237,8 +237,10 @@ def split_allow_list_str(allow_list_str: str) -> list:
     # I've not moved this patter to a compiled constant because it should only
     # happen once per command and I felt splitting the pattern into variables
     # aided readability.
-    nullable_pattern = r"!?"
-    precision_scale_pattern = r"(?:\((?:[0-9 ,\-]+|'UTC')\))?"
+    nullable_pattern = r"!?"  # Matches an optional '!'
+    # Matches precision/scale like (N), (N,M), ('UTC'), or (N, 'UTC')
+    # Crucially, [0-9 ,-]+ allows for digits, spaces, commas, and hyphens (for ranges like "1-18").
```
nj1973, line 242: These comments are a nice addition. Thanks.

third_party/ibis/ibis_mssql/__init__.py
```diff
@@ -81,6 +80,7 @@ def connect(dbapi_connection, connection_record):
             with dbapi_connection.cursor() as cur:
                 cur.execute("SET DATEFIRST 1")
 
+        self.client = engine
```
nj1973, line 83: I don't think this line is used. Is it a hangover from the initial Teradata inspired code?

nj1973 (approved):
Other than one tiny nit this looks good to me. I'll approve so you have flexibility to ignore my comment if required.

third_party/ibis/ibis_mssql/__init__.py
helensilva14: You're right, it's not being used right now but between all the changes inspired on the Teradata code this one I thought it was harmless and could be useful for future functions, maybe. I'll keep it for now, but appreciate your comment!

PR #1535, opened 2025-06-16
test: Add tests for -comp-fields on binary columns

### LINKED ISSUE #1524: validate row: Fails when comparing Spanner bytes with BigQuery
**Describe the bug**
Byte columns from Spanner are encoded in base64, because of that, while trying to compare the same data from Spanner with data from BigQuery, row validation with `--comparison-fields` fails to compare rows

latest

Spanner as source
BigQuery as target

1. Create a table in Spanner with a byte column
2. Create a table in BigQuery with a byte column
3. Insert the same byte data in both tables
4. Launch row validation with `--comparison-fields '*'`
5. See that bytes can`t be compared, because bytes from spanner are base64 encoded

**Expected behavior**
Data from Spanner has to be decoded back to raw bytes to be comparable with the same data from other backends

When executing with --hash '*', everything works fine and data is comparable

**Screenshots**
Image

CLI commands on local computer

**Additional context**
As I can see, the problem can`t be fixed via the IBIS framework, because it doesn't support any additional operation with bytes. It can be fixed via custom query operation, but would like to know, is it possible to do automatically

nj1973:
## Description of changes

This PR currently:
- Adds tests for comparison fields on binary columns
- Moves binary to/from hex functions out of cast and makes them independent actions
- Moves engine specific code out of operations.py into engine specific files

There is more to do therefore this is just a draft PR for now.

## Issues to be closed

Closes #1524

## Checklist

- [ ] I have updated any relevant documentation to reflect my changes, if applicable


nj1973:
Abandoned this method for a simpler solution

PR #1539, opened 2025-07-08
fix: Convert base64 Spanner BYTES string to standard bytes string

### LINKED ISSUE #1524: validate row: Fails when comparing Spanner bytes with BigQuery

nj1973:
## Description of changes
This PR:

1. Adds tests for binary values in comp-fields validations (the newly added Spanner test failed).
2. Changed our Spanner `to_pandas()` method to decode base64 Spanner output from BYTES columns to a binary string. The test added in item 1 now passes.
3. Cleans up some unit tests for optionally installed packages.
4. Limits the version of ibm_db_sa package so that tests pass (I created a new issue for this problem: https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1541)

## Issues to be closed

Closes #1524

## Checklist

### DIFF
--- a/tests/system/data_sources/test_oracle.py
+++ b/tests/system/data_sources/test_oracle.py
@@ -636,6 +636,20 @@ def test_row_validation_binary_pk_to_bigquery():
     "data_validation.state_manager.StateManager.get_connection_config",
     new=mock_get_connection_config,
 )
+def test_row_validation_comp_fields_binary_values_to_bigquery():
+    """dvt_binary row validation with comparison fields."""
+    row_validation_test(
+        tables="pso_data_validator.dvt_binary",
+        tc="bq-conn",
+        primary_keys="int_id",
+        comp_fields="*",
+    )
+
+
+@mock.patch(
+    "data_validation.state_manager.StateManager.get_connection_config",
+    new=mock_get_connection_config,
+)
 def test_fixed_char_pk_row_validation_to_bigquery():
     """Test fixed char primary keys"""
     id_column_row_validation_test("pso_data_validator.dvt_fixed_char_id")
--- a/tests/unit/test_clients.py
+++ b/tests/unit/test_clients.py
@@ -23,6 +23,16 @@ from ibis.backends.pandas import BasePandasBackend as PandasBackend
 
 from data_validation import clients, consts, exceptions
 
+try:
+    import ibm_db_sa
+except Exception:
+    ibm_db_sa = None
+
+try:
+    import oracledb
+except Exception:
+    oracledb = None
+
 
 TABLE_NAME = "my_table"
 DATA = [{"a": 1, "b": 2}]
@@ -35,6 +45,14 @@ SOURCE_CONN_CONFIG = {
     "table_name": "my_table",
     "file_path": SOURCE_TABLE_FILE_PATH,
     "file_type": "json",
+}
+
+DB2_CONN_CONFIG = {
+    consts.SOURCE_TYPE: consts.SOURCE_TYPE_DB2,
+    "database": "db",
+    "host": "127.0.0.1",
+    "user": "u",
+    "password": "p",
 }
 
 ORACLE_CONN_CONFIG = {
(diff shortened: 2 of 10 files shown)


third_party/ibis/ibis_cloud_spanner/to_pandas.py
```diff
@@ -34,9 +37,17 @@ def to_pandas(snapshot, sql, query_parameters):
         for row in data_qry:
             data.append(row)
 
-        columns = [f.name for f in data_qry.fields]
+        columns = [_.name for _ in data_qry.fields]
+        bytes_columns = [
+            _.name for _ in data_qry.fields if _.type_.code == TypeCode.BYTES
+        ]
```
helensilva14, line 43:
```suggestion
        columns = [field.name for field in data_qry.fields]
        bytes_columns = [
            f.name for f in data_qry.fields if f.type_.code == TypeCode.BYTES
        ]
```
I'd say, when possible, we could keep our code simple, easy to read and maintain. That's the case here for me, but since it doesn't change the result per se, feel free to skip this suggestion if you feel like!

helensilva14 (approved):
LGTM! Just pointed one suggestion.

third_party/ibis/ibis_cloud_spanner/to_pandas.py
nj1973: Thanks for the review Helen.

I use underscore variables for variables that are temporary in nature and can be ignored. i.e. I have no intention of using `field` or `f` anywhere else.

Personally I find the suggested code more confusing because we use `field` to loop through `fields` the first time and `f` the second time.

I understand it is personal preference but I think I'll keep the underscore version for now - I'm sure the next person in this code will change it to match their preference :-D

Thanks again.

tests/system/data_sources/test_oracle.py
sundar-mudupalli-work, line 1: Wonder if it would make sense to define a function in file `common_functions.py` as follows:
```
def row_validation_comp_fields_binary_values_to_bigquery():
    """dvt_binary row validation with comparison fields."""
    row_validation_test(
        tables="pso_data_validator.dvt_binary",
        tc="bq-conn",
        primary_keys="int_id",
        comp_fields="*",
    ):
    row_validation_test(tables=tables, tc=tc, primary_keys=primary_keys, comp_fields=comp_fields)
```
Then in each of the test files (except teradata), just have:
```
@mock.patch ...
...
def test_row_validation_comp_fields_binary_values_to_bigquery():
       row_validation_comp_fields_binary_values_to_bigquery()
```
Teradata alone would have `row_validation_comp_fields_binary_values_to_bigquery(tables="udf.dvt_binary=pso_data_validator.dvt_binary")`.

tests/unit/test_clients.py
sundar-mudupalli-work, line 1: Hi,

Generally I am opposed to unit tests assuming they can get connections to databases - they are then not unit tests. These tests appear to connect to the databases, not completely sure - since it should not work in practice.

What could be tested is that the required libraries are loaded (and if not print an exception saying "pip install ..."). 

I am not sure if that is what is happening here.

Thanks.

Sundar Mudupalli

Thanks.

Sundar Mudupalli

nj1973: I completely agree that unit tests should not make connections. These do not do that. But the Oracle test did assume that the Oracle package was not installed so whenever I ran it locally it would fail for me, which was annoying. So I wrapped it in a try block so it would test for correct behaviour in both cases.

tests/system/data_sources/test_oracle.py
nj1973: Unfortunately the PR was already approved and merged so we'll have to come back to this optimization in a future PR.

PR #1543, opened 2025-07-09
fix: Move exclude columns processing to cli_tools from config manager

### LINKED ISSUE #1542: row validation: Exception validating PostgreSQL table with over 100 columns with --exclude-columns option
**Describe the bug**
When a PostgreSQL table with over 100 columns is row validated DVT automatically chunks the columns to avoid the following exception:
```
sqlalchemy.exc.OperationalError: (psycopg2.errors.TooManyArguments) cannot pass more than 100 arguments to a function
```

It seems we have a bug where providing `--exclude-columns` tricks the logic into seeing the wrong column list, i.e. the exclusion list.

7.7.0

Oracle, Postgres

Test command using standard DVT test table:
```
data-validation validate row -sc ora -tc pg -tbls pso_data_validator.dvt_many_cols  \
--format=csv --hash="col_004,col_005,col_006" --exclude-columns --primary-keys=id
```

nj1973:
## Description of changes
This PR:
- Adds an PostgreSQL integration test to trigger exception referenced in issue 1542
- Adds BigQuery integration tests covering `--exclude-columns` for row concat/hash/comp fields and column count/sum
- Moves processing for `--exclude-columns` option for concat and hash validations from config manager cli_tools, I needed to do this because of the `--max-concat-columns` logic which needed to be earlier in the process.
- I looked at moving all `--exclude-columns` logic to the same place in cli_tools but it was difficult and was turning my PR into a monster, so I abandoned that thought and stuck to the proble, I was trying to fix.
- Adds unit tests for the cli_tools function that includes the exclude columns processing

## Issues to be closed

Closes #1542

## Checklist

### DIFF
--- a/data_validation/__main__.py
+++ b/data_validation/__main__.py
@@ -108,14 +108,16 @@ def get_aggregate_config(args, config_manager: ConfigManager):
     if args.count:
         col_args = None if args.count == "*" else cli_tools.get_arg_list(args.count)
         aggregate_configs += config_manager.build_config_column_aggregates(
-            "count", col_args, args.exclude_columns, None, cast_to_bigint=cast_to_bigint
+            "count",
+            col_args,
+            None,
+            cast_to_bigint=cast_to_bigint,
         )
     if args.sum:
         col_args = None if args.sum == "*" else cli_tools.get_arg_list(args.sum)
         aggregate_configs += config_manager.build_config_column_aggregates(
             "sum",
             col_args,
-            args.exclude_columns,
             supported_data_types,
             cast_to_bigint=cast_to_bigint,
         )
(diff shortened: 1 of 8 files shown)


data_validation/__main__.py
sundar-mudupalli-work, line 1: Neil,

Great work.

In your comment you say "I looked at moving all --max-concat-columns logic to the same place in cli_tools, it was turning my PR into a monster". Can you elaborate further why this is in the comments - perhaps it should not be ever attempted or maybe there is a blocker - which if removed can make it possible / simpler ?

Thanks.

Sundar Mudupalli

sundar-mudupalli-work (approved):

Issue #1428, opened 2025-02-07
generate-table-partitions could be better with configs run for custom queries


sundar-mudupalli-work:
Hi,

The process of generating partitions and running validations has a difficult  workflow for custom queries.

Current Behavior and Problem 😥
The generate-table-partitions command uses the `-cdir CONFIG_DIR` option to specify a root directory for storing configuration (YAML) files.
| Scenario | Directory created within CONFIG_DIR | Directory Name | Problem for Users |
| --------- | ----------------- | ------------------ | ------------------------------------------------------|
| Regular tables (using `-tbls`) | Yes | `<source_schema>.<source_table>` | Requires users to specify the path for `configs run`. This is manageable as the name is predictable |
| Custom Query (using `-sq` or `-sqf`) | Yes | `custom.<source_query_hash>` | Users cannot easily calculate the `source_query_hash`. This makes it difficult to automate subsequent commands like `configs run` within a script as the required path is unknown |

Since `generate-table-partitions` only supports a single custom query at a time, the existing behavior of creating a sub-directory for YAML files is unnecessary and causes the aforementioned automation problem.

The feature request is to modify `generate-table-partitions` so that when invoked with a custom query, it interprets the provided CONFIG_DIR (from `-cdir CONFIG_DIR`) as the final location for storing the YAML files, without creating an additional, unpredictable sub-directory within it.

This change will allow customers to use a consistent path for both generate-table-partitions `-cdir CONFIG_DIR` and the subsequent `configs run -cdir CONFIG_DIR` command when working with custom queries.

Sundar Mudupalli

PR #1553, opened 2025-07-19
fix: generate-table-partitions generates correct yaml with multiple tables

### LINKED ISSUE #1549: generate-table-partitions generates incorrect config files with Oracle table with 190 columns
**Describe the bug**
When Oracle table with 190 columns (Many having NULL values) is row validated using DVT it's throws below error : 
**cx_Oracle.DatabaseError: ORA-01489: result of string concatenation is too long**

We suspect, it could be because of default replacement string used by DVT.

Ex: 7.5.0

Ex: Oracle 11g for source, Oracle 11g for target

This document has all details : https://docs.google.com/document/d/1X2JluP8PRCjKhqgr0JGuL5Yhg8dFkNlip0qz5F4CYEA/edit?resourcekey=0-V3twA8zxZZxg6nk4yr6HUQ&tab=t.0

sundar-mudupalli-work:
## Description of changes
Fixed generate-table-partitions to produce correct yaml when multiple sets of tables are used, e.g. `-tbls=a=b,c=d`. Customer invoking the command in this way is likely to be rare. DVT creates this situation when a table with numerous columns is validated - that gets split into multiple validations, where the table pairs are the same, but the column list is different.

This PR includes minor functionality change. As stated in [PR 1428](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1428), when generate-table-partitions is invoked with a custom query a random directory name is generated. With this PR, the directory name is no longer random, but a function of the 5 letter hash of the source query. This ensure that when we receive a custom query the directory name is pre-determined. This PR uses that "pre-determined" name to decide where to add the second validation for the same query/table.

## Issues to be closed
Closes #1549

## Checklist

### DIFF
--- a/data_validation/clients.py
+++ b/data_validation/clients.py
@@ -90,7 +90,8 @@ except Exception:
 # DB2 requires ibm_db_sa
 try:
     from third_party.ibis.ibis_db2.api import db2_connect
-except Exception:
+except Exception as e:
+    logging.error(f"Exception {str(e)} while importing db2_connect")
     db2_connect = _raise_missing_client_error("pip install ibm_db_sa")
 
 
--- a/data_validation/config_manager.py
+++ b/data_validation/config_manager.py
@@ -14,8 +14,6 @@
 
 import copy
 import logging
-import string
-import random
 from typing import TYPE_CHECKING, Dict, List, Optional, Union, Tuple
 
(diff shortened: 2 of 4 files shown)


nj1973 (approved):
LGTM

PR #1554, opened 2025-07-20
fix: deduplicate spanner client in SpannerBackend

### LINKED ISSUE #1484: Spanner DVT client does not pass project id through from connection file
Looking at the Spanner client I noticed we create the client twice, once with the project defined and once without.

https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/8feafe2657bd9d2fc245c190a452ca54b959b87c/third_party/ibis/ibis_cloud_spanner/__init__.py#L46-L60

Notice above we set `self.spanner_client` and use it to process the instance and database args but then we create the client a second time as `self.client` without the project setting and this is the one we use throughout a validation.

I think we need to combine these so we only take a single connection, in `self.client`.

I've not noticed it before because the containing GCE VM had the correct project as it's default. This breaks when validating with Spanner in a different project.

jx2lee:
## Description of changes

## Issues to be closed

Closes #1484

## Checklist

### DIFF
--- a/third_party/ibis/ibis_cloud_spanner/__init__.py
+++ b/third_party/ibis/ibis_cloud_spanner/__init__.py
@@ -49,7 +49,7 @@ class Backend(BaseSQLBackend):
         self.client = spanner.Client(
             project=project_id, credentials=credentials, client_options=options
         )
-        self.instance = self.spanner_client.instance(instance_id)
+        self.instance = self.client.instance(instance_id)
         self.database_name = self.instance.database(database_id)
         (
             self.data_instance,
--- a/third_party/ibis/ibis_cloud_spanner/__init__.py
+++ b/third_party/ibis/ibis_cloud_spanner/__init__.py
@@ -15,7 +15,6 @@
 from typing import Any, Mapping, Optional, Tuple
 
 from google.api_core import client_options
-import google.cloud.spanner as cs
 import ibis.expr.schema as sch
 import ibis.expr.types as ir
 from google.cloud import spanner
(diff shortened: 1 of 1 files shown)


third_party/ibis/ibis_cloud_spanner/__init__.py
```diff
@@ -43,8 +42,12 @@ def do_connect(
...
-        self.spanner_client = spanner.Client(
-            project=project_id, credentials=credentials
+        options = None
+        if api_endpoint:
+            options = client_options.ClientOptions(api_endpoint=api_endpoint)
+
+        self.client = spanner.Client(
+            project=project_id, credentials=credentials, client_options=options
         )
         self.instance = self.spanner_client.instance(instance_id)
```
nj1973: This line should be:
```
self.instance = self.client.instance(instance_id)
```

nj1973 (changes requested):
Thanks for picking this up. Just one small edit required.

third_party/ibis/ibis_cloud_spanner/__init__.py
jx2lee: Ooops, I missed it. Done, Thanks! 🙇🏽

nj1973:
Please excuse the many "/gcbrun" comments. We have some flaky DB2 tests I'm trying work around.

nj1973 (approved):

nj1973:
@jx2lee , thanks for the contribution. Much appreciated.

Issue #1555, opened 2025-07-21
Column validation: Implement support for INTERVAL columns


nj1973:
Column validations of Interval columns are skipped:

```
07/21/2025 03:26:13 PM-INFO: Skipping sum on col_interval_ds due to data type: interval
07/21/2025 03:26:13 PM-INFO: Skipping sum on col_interval_ym due to data type: interval
07/21/2025 03:26:13 PM-INFO: Skipping min on col_interval_ds due to data type: interval
07/21/2025 03:26:13 PM-INFO: Skipping min on col_interval_ym due to data type: interval
```

For this issue we need to look into which aggregations we can support and what they should look like.

Tests `test_column_validation_intervals` should test supported agg types as they are implemented.

See https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1214 for the Row validation counterpart to this issue.

PR #1556, opened 2025-07-21
feat: Prevent exceptions when using Oracle INTERVAL YM and PostgreSQL INTERVAL

### LINKED ISSUE #1547: Oracle Interval YM Validations
Now that we have migrated from cx-Oracle to oracledb we no longer get the following exception when validating Interval Yeat to Month columns:

```
Oracle type 2016 not supported.
```

However we still get an exception. Test case below.

Oracle test table:
```
CREATE TABLE dvt_test.tab_intervals
(id NUMBER(5) NOT NULL PRIMARY KEY
, col_interval_ds INTERVAL DAY TO SECOND
, col_interval_ym INTERVAL YEAR TO MONTH);
INSERT INTO dvt_test.tab_intervals VALUES
(1,INTERVAL '1 2:03:44.0' DAY TO SECOND(0)
  ,INTERVAL '1-2' YEAR TO MONTH);
COMMIT;
```

PostgreSQL test table:
```
CREATE TABLE dvt_test.tab_intervals
(id integer NOT NULL PRIMARY KEY
, col_interval_ds INTERVAL
, col_interval_ym INTERVAL);
INSERT INTO dvt_test.tab_intervals VALUES
(1,INTERVAL '1 day 2 hours 3 minutes 44 seconds'
  ,INTERVAL '1 year 2 months');
```

Test command:
```
data-validation validate schema -sc=ora -tc=pg -tbls=dvt_test.tab_intervals
...
sqlalchemy.exc.DatabaseError: (oracledb.exceptions.DatabaseError) ORA-00942: table or view does not exist
Help: https://docs.oracle.com/error-help/db/ora-00942/
[SQL: SELECT * FROM tab_intervals t0 WHERE ROWNUM <= 1]
```

There are two layers to the problem, both in `site-packages/ibis/backends/base/sql/alchemy/__init__.py`

1. In `_get_sqla_table`, `sa.Table` returns the interval column with a type of `NullType`.
2. In `_handle_failed_column_type_inference` the table name is passed into `_metadats` without a schema prefix which then fails.

We should investigate item 1 first to see if we can add INTERVAL_YM support. If not then we should look into item 2 instead.

nj1973:
## Description of changes
This PR adds basic support for INTERVAL types in:

- Oracle
- PostgreSQL
- BigQuery
- Teradata

Spanner and Snowflake don't appear to allow intervals as table columns, only in expressions. As far as I can see, other supported engines don't have an interval data type.

By basic support I mean no exception is raised and the column is recognized as being of type `dt.Interval`.

Column and Row validations are not implemented by this PR.

- Column: Currently column validations skip for of type interval. See https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1555
- Row: Currently row validations fail due to inconsistent string formats. See https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1214

**Please note I'll re-run tests once this has been approved. The new test table breaks Oracle tests until the fixes on this branch are implemented.**

## Issues to be closed

Closes #1547

## Checklist


sundar-mudupalli-work (changes requested):
Hi,

I think the changes here are inconsistent with the list of issues. 
1. [Issue 1555](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1555) says that an error message will be printed if someone tries to do a column validation of an interval-y-m column with the sum aggregate. However the test suites here try to do exactly that. So issue 1555 is either closed by this PR or needs to be updated. 
2. Because the representation for interval is different in different databases, we should understand what might happen if they are "summed" up and if the results are "true" to what customers might expect or if DVT should disallow some comparisons. This can remain an open issue if this PR raises an error on column validation for sum, min, max aggregates - not clear it does.

Sundar Mudupalli

nj1973:
> I think the changes here are inconsistent with the list of issues.
> 
> 1. [Issue 1555](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1555) says that an error message will be printed if someone tries to do a column validation of an interval-y-m column with the sum aggregate. However the test suites here try to do exactly that. So issue 1555 is either closed by this PR or needs to be updated.
> 2. Because the representation for interval is different in different databases, we should understand what might happen if they are "summed" up and if the results are "true" to what customers might expect or if DVT should disallow some comparisons. This can remain an open issue if this PR raises an error on column validation for sum, min, max aggregates - not clear it does.

1. I created 1555 after adding the tests to this branch. Issue 1555 is not fixed by these changes but is a product of it. The tests on this branch do not fail because the existing code path skips the columns. During issue 1555 work the tests will start to actually test something. 
2. I agree 'we should understand what might happen if they are "summed" up'. My intention was that that is part of issue 1555, part of implementing column validation support.
 
The plan was this:

a) Issue 1547 (this issue): Ensure that INTERVAL columns do not cause exceptions and the data type is identified as an Interval.
b) Issue 1214: Work on a consistent format for interval columns in hash and concat validations.
c) Issue 1555: Look at which aggregations we can support and how we can implement them.

sundar-mudupalli-work (approved):

PR #1561, opened 2025-08-01
feat: Reinstate support for SQLAlchemy url for Oracle connections

### LINKED ISSUE #1564: Reinstate support for SQLAlchemy url for Oracle connections
As part of our switch from cx_Oracle to oracledb we disabled use of a SQLAlchemy URL for Oracle connections.

I recently found out that one of our larger customers makes extensive use of connections by url. They store the url, with username/password/service name as a secret in Secret Manager and pull the url at run time.

nj1973:
## Description of changes
As part of our switch from cx_Oracle to oracledb we disabled use of a SQLAlchemy URL for Oracle connections.

I recently found out that one of our larger customers makes extensive use of connections by url. They store the url, with username/password/service name as a secret in Secret Manager and pull the url at run time.

This PR reinstates connection to oracle via url and adds a test.

## Issues to be closed
Closes #1564

## Checklist

### DIFF
--- a/third_party/ibis/ibis_oracle/__init__.py
+++ b/third_party/ibis/ibis_oracle/__init__.py
@@ -131,7 +131,7 @@ class Backend(BaseAlchemyBackend):
         else:
             connect_args = {} if not connect_args else connect_args
             if thick_mode:
-                # Configuration explicitly requests thick_mode or user not specified, credentials in wallet - requires thick_mode
+                # Configuration explicitly requests thick_mode.
                 oracledb.init_oracle_client(
                     config_dir=connect_args.get("config_dir", None)
                 )
--- a/third_party/ibis/ibis_oracle/__init__.py
+++ b/third_party/ibis/ibis_oracle/__init__.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Iterable, Literal, Tuple, Dict, Any
+from typing import Iterable, Literal, Optional, Tuple, Dict, Any
 
 import sqlalchemy as sa
 from sqlalchemy.dialects.oracle.base import (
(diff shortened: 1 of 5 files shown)


sundar-mudupalli-work:
Neil,

Let us document the behavior clearly. In the past, if a URL was provided, DVT ignored all the other parameters.
Here is one behavior we could have:
- If a URL is provided, all other parameters except "thick-mode" are ignored. If we ignored the "thick-mode" parameter, customers cannot use "thick-mode" and URL - because "thin-mode" is default. The format of the URL is defined by SQLAlchemy [here](https://docs.sqlalchemy.org/en/20/dialects/oracle.html#module-sqlalchemy.dialects.oracle.oracledb).
- If a URL is not provided, then we pass the parameters provided to `python-oracledb` library.
My research shows URL's are an active area of change in the ibis library. Looking at the [release notes](https://ibis-project.org/release_notes#9.0.0), there are 36 references to URL.

Sundar Mudupalli

third_party/ibis/ibis_oracle/__init__.py
```diff
@@ -97,35 +97,48 @@ def __init__(self, arraysize: int = 500):
...
+                        "port": port,
+                        "service_name": database,
+                        "protocol": protocol,
+                    }
+                )
+            sa_url = "oracle+oracledb://@"
+        else:
+            connect_args = {} if not connect_args else connect_args
+            if thick_mode:
+                # Configuration explicitly requests thick_mode or user not specified, credentials in wallet - requires thick_mode
```
sundar-mudupalli-work: Minor nit - here it only thick mode - adjust comment appropriately.

sundar-mudupalli-work (approved):
Please update Connections.md to clearly show URL format (from SQLAlchemy) and that using URL means all other parameters are ignored.

To move this along faster, approving this - assuming documentation will be updated.

Great work.

Thank you.

Sundar Mudupalli

nj1973:
@sundar-mudupalli-work , I've updated `Connections.md` to better document the URL option. But I don't believe that using URL means all other parameters are ignored, I tested and they can be mixed and matched. I've documented both examples. Hopefully it makes sense.

sundar-mudupalli-work:
Neil,

Apologies. You are correct. We pass `connect_args` to SQLAlchemy (I made that code change!!), so it is up to SQLAlchemy to decide which one takes precedence. For some reason, I kept looking at the old code (before my code changes). I did not see that in the old code and it did not show up as a change in the new code (which I was expecting), but was already there!!

Excellent work.

Sundar Mudupalli

PR #1572, opened 2025-08-18
feat: BigQuery result handler optimization

### LINKED ISSUE #1568: Investigate BigQuery result handler optimization
The BigQuery result handler currently writes using `insert_rows_from_dataframe()`:
```
       table = self._bigquery_client.get_table(self._table_id)
        chunk_errors = self._bigquery_client.insert_rows_from_dataframe(
            table, result_df
        )
```

An alternative is `load_table_from_dataframe()`: https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.client.Client#google_cloud_bigquery_client_Client_load_table_from_dataframe

A brief test writing 100 million rows showed the following results

`insert_rows_from_dataframe`:
```
04/03/2025 11:16:18 AM-DEBUG: Write results to BigQuery elapsed: 178.15s
```

`load_table_from_dataframe`:
```
08/08/2025 09:09:03 AM-DEBUG: Write results to BigQuery elapsed: 10.88s
```

BigQuery result handler

The code change I tested with was:
```
-        chunk_errors = self._bigquery_client.insert_rows_from_dataframe(
-            table, result_df
-        )
-        if any(chunk_errors):
+        job = self._bigquery_client.load_table_from_dataframe(result_df, table)
+        job.result()
+        if job.errors:
```

We would need to:
- test with > 1 million rows, we need to understand any downsides to switching load methods (e.g. more local memory usage perhaps?)
- look at what should be done inside the error handling code to replicate to advisory messages
- double check that written data matches for both methods
- attempt to compare resource usage in BigQuery, e.g. slot usage

nj1973:
## Description of changes
Switch BigQuery result handler code from `insert_rows_from_dataframe` to `load_table_from_dataframe`.

## Issues to be closed

Closes #1568

## Checklist

### DIFF
--- a/data_validation/result_handlers/bigquery.py
+++ b/data_validation/result_handlers/bigquery.py
@@ -16,6 +16,8 @@
 
 import logging
 
+from google.api_core.exceptions import BadRequest
+
 from data_validation import clients, consts, exceptions, util
 from data_validation.result_handlers.base_backend import BaseBackendResultHandler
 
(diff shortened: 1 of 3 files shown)


sundar-mudupalli-work:
Hi,

I have an issue with this change - BQ has a limit of 1500 modifications per table per day and that includes Load jobs, and other overwrites - details [here](https://cloud.google.com/bigquery/quotas#standard_tables). This does not apply to streaming uploads. 

Customers validate multiple table pairs use the same report handler BQ table. If the tables are large - they could be running more than 10,000 partitions validating within hours. Each partition validation creates a separate load job. With this fix, that validation and all other validations will fail for the rest of the day.

We could consider this as an option for customers who want to run validations when they know it will be fewer than 1,500 per day.

Thanks.

Sundar Mudupalli

nj1973:
> I have an issue with this change - BQ has a limit of 1500 modifications per table per day and that includes Load jobs, and other overwrites - details [here](https://cloud.google.com/bigquery/quotas#standard_tables). This does not apply to streaming uploads.

This is good information, thanks. I think this renders these changes irrelevant, seeing as the driver for them was a customer running many validations in a single day.

Perhaps we instead just need to document why use the Streaming API to avoid future wasted effort. We can discuss on next team call.

nj1973:
Abandoning this PR due to information highlighted in Sundar's comment. Switching to load jobs would save a lot of slot pain for the customer and speed up validations, but it would give them a new headache of a 1500 job daily limit updating the results table. The new headache would be a blocker rather than an annoyance so is worse :-(
I won't delete the branch in case we want to refer back to these changes in the future.

Issue #1578, opened 2025-08-28
SQL Server image data type is not supported


nj1973:
We have a customer validating SQL Server and PostgreSQL data. They have some columns of "image" data type. When working on https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1577 I added a table of many SQL Server data types, the image column caused errors.

To reproduce errors we can add "col_image" back into the test constant `DVT_SQL_SERVER_TYPES_COLUMNS`.

```
FAILED tests/system/data_sources/test_sql_server.py::test_column_validation_sql_server_types - sqlalchemy.exc.ProgrammingError: (pyodbc.ProgrammingError) ('42000', '[42000] [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Operand data type image is invalid...
FAILED tests/system/data_sources/test_sql_server.py::test_row_validation_sql_server_types - sqlalchemy.exc.DataError: (pyodbc.DataError) ('22018', '[22018] [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Explicit conversion from data type image to varc...
```

nj1973:
This issue needs the tests added in https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1577.

nj1973:
This is worth noting (https://learn.microsoft.com/en-us/sql/t-sql/data-types/ntext-text-and-image-transact-sql?view=sql-server-ver17):

```
The ntext, text, and image data types will be removed in a future version of SQL Server. Avoid using these data
types in new development work, and plan to modify applications that currently use them. Use nvarchar(max),
varchar(max), and varbinary(max) instead.
```

For this reason it's probably not worth investing a lot of effort into supporting image. Any easy wins can be implemented but it would also be okay to skip them.

Perhaps we have a temporary change to skip them for now and keep this issue open for a future effort.

PR #1579, opened 2025-08-28
fix: SQL Server string length fixes

### LINKED ISSUE #1577: Column validation of nvarchar on SQL Server is incorrect
**Describe the bug**
Reporting customer issue:

"
While working on DVT for SQL server, I have observed that for columns where datatype is nvarchar, the the datalength() function is showing double the length of the string.

On exploring little bit, I came to know that nvarchar will take 2 bytes per character and varchar will take 1 byte per character. So if you use datalength() on an nvarchar column it returns twice the no. of characters.

However, the len() function provides the correct length.
"

7.7.0

SQL Server/PostgreSQL

**Expected behavior**
Expected length based string validations to be character based, not byte based.

nj1973:
## Description of changes
This PR fixes three issues:

1. DVT used `datalength()` to get the length of a string, but `datalength()` returns the "byte length" which is incompatible with other SQL engines. This PR switches support to `len()` which returns the length in characters.  However...
2. `len()` ignores trailing spaces! This is also incompatible with other SQL engines. This PR also protects spaces in strings before calculating the string length in characters.
3. Columns of type "text" or "ntext" are incompatible with the `len()` function. For these we fall back to using a byte length which means "text" data in a multi byte character set will fail validation when compared to other systems. I don't believe we have a way forward for this issue. SQL Server documentation advises users not to use "text"/"ntext" so I think this is an acceptable limitation.

In this PR we also introduce a new SQL Server test table that contains all SQL Server data types. Only the "image" data type threw up problems and I created a new issue for that: https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1578

## Issues to be closed

Closes #1577

## Checklist


nj1973:
Apologies Helen, I decided to make a last minute test alteration. Hopefully the tests pass this time.

helensilva14:
> Apologies Helen, I decided to make a last minute test alteration. Hopefully the tests pass this time.

All good, no problem! I'll wait for the build to finish, but let me know if you need more time.

data_validation/config_manager.py
```diff
@@ -809,7 +828,12 @@ def append_pre_agg_calc_field(
             depth = 1
             calc_func = consts.CALC_FIELD_LENGTH
         elif column_type in ["string", "!string"]:
-            calc_func = consts.CALC_FIELD_LENGTH
+            if self._is_sql_server_text(source_column) or self._is_sql_server_text(
```
helensilva14: Wondering if at this line we could check if source and/or target engines are equal to SQL Server before checking the TEXT data type... Because right now it'll be called every time for every string column per my understanding. But not sure if checking the engine that frequently can affect performance or so, what do you think? 🤔

third_party/ibis/ibis_mssql/registry.py
```diff
@@ -84,6 +84,16 @@ def sa_epoch_seconds(translator, op):
     )
 
 
+def sa_format_string_length(translator, op):
+    """Calculate SQL Server string length in characters, not bytes.
+
+    This uses len() rather than datalength() to give an answer compatible with BigQuery. But we
+    need to protect trailing spaces due to an odd quirk in SQL Server len() behaviour. This is
+    done by first replacing spaces with underscores."""
```
helensilva14:
```suggestion
    This uses len() rather than datalength() to give an output compatible with BigQuery.
    We need to protect trailing spaces due to an odd quirk in SQL Server len() 
    behaviour that ignores them, so this is done by first replacing spaces with 
    underscores before passing the string to len()."""
```

data_validation/config_manager.py
nj1973: You are right but not for performance reasons. I think looking up the type in a dict will not be much slower than first checking a scalar variable. The bigger reason you are right is that I need to check for SQL Server because another engine may also have a text or ntext type that does not need this workaround.

Thanks, I'll make the change.

helensilva14 (approved):
LGTM, great job adding the engine check in the new method!

nj1973:
> LGTM, great job adding the engine check in the new method!

Thanks for the review @helensilva14 , without your keen eye I would accidentally included a future bug in my code. 💯

Issue #1584, opened 2025-09-02
SQL Server sum of tinyint, smallint, int columns can overflow with exception


nj1973:
**Describe the bug**

When a sum validation is executed on SQL Server tinyint, smallint, int columns the sum expression can overflow causing an exception. This is SQL Server behaviour but something we should cater for.

Test table:
```
CREATE TABLE dvt_test.dvt_ints (
    id            integer NOT NULL PRIMARY KEY
,   col_tinyint   tinyint
,   col_smallint  smallint
,   col_int       integer);

insert into dvt_test.dvt_ints values (1,210,31000,1000000000);
insert into dvt_test.dvt_ints values (2,220,32000,2000000000);
```

Test command:
```
$ data-validation -v validate column -sc=sqlserver -tc=sqlserver -tbls=dvt_test.dvt_ints --sum="*"
...
sqlalchemy.exc.DataError: (pyodbc.DataError) ('22003', '[22003] [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Arithmetic overflow error converting expression to data type int. (8115) (SQLExecDirectW)')
[SQL: SELECT count(*) AS count, sum(t0.id) AS sum__id, sum(t0.col_tinyint) AS sum__col_tinyint, sum(t0.col_smallint) AS sum__col_smallint, sum(t0.col_int) AS sum__col_int
FROM dvt_test.dvt_ints AS t0]
```

**What version of DVT are you using?**
8.1.1

**What type of connections are you using for source and target?**
SQL Server

**Expected behavior**
I suspect we should first cast columns of these "smaller" data types to bigint and then sum them.

nj1973:
It turns out this was already fixed in https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/461.

Retested with the `--cast-to-bigint` option and the command completed successfully.

PR #1590, opened 2025-09-17
fix: Cast SQL Server len() expression to BIGINT

### LINKED ISSUE #1589: SQL Server column validation exception for sum of string lengths when int overflows
**Describe the bug**
When we use `--sum` on a string column we sum up the string lengths. It is possible for this aggregation to overflow the integer data type. We should cast these length expressions to bigint to prevent this.

8.1.0

SQL Server

nj1973:
## Description of changes
Casts SQL Server len/datalength expressions to bigint to prevent column validations on large tables from overflowing the int data type.

## Issues to be closed

Closes #1589

## Checklist

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -303,7 +303,9 @@ def integration_db2(session):
     """Run DB2 integration tests.
     Ensure DB2 validation is running as expected.
     """
-    _setup_session_requirements(session, extra_packages=["ibm-db-sa<0.4.2", "ibm-db<3.2.7"])
+    _setup_session_requirements(
+        session, extra_packages=["ibm-db-sa<0.4.2", "ibm-db<3.2.7"]
+    )
 
     expected_env_vars = [
         "PROJECT_ID",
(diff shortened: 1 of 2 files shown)


helensilva14 (approved):
LGTM!

sundar-mudupalli-work:
Neil,

This resolves the issue of string lengths adding up over the size of the integer. However, even int32 columns have the same issue. It may be best to focus on the issues raised by customers - so we know we are fixing things that need to be fixed.

Thank you.

Sundar Mudupalli

nj1973:
> Neil,
> 
> This resolves the issue of string lengths adding up over the size of the integer. However, even int32 columns have the same issue. It may be best to focus on the issues raised by customers - so we know we are fixing things that need to be fixed.
> 
> Thank you.
> 
> Sundar Mudupalli

This was raised by a customer, but directly to me and I created the issue on their behalf.

The same customer did raise the "int32 columns" issue with me (https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1584) but I realised we already had an option to cater for it so closed it. I think we should automate the option in the future because it is unnecessary but I haven't created an issue for that (yet).

PR #1601, opened 2025-09-25
fix: Lower case schema.table find-tables keys (if possible) when matching names

### LINKED ISSUE #1600: find-tables fails to match when there's a case mismatch in two systems
**Describe the bug**
`find-tables` fails to match tables when their names differ through case.

A validation works fine, it matches on lower cased table names. But find-tables does not.

8.2.0

SQL Server/PostgreSQL but applies to all engines

SQL Server test table:
```
create table [dvt_test].[TAB_UPPER] (id int);
```

BigQuery test table:
```
CREATE OR REPLACE TABLE dvt_test.tab_upper(id int64);
```

Validation works because we coerce names to lower case:
```
$ data-validation -v validate schema -sc=sqlserver -tc=bq -tbls=dvt_test.tab_upper
╒══════════════════════════════════════╤═══════════════════╤═══════════════════╤══════════╤═════════════════════╤═════════════════════╤══════════════════════╤══════════════════════╤════════════════════╤════════════════════╤═════════════════════╕
│ run_id                               │ validation_name   │ validation_type   │ labels   │ source_table_name   │ target_table_name   │ source_column_name   │ target_column_name   │ source_agg_value   │ target_agg_value   │ validation_status   │
╞══════════════════════════════════════╪═══════════════════╪═══════════════════╪══════════╪═════════════════════╪═════════════════════╪══════════════════════╪══════════════════════╪════════════════════╪════════════════════╪═════════════════════╡
│ 9a2288f0-4160-443e-92be-9a6665163d4e │ Schema            │ Schema            │ []       │ dvt_test.tab_upper  │ dvt_test.tab_upper  │ id                   │ id                   │ int32              │ int64              │ fail                │
╘══════════════════════════════════════╧═══════════════════╧═══════════════════╧══════════╧═════════════════════╧═════════════════════╧══════════════════════╧══════════════════════╧════════════════════╧════════════════════╧═════════════════════╛
```

But find-tables fails, it does not coerce to lower case.
```
data-validation find-tables --source-conn sqlserver --target-conn bq --allowed-schemas dvt_test
[]
```

**Expected behavior**
Expect the names to be matched

nj1973:
## Description of changes
This PR lower cases the key used to match names in `find-tables` when no lower cased key already exists. This is useful when names in source/target systems have different cases. Below is an example of the output when this is the case:
```
[{"schema_name": "dvt_test", "table_name": "TAB_UPPER", "target_schema_name": "dvt_test", "target_table_name": "tab_upper"}]
```

## Issues to be closed

Closes #1600

## Checklist

### DIFF
--- a/data_validation/find_tables.py
+++ b/data_validation/find_tables.py
@@ -71,10 +71,10 @@ def _get_table_map_from_obj_list(table_objs: list) -> dict:
         }
 
     # Post process table_map and lower case table_keys, if possible.
-    for table_key in [_ for _ in table_map if _ != _.lower()]:
+    for table_key in [_ for _ in table_map if _ != _.casefold()]:
         if table_key.lower() not in table_map:
             # Only lower case the key if there isn't one already.
-            table_map[table_key.lower()] = table_map.pop(table_key)
+            table_map[table_key.casefold()] = table_map.pop(table_key)
 
     return table_map
 
--- a/data_validation/find_tables.py
+++ b/data_validation/find_tables.py
@@ -72,7 +72,7 @@ def _get_table_map_from_obj_list(table_objs: list) -> dict:
 
     # Post process table_map and lower case table_keys, if possible.
     for table_key in [_ for _ in table_map if _ != _.casefold()]:
-        if table_key.lower() not in table_map:
+        if table_key.casefold() not in table_map:
             # Only lower case the key if there isn't one already.
             table_map[table_key.casefold()] = table_map.pop(table_key)
 
--- a/data_validation/find_tables.py
+++ b/data_validation/find_tables.py
@@ -60,15 +60,9 @@ def _compare_match_tables(source_table_map, target_table_map, score_cutoff=0.8) 
     return table_configs
 
 
-def _get_table_map(
-    client: "ibis.backends.base.BaseBackend", allowed_schemas=None, include_views=False
-):
-    """Return dict with searchable keys for table matching."""
+def _get_table_map_from_obj_list(table_objs: list) -> dict:
+    """Convert list of schema, table tuples into dict with searchable keys for table matching."""
     table_map = {}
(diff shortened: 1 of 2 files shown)


sundar-mudupalli-work (changes requested):
Neil,

This is great stuff. Would it be better to use casefold() instead of lower. Table names can be unicode characters, though that is not very common. I am Ok if you have a reason to stick with lower. Seeing that some of our customers are German, I thought maybe casefold would be better.

Sundar Mudupalli

nj1973:
> This is great stuff. Would it be better to use casefold() instead of lower. Table names can be unicode characters, though that is not very common. I am Ok if you have a reason to stick with lower. Seeing that some of our customers are German, I thought maybe casefold would be better.

Good idea, change made. Thanks

data_validation/find_tables.py
```diff
@@ -60,25 +60,35 @@ def _compare_match_tables(source_table_map, target_table_map, score_cutoff=0.8)
...
     for table_obj in table_objs:
         table_key = ".".join([t for t in table_obj if t])
         table_map[table_key] = {
             consts.CONFIG_SCHEMA_NAME: table_obj[0],
             consts.CONFIG_TABLE_NAME: table_obj[1],
         }
 
+    # Post process table_map and lower case table_keys, if possible.
+    for table_key in [_ for _ in table_map if _ != _.casefold()]:
+        if table_key.lower() not in table_map:
```
sundar-mudupalli-work: should this say `table_key.casefold()`?

To detect these issues, perhaps add "Straße" to unit test. According to AI - "Straße".lower() = '"straße", while "Straße".casefold() is 'strasse'

Enjoy,

Sundar Mudupalli

sundar-mudupalli-work (approved):
Thank you very much

PR #1609, opened 2025-10-13
fix: Prevent result handler credentials from being stored in config files

### LINKED ISSUE #1608: Specifying CONNECTION_NAME.DATASET.TABLE for result handler results in parameters including passwords in config files
I am working on setting up DVT for ~600 tables. In QA we are executing each of these and generating yaml config files so we can run those in higher environment to keep our validation logic consistent. I have results handler on postgres which is also my target connection. so i define my -rh as target_conn.pso_data_validator.results. 
When generating the config YAML files using --config-file option, i noticed that yaml file has hardocded all connection param values including password of the target_conn for the results handler. I have run this w/o providing a results handler string, it does not do it for source and target connection.  At this  point I am unable to load all my config files to github because of the hardcoded password.

8.1.1

oracle for source, postgres for target

Output YAML file generated using "-c" option should have results handler point to connection name  and not its individual values similar to it working for source and target connection. 
Note: This connection is created using the connections -add  option. 

Setting up via BigQuery but my results data is not huge since i am only collecting failed results. It is quicker to set up with psotgres. 

on prem

**Additional context**

nj1973:
## Description of changes
Use connection names in result handler config instead of the actual connection content. This prevents credentials from being written to config files. It's also a cleaner way to do it.

## Issues to be closed

Closes #1608

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -594,9 +594,8 @@ The `--config-dir` flag will specify the directory with the YAML files to be exe
 
 ### Validation Reports
 
-The result handlers tell DVT where to store the results of each validation. The
-tool can write the results of a validation run to Google BigQuery, PostgreSQL
-or print to stdout (default). View the schema of the results table [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/terraform/results_schema.json).
+The result handlers tell DVT where to store the results of each validation. The tool can write the results of a validation run to Google BigQuery, PostgreSQL or print to stdout (default). See [result handler setup](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/docs/installation.md#result-handler-setup) for pre-requisites. View the schema of the BigQuery results table [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/terraform/results_schema.json).
+
 
 To output to BigQuery or PostgreSQL, simply include the `-rh` flag during a validation run including
 the schema and table name for the results.
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -1308,21 +1308,16 @@ def _get_result_handler(rc_value: str, sa_file=None) -> dict:
     if config[0] in connections:
         # We received connection_name.results_table.
         conn_from_file = get_connection(config[0])
-        if conn_from_file[consts.SOURCE_TYPE] == consts.SOURCE_TYPE_BIGQUERY:
-            result_handler = {
-                consts.RH_TYPE: conn_from_file[consts.SOURCE_TYPE],
-                consts.PROJECT_ID: conn_from_file["project_id"],
-                consts.TABLE_ID: config[1],
-                consts.API_ENDPOINT: conn_from_file.get("api_endpoint", None),
-                consts.STORAGE_API_ENDPOINT: conn_from_file.get(
-                    "storage_api_endpoint", None
-                ),
-            }
-        elif conn_from_file[consts.SOURCE_TYPE] == consts.SOURCE_TYPE_POSTGRES:
+        if conn_from_file[consts.SOURCE_TYPE] in [
+            consts.SOURCE_TYPE_BIGQUERY,
+            consts.SOURCE_TYPE_POSTGRES,
+        ]:
             result_handler = {
                 consts.RH_TYPE: conn_from_file[consts.SOURCE_TYPE],
                 consts.TABLE_ID: config[1],
-                consts.RH_CONN: conn_from_file,
+                # Only store the connection name in the result handler to avoid accidentally
+                # storing credentials in config files.
+                consts.RH_CONN: config[0],
             }
         # TODO Add filesytem handler too.
         else:
--- a/data_validation/result_handlers/base_backend.py
+++ b/data_validation/result_handlers/base_backend.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import logging
-from typing import TYPE_CHECKING
+from typing import Optional, TYPE_CHECKING
 
 import ibis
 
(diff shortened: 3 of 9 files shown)


sundar-mudupalli-work:
Hi,

I am not clear how to define and use the Postgres results table. Can you provide the DDL ?

Here is what I tried and the error I got
```
data-validation validate row -tc postgres -sc bq -tbls=data_validator_issues.issue1610=public.dvt_core_types -pk id -hash='col_datetime' -rh postgres.public.results
...
File "/home/user/professional-services-data-validator/env/lib/python3.12/site-packages/data_validation/result_handlers/postgres.py", line 162, in _insert_postgres
    _ = clients.get_ibis_table(self._client, schema_name, table_name)
...
__
    raise NotImplementedError(
NotImplementedError: Could not find signature for to_sqla_type: <PGDialect_psycopg2, NoneType>
```
Here is the results table definition and maybe that is not correct
```
CREATE TYPE PUBLIC.err_record AS ( code bigint, message text, details text );

CREATE TYPE PUBLIC.label_rec AS ( KEY text, VALUE text );

CREATE TABLE PUBLIC.results
  (
     run_id             TEXT,
     validation_name    TEXT,
     validation_type    TEXT,
     start_time         TIMESTAMP,
     end_time           TIMESTAMP,
     source_table_name  TEXT,
     target_table_name  TEXT,
     source_column_name TEXT,
     target_column_name TEXT,
     aggregation_type   TEXT,
     group_by_columns   TEXT,
     primary_keys       TEXT,
     num_random_rows    BIGINT,
     source_agg_value   TEXT,
     target_agg_value   TEXT,
     difference         DOUBLE PRECISION,
     pct_difference     DOUBLE PRECISION,
     pct_threshold      DOUBLE PRECISION,
     validation_status  TEXT,
     labels             LABEL_REC[],
     configuration_json TEXT,
     error_result       ERR_RECORD
  );
```

nj1973:
You don't need to pre-create the table. The docs ask you just create a schema.

https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/docs/installation.md#postgresql-1

I tested it and it works for me.

```
postgres=# \d pso_data_validator_results.results
Did not find any relation named "pso_data_validator_results.results".
```

```
data-validation validate row -sc ora_local -tc pg_local -tbls=pso_data_validator.dvt_core_types  --hash="*" -rh=pg_rh.pso_data_validator_results.results --labels=tag1=neil-dvt1
```

The table exists and holds the results:
```
postgres=# \d pso_data_validator_results.results
                   Table "pso_data_validator_results.results"
       Column       |           Type            | Collation | Nullable | Default
--------------------+---------------------------+-----------+----------+---------
 validation_name    | character varying(3000)   |           | not null |
 validation_type    | character varying(3000)   |           | not null |
 aggregation_type   | character varying(3000)   |           |          |
 source_table_name  | character varying(3000)   |           |          |
 source_column_name | character varying(3000)   |           |          |
 source_agg_value   | character varying(3000)   |           |          |
 target_table_name  | character varying(3000)   |           |          |
 target_column_name | character varying(3000)   |           |          |
 target_agg_value   | character varying(3000)   |           |          |
 group_by_columns   | character varying(3000)   |           |          |
 primary_keys       | character varying(3000)   |           |          |
 num_random_rows    | bigint                    |           |          |
 difference         | double precision          |           |          |
 pct_difference     | double precision          |           |          |
 pct_threshold      | double precision          |           |          |
 validation_status  | character varying(3000)   |           | not null |
 run_id             | character varying(3000)   |           | not null |
 labels             | character varying(3000)[] |           |          |
 start_time         | timestamp with time zone  |           | not null |
 end_time           | timestamp with time zone  |           | not null |

postgres=# select validation_name,labels from pso_data_validator_results.results;
 validation_name |       labels
-----------------+--------------------
 hash__all       | {'tag1=neil-dvt1'}
 hash__all       | {'tag1=neil-dvt1'}
 hash__all       | {'tag1=neil-dvt1'}
```

sundar-mudupalli-work:
Neil,

Yes, it does work. However we need some doc change(s):
1. The [validation reports section](https://github.com/GoogleCloudPlatform/professional-services-data-validator?tab=readme-ov-file#validation-reports) should clearly mention that for the Postgres result handler if the table does not exist the account must have `CREATE TABLE` privileges in order to create the table and write the results. The docs already say that you need to create a table for Big Query and provide the schema for the results table. We should also provide the schema if the customer wants to create the table before using DVT.
2. I am not clear on the right schema. This [line](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/b48d6668918d3c60ff700ab3ef2f61c636a18e13/terraform/results_schema.json#L121) says the schema has a column called `error_result`. I don't see that in the Postgres table. Does this json schema need to be updated ?
3. Part of the problem is that we are using two different mechanisms to write - one for Postgres and one for BigQuery. BigQuery has a limitation on the number of mutations per table per day. For customers running large number of Cloud Run jobs using `pandas_gbq.to_gbq` which is equivalent to `<dataframe>.to_sql` will not work.

Given there is a customer waiting and it is only doc changes, I have approved the PR.

Sundar Mudupalli

sundar-mudupalli-work (approved):
Please change the documentation as requested.

nj1973:
> Please change the documentation as requested.

For item 1 in your list I have added a link to the existing result handler setup documentation from the validation report section instead of duplicating the setup notes.

Item 2: I believe the `error_result` field is part of the BigQuery streaming and not applicable for a PostgreSQL table.

Item 3: Yes it is unfortunate that we can't reliably use load jobs for BigQuery (which would be a big performance improvement - https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1572). I've added a small note to the installation doc.

Issue #1613, opened 2025-10-20
Row Validation with BQ fails when a primary key column is a timestamp for Oracle, Postgres, Sqlserver and Teradata (MySQL succeeds)


nj1973:
As part of the work on [issue ](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1538) we added an end-to-end validation test for `generate-table-partitions`. Prior tests only generated partition config and didn't run the configs.

The new tests fail on Oracle and PostgreSQL.

Oracle:
```
E           TypeError: Arguments approved:boolean and Literal(N):string are not comparable
```

PostgreSQL:
```
E                   data_validation.exceptions.ValidationException: Some of the validations raised an exception

data_validation/__main__.py:417: ValidationException
------------------------------------------------------------------------- Captured log call --------------------------------------------------------------------------
ERROR    root:__main__.py:411 Error 'You are trying to merge on datetime64[ns] and datetime64[ns, UTC] columns. If you wish to proceed you should use pd.concat' occurred while running config file 0001.yaml. Skipping it for now.
ERROR    root:__main__.py:411 Error 'You are trying to merge on datetime64[ns] and datetime64[ns, UTC] columns. If you wish to proceed you should use pd.concat' occurred while running config file 0000.yaml. Skipping it for now.
```

Search for issue-1613 in integration tests after the Sybase implementation has been merged.

sundar-mudupalli-work:
Hi,

Post [PR 1619](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1619), both Oracle and Postgres fail with the following error when a timestamp field is presented as the primary key
```
11/12/2025 05:12:23 AM-ERROR: Error 'You are trying to merge on datetime64[ns] and datetime64[ns, UTC] columns. If you wish to proceed you should use pd.concat' occurred while running config file
```
This can be reproduced by the following commands:
`data-validation validate row -sc postgres -tc bq -tbls=pso_data_validator.test_generate_partitions_v2 -pk=recd_timestamp -hash='*'`
OR
`data-validation validate row -sc oracle21 -tc bq -tbls=pso_data_validator.test_generate_partitions_v2 -pk=recd_timestamp 
-hash='*'`

This is not specific to `test-generate-partitions`

Sundar Mudupalli

nj1973:
Sybase tests also have this same issue.

PR #1614, opened 2025-10-20
feat: Add Sybase support

### LINKED ISSUE #1538: Request for Sybase ASE support
We have a customer running both SQL Server and Sybase migrations to Cloud SQL for PostgreSQL. They can use DVT for the SQL Server migrations but not the Sybase ones.

The customer pointed out that Sybase is a fork of SQL Server and may not be too challenging to implement.

SQL Server and Sybase for source, PostgreSQL for target.

A new connection type for Sybase systems.

N/A

CLI commands on a GCE VM and Cloud Run jobs

**Additional context**
It appears that there two free to use versions of Sybase: Express and Developer.
https://hub.docker.com/r/nguoianphu/docker-sybase

nj1973:
## Description of changes
This PR:

- Adds Sybase support for DVT
- Add a limitations documentation page where we can build up a list of limitations
- Adds an internal test instance with Sybase ASE 16 and a full set of integration tests
- Adds retry logic for flaky test infrastructure (e.g. Db2 and Sybase)

In general I've referred to Sybase as Sybase ASE. There appears to be other flavours of Sybase out there (with different SQL/function support) but the customer interactions I've had have all related to Sybase ASE.

It's probably worth elaborating on Sybase limitations here:

- There is no SHA256 function available so we are restricted to concat and comp-fields row validation. (TODO create issue to explore custom UDF option and update text in registy.py)

## Issues to be closed

Closes #1538

## Checklist


tests/system/data_sources/common_functions.py
```diff
@@ -65,6 +67,13 @@
 ]
 
 
+def get_random_string(length=5):
```
sundar-mudupalli-work, line 70: This function is likely not required. See comment below

tests/system/data_sources/common_functions.py
```diff
@@ -648,6 +660,56 @@ def partition_query_test(
...
+def generate_and_run_table_partitions_test(
+    sc="bq-conn",
+    tc="bq-conn",
+    tables="pso_data_validator.test_generate_partitions_v2",
+    pk="course_id,quarter_id,recd_timestamp,registration_date,approved",
+    hash: str = "*",
+    concat: Optional[str] = None,
+    partition_num: int = 2,
+):
+    cdir = f"/tmp/yaml/{get_random_string()}"
```
sundar-mudupalli-work: If you want to have a file location, best to use `tmp_path` fixture that is available in pytest. Documentation [here](https://docs.pytest.org/en/stable/how-to/tmp_path.html) and an example usage [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/73baad71e2b1d78f6b6055547a89d726f38ae314/tests/system/data_sources/test_oracle.py#L228).

Sundar Mudupalli

tests/system/data_sources/common_functions.py
```diff
@@ -648,6 +660,56 @@ def partition_query_test(
     assert partition_filters[0] == expected_filter
 
 
+def generate_and_run_table_partitions_test(
```
sundar-mudupalli-work, line 663: Neil,

I don't think this test is needed. We already have tests:
1. (partition unit tests) that correct configs are generated
2. (partition unit tests) that files are correctly saved
3. (main unit tests) that configs are run
4. (partition system tests) that correct SQL statements are generated.

Issue 1613 turns out to be an issue with row validation with BQ when the primary key is a timestamp field. Most databases have this issue (Oracle, Postgres, Teradata and SQLServer). I have updated [the issue](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1613).

Sundar Mudupalli

tests/system/data_sources/test_sql_server.py
```diff
@@ -785,7 +785,7 @@ def test_row_validation_comp_fields_tricky_dates_to_bigquery():
 def test_row_validation_tricky_strings_to_bigquery():
     """Test with string values containing special characters."""
     pytest.skip(
-        "Skipping test_row_validation_tricky_dates_to_bigquery because the version of SQL Server we have does not support rtrim of all whitespace."
+        "Skipping test_row_validation_tricky_strings_to_bigquery because the version of SQL Server we have does not support rtrim of all whitespace."
```
sundar-mudupalli-work, line 788: Neil,

I am not sure of the context, the SQL standard does not require rtrim to trim all whitespace. The standard is that it should trim the pad character - which almost always is blank.

Sundar Mudupalli

sundar-mudupalli-work (changes requested):

tests/system/data_sources/test_sql_server.py
nj1973: The context is I was just fixing an incorrect name in a string, not modifying or adding a test. No change in functionality.

tests/system/data_sources/common_functions.py
nj1973: I think this new test is useful because item 4 in your list does not execute the final validation. It only compares the predicates to hardcoded literals. When a developer makes changes this relies on them manually testing on all engines and updating the constants.

When I added the new test it failed on multiple systems that appeared to be working fine because the hardcoded literal contained errors.

The new tests runs end-to-end and therefore gives extra confidence that the validation will actually work.

So I think the list then gets a 5th entry:

5. (partition end-to-end system tests) that the generated SQL is valid and will execute

tests/system/data_sources/common_functions.py
nj1973: Yep, good shout.

tests/system/data_sources/common_functions.py
```diff
@@ -648,6 +660,57 @@ def partition_query_test(
...
+    tmp_path: pathlib.Path,
```
sundar-mudupalli-work: I suggest using a different name, not `tmp_path` as it is confusing. You are using `tmp_path` as a parameter. In a pytest test (i.e. starting with prefix `test_`),  `tmp_path`  is a reference to the fixture, that is you are asking pytest to substitute that with a unique working directory when the test is run. That is not the case here.

sundar-mudupalli-work (approved):
There is a ton of code here that I don't really know how to review. Since all the tests pass and I did not see issues with code quality, it is OK as best as I can tell. If there is something specific that you want me to focus, please let me know.

PR #1619, opened 2025-10-31
fix: Include dialect when compiling predicates in partition builder

### LINKED ISSUE #1443: generate-table-partitions exception when a timestamp field is in primary-key
This is related to issue https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/923 but that issue is blocked due to Redshift dependencies and this has now been reported for an Oracle to PostgreSQL migration.

**Test tables**

Oracle:
```
DROP TABLE pso_data_validator.dvt_datetime_id;
CREATE TABLE pso_data_validator.dvt_datetime_id
(   id          DATE NOT NULL PRIMARY KEY
,   other_data  VARCHAR2(100)
);
COMMENT ON TABLE pso_data_validator.dvt_datetime_id IS 'Integration test table used to test datetime pk matching.';
INSERT INTO pso_data_validator.dvt_datetime_id VALUES (DATE'2020-01-01', 'Row 1');
INSERT INTO pso_data_validator.dvt_datetime_id VALUES (DATE'2020-02-01', 'Row 2');
INSERT INTO pso_data_validator.dvt_datetime_id VALUES (DATE'2020-03-01', 'Row 3');
INSERT INTO pso_data_validator.dvt_datetime_id VALUES (DATE'2020-04-01', 'Row 4');
INSERT INTO pso_data_validator.dvt_datetime_id VALUES (DATE'2020-05-01', 'Row 5');
COMMIT;
```

PostgreSQL:
```
DROP TABLE IF EXISTS pso_data_validator.dvt_datetime_id;
CREATE TABLE pso_data_validator.dvt_datetime_id
(   id          timestamp NOT NULL PRIMARY KEY
,   other_data  varchar(100)
);
COMMENT ON TABLE pso_data_validator.dvt_datetime_id IS 'Integration test table used to test datetime pk matching.';
INSERT INTO pso_data_validator.dvt_datetime_id VALUES (DATE'2020-01-01', 'Row 1');
INSERT INTO pso_data_validator.dvt_datetime_id VALUES (DATE'2020-02-01', 'Row 2');
INSERT INTO pso_data_validator.dvt_datetime_id VALUES (DATE'2020-03-01', 'Row 3');
INSERT INTO pso_data_validator.dvt_datetime_id VALUES (DATE'2020-04-01', 'Row 4');
INSERT INTO pso_data_validator.dvt_datetime_id VALUES (DATE'2020-05-01', 'Row 5');
```

**Test command**

```
data-validation generate-table-partitions  -sc ora  -tc pg \
 -tbls=pso_data_validator.dvt_datetime_id  \
--comparison-fields="*"  \
--config-dir /tmp/dtpart  --partition-num 2
...
    raise exc.CompileError(
sqlalchemy.exc.CompileError: No literal value renderer is available for literal value "datetime.date(2020, 3, 1)" with datatype DATE
```

nj1973:
## Description of changes
Ensure predicates are compiled with correct dialect so that correctly formatted literals are included in partition config files. This not only fixes the date formatting but also bugs related to booleans. Prior to this fix SQL Server predicates included `true`/`false` which should have been `1`/`0`.

This has a secondary benefit. In the past it has frustrated me that the SQL displayed in dry run mode doesn't quite match the real SQL we would execute. Now that we include the dialect we get accurate logged SQL.

## Issues to be closed

Closes #1443

## Checklist


data_validation/util.py
```diff
@@ -83,8 +87,13 @@ def ibis_table_to_sql(ibis_table: "ibis.expr.types") -> str:
     # If the backend uses sqlalchemy, we will need to request sqla to bind variables
     # for a non sqlalchemy backend, the parameters are already bound
     backend_name = ibis_table._find_backend().name
```
sundar-mudupalli-work: This is not a bug - I don't like us depending on an ibis internal function `_find_backend()`. Can this line be replaced    or removed. I recommend using `client.name` wherever `backend_name` is used.

sundar-mudupalli-work (changes requested):
Thank you Neil. Always better to use dialect when generating SQL. I don't know how I managed to work around this in the past - not too well.

Sundar Mudupalli

data_validation/util.py
nj1973: Makes sense. I guess it wasn't trivial before these changes because we didn't have the client available. But now we do - another nice side effect of this PR.

sundar-mudupalli-work (approved):
Looks Great. 

Thank you.

Sundar Mudupalli

PR #1620, opened 2025-11-06
feat: Support ODBC driver in --query arg

### LINKED ISSUE #1607: docs: SQL Server connection documentation does not show how to set driver name
SQL Server connection documentation does not show how to set driver name

https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/docs/connections.md#mssql-server

This is confusing users on how to get started.

nj1973:
## Description of changes
Modify MSSQL connection code to pick up "driver" from `--query` arg. Without this option the only way to provide a driver name is via a URL which is not ideal.

## Issues to be closed
Closes #1607

## Checklist

### DIFF
--- a/third_party/ibis/ibis_mssql/__init__.py
+++ b/third_party/ibis/ibis_mssql/__init__.py
@@ -23,6 +23,9 @@ from ibis.backends.mssql.datatypes import _type_from_result_set_info
 import json
 
 
+DEFAULT_DRIVER_NAME = "ODBC Driver 17 for SQL Server"
+
+
 # The MSSQL backend uses the Ibis MSSQL compiler, but overrides
 # the Backend class to use pyodbc instead of pymssql
 class Backend(BaseAlchemyBackend):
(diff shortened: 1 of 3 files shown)


third_party/ibis/ibis_mssql/__init__.py
```diff
@@ -51,9 +54,11 @@ def do_connect(
                 )
 
             if query:
-                query = json.loads(query)
+                query_dict = json.loads(query)
+                if "driver" not in query:
+                    query_dict["driver"] = odbc_driver or DEFAULT_DRIVER_NAME
```
helensilva14, line 59: Since `DEFAULT_DRIVER_NAME` already is the default value for the `odbc_driver` parameter, I'm wondering if putting it here on this condition is really needed 🤔

helensilva14 (changes requested):
This change was very needed! I had to do some local changes in the past when testing on my SQL Server instances from different environments with different drivers.

I added a minor comment about the current code, it could be something I missed.

third_party/ibis/ibis_mssql/__init__.py
nj1973: It is a bit of an edge case but, in theory, I could pass `odbc_driver=None` and then the code above would put `None` into the dict. The code above ensure that we don't leave that loophole open.

helensilva14: Ahh okay, makes sense!

helensilva14 (approved):

PR #1637, opened 2025-12-18
fix: Ensure we use BigQuery data project when prefixing tables

### LINKED ISSUE #1631: Quota project not being selected correctly
**Describe the bug**
Upon defining the 'quota_project_id' parameter, the BigQuery job is still being run by the project where the dataset belongs. I already have the fix ready to be pushed.

Target: BigQuery (source does not matter).

nj1973:
## Description of changes
When BigQuery execution project needs to be different to the data project we were not prefixing tables with the correct project.

This PR adds `self.data_project_id` to the BigQuery `Backend` to ensure correct prefixing when there are two projects in play. This is similar to in more recent Ibis versions but not quite the same because Ibis expects the dataset to define the project but we (DVT) don't expect users to prefix BigQuery datasets with the project. Plus `find-tables` does not include the data project.

Tested with schema, column and row validation across 3 projects. One billing/execution project and two data projects.

Custom query is not relevant because the user must prefix datasets with the project if they are anything other than the default.

## Issues to be closed

Closes #1631

## Checklist

### DIFF
--- a/data_validation/clients.py
+++ b/data_validation/clients.py
@@ -16,7 +16,7 @@
 from contextlib import contextmanager
 import copy
 import logging
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional
 import warnings
 
 import google.oauth2.service_account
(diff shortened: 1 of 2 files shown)


nehanene15 (approved):
LGTM!

PR #1641, opened 2026-01-12
fix: Db2 fixes for padded char identification and default IfNull replacement token

### LINKED ISSUE #1638: Db2 row validation of test col_char_2 columns fails
**Describe the bug**
While manually testing Db2 z/OS support I noticed the Db2 LUW row validation tests only hashed a single column.

I added all columns. into the validation and started getting errors like below:
```
sqlalchemy.exc.DataError: (ibm_db_dbi.DataError) ibm_db_dbi::DataError: Statement Execute Failed: [IBM][CLI Driver] CLI0109E  String data right truncation. SQLSTATE=22001 SQLCODE=-99999
```

I believe this is because the `coalesce(t1.cast__col_char_2, ?) AS ifnull__cast__col_char_2` expression overflows when the replacement token is too long for CHAR(2).

8.3.0

Db2 LUW

Add col_char_2 into test `test_row_validation_core_types`.

**Expected behavior**
Success.

nj1973:
## Description of changes
We had a customer with issues on Db2 with a CHAR padded primary key.

To fix this I needed to resolve part of issue 1470, although only for Db2 so this PR does NOT close that issue. Unfortunately the fix only covers table level validations. There does not appear to be a way to get the raw data type for a custom query via the IBM Db2 driver.

The small changes to files below are to remove deprecation warnings when customers run DVT, they should not change functionality:

- operations.py: `FutureWarning: Series.view is deprecated and will be removed in a future version. Use ``astype`` as an alternative to change the dtype.`
- combiner.py: `FutureWarning: A value is trying to be set on a copy of a DataFrame or Series through chained assignment using an inplace method.`

## Issues to be closed

Closes #1638

## Checklist


nj1973:
I've put this back to draft while I move some dependent changes from this PR to new issue https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1661

nj1973:
This PR is ready for review.

third_party/ibis/ibis_db2/__init__.py
```diff
@@ -102,24 +102,69 @@ def list_primary_key_columns(self, database: str, table: str) -> list:
...
+                    yield (
+                        colname,
+                        typename,
+                        col_length,
+                        col_length,
+                        col_length,
+                        col_scale,
+                        nullable,
+                    )
+        elif query:
```
sundar-mudupalli-work, line 151: From some preliminary investigation, it appears that you will need to call the stored procedure `SQLGetTypeInfo` to get type information on a query. This can be done via the `ibm_db_sa` and is likely not trivial - you need to open a cursor for the query and another cursor for calling the stored procedure. These cursors will need to be "closed" once the operation is completed. If you don't need to support custom query now, can you add this comment either to this PR or to issue 1470 for future reference.

Thanks.

third_party/ibis/ibis_db2/__init__.py
```diff
@@ -102,24 +102,69 @@ def list_primary_key_columns(self, database: str, table: str) -> list:
...
             list: A list of tuples containing the standard 7 DB API fields:
                   https://peps.python.org/pep-0249/#description
         """
-        return []
+        assert (database and table) or query, "We should never receive all args=None"
+        if database and table:
+            # For table-based validation, query the system catalog to get the true data type.
+            get_column_metadata_sql = """
+                SELECT COLNAME, TYPENAME, LENGTH, SCALE, NULLS
+                FROM SYSCAT.COLUMNS
```
sundar-mudupalli-work, line 125: According to the documentation, the catalog for z/OS is SYSIBM.SYSCOLUMNS. Perhaps you know that already. Is there anywhere in this code you are doing things for z/OS and have a flag that shows the version.

sundar-mudupalli-work (approved):
Thank you Neil. Excellent work. Some comments, nothing to change if you don't need to do it.

third_party/ibis/ibis_db2/__init__.py
nj1973: Yes, you are correct. This change is for Db2 LUW only, once the change is merged and across to my z/OS feature branch I'll make whatever modifications are required.

third_party/ibis/ibis_db2/__init__.py
nj1973: This is interesting information, I didn't make that discovery. Thanks.

I'll go ahead with the merge and log this on my z/OS issue to revisit. Thanks for the research.

PR #1649, opened 2026-01-20
feat: Add column name mappings option

### LINKED ISSUE #1617: Cater for column name mismatches across systems
Columns names occasionally differ between systems. For example when copying data from a transactional system to a data warehouse we may choose to give id columns a more descriptive name.

It would be good if this could be automated in some way.

- If DVT already assumed the columns are in the correct order then maybe it should honour the names
- If there is no expected order then perhaps we need a `--column-name-mappings` option to map the names
- This needs to work for primary keys too.

**Test case**
Oracle table:
```
create table dvt_test.tab1_a
(id number(5),col_data varchar2(30));
```

PostgreSQL table:
```
create table dvt_test.tab1_a
(tab_id int,col_data varchar(30));
```

DVT command:
```
data-validation validate column -sc=ora_local -tc=pg_local -tbls=dvt_test.tab1_a --count="*" --sum="*"
...
  File "/path/github/professional-services-data-validator/data_validation/config_manager.py", line 1059, in build_config_column_aggregates
    casefold_target_columns[column]
    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
KeyError: 'id'
```

8.1.1

Storing the validations to YAML files and then editing and correcting manually.

CLI commands

**Additional context**

FelipeKalinoski:
##PLEASE CAREFULLY REVIEW THE CHANGES BEFORE ACCEPTING IT##

## Description of changes
This PR implements support for column name mapping (`--column-name-map`) across row validation and partition generation. Key changes include:
- **ConfigManager Updates**: Modified `build_config_comparison_fields`, `build_dependent_aliases`, and `build_column_configs` to correctly resolve source column names to their target equivalents using the mapping provided.
- **Calculated Field Refactoring**: Updated `_get_calculated_config` and `_get_source/target_ibis_calculated_table` to iteratively apply calculated fields, resolving issues where dependencies (like `hash__all`) were not found when mappings were active.
- **System Testing**: Added system tests in `tests/system/data_sources/test_oracle.py` covering row validation with `--concat=*` and partition generation with mapped primary keys.
- **Unit Test Stability**: Patched `tests/unit/test__main.py` to correctly mock GCS directory listing, ensuring local test runs pass without requiring live Google Cloud credentials.

## Issues to be closed
Closes #1617

## Checklist

- [x] I have manually executed end-to-end testing (E2E) with the affected databases/engines (verified via mocked system tests and standalone demo script)

### DIFF
--- a/tests/system/data_sources/test_oracle.py
+++ b/tests/system/data_sources/test_oracle.py
@@ -14,11 +14,12 @@
 
 import os
 import pathlib
+import pandas as pd
+import ibis
 from unittest import mock
-
 import pytest
 
-from data_validation import cli_tools, data_validation, consts
+from data_validation import cli_tools, data_validation, consts, clients
 from tests.system.data_sources.common_functions import (
     DVT_TRICKY_DATES_COLUMNS,
     binary_key_assertions,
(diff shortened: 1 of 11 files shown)


nj1973:
I tested the changes and my previously passing test `test_column_validation_column_name_map_to_bigquery` now fails with a duplicate column error. I need to try and understand the changes you've made to figure out why my test now fails.

tests/system/data_sources/test_oracle.py
```diff
@@ -70,6 +71,59 @@
 }
 
 
+class MockIbisClient:
```
nj1973, line 74: Did you add this so you could test without BigQuery?

I tested `test_row_validation_column_name_map_to_bigquery` without this and it passed.

e.g.:
```
-@mock.patch("data_validation.clients.get_data_client", side_effect=mock_get_data_client)
+# @mock.patch("data_validation.clients.get_data_client", side_effect=mock_get_data_client)
 @mock.patch(
 )
-def test_row_validation_column_name_map_to_bigquery(mock_get_client):
+# def test_row_validation_column_name_map_to_bigquery(mock_get_client):
+def test_row_validation_column_name_map_to_bigquery():
```

Ideally I'd rather these integration tests run with real tables and not mocked ones.

I appreciate this makes it harder for you to test but it matches other tests in this file.

Can we remove the mocked client please?

nj1973 (changes requested):
Requested a change to oracle integration tests.

I also need to try and debug the changes for column validation which no longer works.

tests/system/data_sources/test_oracle.py
FelipeKalinoski: Yes, I did use mock tables for testing without BigQuery. We can delete it.

FelipeKalinoski:
> Requested a change to oracle integration tests.
> 
> I also need to try and debug the changes for column validation which no longer works.

@nj1973 did you create a unit test for testing column validation or were you executing a "manual" test?

sundar-mudupalli-work:
Some workarounds are offered in the [issue](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1617).

Unclear at this point that the customer prefers this change rather than building views and/or using custom queries.

Closing this pull request until it is clear how the issue is to be resolved.

Sundar Mudupalli

PR #1656, opened 2026-01-26
fix: revert breaking redshift change

nehanene15:
## Description of changes
Revert change, which caused no columns to be picked up:
`packages/ibis/expr/types/relations.py", line 1724, in select\n    raise com.IbisTypeError(\nibis.common.exceptions.IbisTypeError: You must select at least one column for a valid projection\n'`.

Also test updating to 'redshift' dialect since customer is getting NoSuchTable error due to internal catalog searching for PG metadata tables.

## Checklist


third_party/ibis/ibis_redshift/__init__.py
```diff
@@ -47,7 +47,7 @@ def do_connect(
             user=user,
             password=password,
             database=database,
-            driver=f"postgresql+{driver}",
+            driver=f"redshift+{driver}",
```
sundar-mudupalli-work: I am not sure why this change is being made. This line has not changed in 3 years. Can you explain ?

third_party/ibis/ibis_redshift/__init__.py
```diff
@@ -93,7 +93,7 @@ def _metadata(self, query: str) -> Iterable[Tuple[str, dt.DataType]]:
         if self.inspector.has_table(query):
             query = f"TABLE {query}"
         with self.begin() as con:
-            con.exec_driver_sql(f"CREATE VIEW {name} AS {query} WITH NO SCHEMA BINDING")
+            con.exec_driver_sql(f"CREATE VIEW {name} AS {query}")
```
sundar-mudupalli-work: This appears to be a revert, so is OK.

sundar-mudupalli-work (changes requested):

third_party/ibis/ibis_redshift/__init__.py
nehanene15: There's a customer seeing NoSuchTableError when the table exists in Redshift due to the driver not querying the correct metadata tables since PG and Redshift have slightly different internal catalog tables. Based on my research, it is caused bc it is not recommended to use the postrgresql+psycopg2 anymore but rather the redshift+redshift_connector.

Since psycopg2 is generally compatible, I'm trying to just update the dialect correctly to redshift. I'm also trying to find a Redshift test instance to test

sundar-mudupalli-work: I am going to approve it - though it might be best to test this branch, rather than roll it into dev and then test it.

Thanks.

Sundar Mudupalli

sundar-mudupalli-work (approved):

PR #1658, opened 2026-01-27
fix: Move global cast to string length limit of 3000 to Db2 specific code path

### LINKED ISSUE #1634: Db2 implementation technical debt
While working on adding Sybase support I had some confusion with a cast to `varchar(3000)` that I didn't implement. It turned out to be due to this line of code in the Db2 modifications:

https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/49ec473bf541a468aa400f5112875009c2b00a41/third_party/ibis/ibis_db2/datatypes.py#L36

This code is changing a global structure that is leaking into other supported engines.

It looks like we do something similar in the Oracle code too (although those lines were not manifesting in the Sybase testing I was doing).

It was further complicated by me seeing different results depending on which Python version I was using.

Either way - we should revisit this technique and consider an engine specific change rather than a global change.

nj1973:
## Description of changes

This PR changes how we currently define the target data type for a CAST to string in Db2.

Db2 has a conservative approach to estimating string lengths, assuming the worst case/longest possible length. At some point during initial Db2 implementation a change was implemented to apply a length of 3000 to all casts to string. While it was intended for this to be Db2 specific the change was actually global and active across other engines. I saw it active during Sybase and Oracle testing. 

This PR does not change the limit of 3000 on Db2 but modifies the location of the change to be Db2 specific. This comes with a risk though... what if by fixing one problem we create a new one in another engine that we didn't know was benefiting from the same code?

I've run all tests and they pass, I think that's all I can realistically do. But we should be prepared for subsequent changes if customer notice a negative impact.

In summary, there should be no functional change from this PR, we are only removing unintentional side effects.

PS 1: I also added a small change to prevent a Pandas 3 FutureWarning.

PS 2: I also used this PR to update the image we use for python-multi testing. Our testing on `develop` is currently using a deprecated image and Python 3.11 only. The `noxfile` changes here bring Python 3.10 back into our tested versions.

## Issues to be closed

Closes #1634

## Checklist


sundar-mudupalli-work (approved):
Neil - thank you for the excellent work in supporting DB2.

Sundar Mudupalli

PR #1660, opened 2026-01-27
feat: support role and warehouse params

### LINKED ISSUE #1659: Add 'role' support for Snowflake connections
Add support for 'role' and 'warehouse' being added to the Snowflake SqlAlchemy URL as documented [here](https://docs.snowflake.com/en/developer-guide/python-connector/sqlalchemy#connection-string-examples).

Latest

Snowflake to BQ

Customer is unable to connect without passing in a 'role'.

Cloud Run Job

nehanene15:
Adds optional fields of role and warehouse as **kwargs that the `ibis.snowflake.connect()` method supports. Needed for customer. Tested on Snowflake instance.

Closes #1659 

## Checklist

### DIFF
--- a/third_party/ibis/ibis_snowflake/api.py
+++ b/third_party/ibis/ibis_snowflake/api.py
@@ -35,38 +35,17 @@ def snowflake_connect(
 ):
     if connect_args:
         connect_args = dvt_config_string_to_dict(connect_args)
-    if role and warehouse:
-        return ibis.snowflake.connect(
-            user=user,
-            password=password,
-            account=account,
-            database=database,
-            connect_args=connect_args,
-            role=role,
-            warehouse=warehouse,
-        )
-    elif role:
-        return ibis.snowflake.connect(
-            user=user,
-            password=password,
-            account=account,
-            database=database,
-            connect_args=connect_args,
-            role=role,
-        )
-    elif warehouse:
-        return ibis.snowflake.connect(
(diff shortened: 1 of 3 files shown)


third_party/ibis/ibis_snowflake/api.py
```diff
@@ -29,10 +29,40 @@ def snowflake_connect(
     password: str,
     account: str,
     database: str,
+    warehouse: str = None,
+    role: str = None,
     connect_args: str = None,
 ):
     if connect_args:
         connect_args = dvt_config_string_to_dict(connect_args)
+    if role and warehouse:
```
nj1973: Just a thought, if we put all of those keyword args in a dict would we then be able to combine all of these in a single call, e.g.:

```
client_args = {"user": user, ...}
if role:
    client_args["role"] = role
if warehouse:
    client_args["warehouse"] = warehouse
return ibis.snowflake.connect(**client_args)
```

I've not tested it, just speculating :-)

nj1973 (approved):
I've approved anyway - but I think it would be worth seeing if my suggestion works and produces cleaner code. No probs if not.

third_party/ibis/ibis_snowflake/api.py
nehanene15: So smart, that worked! Updated it accordingly. Thanks 😄

PR #1662, opened 2026-01-30
fix: Trim default IfNull replacement token on Db2

### LINKED ISSUE #1661: Db2 IfNull replacement token overflows coalesce calls
In Db2 a coalesce of a column cannot exceed the length of the initial argument resulting in:
```
            String data right truncation. SQLSTATE=22001
```

On Db2 we need to ensure the replacement token is trimmed to the length of the value it is replacing.

nj1973:
## Description of changes
Introduces a Db2 specific workaround for coalesce limitation where the second argument cannot exceed the length of the initial argument resulting in:
```
            String data right truncation. SQLSTATE=22001
```

On Db2 we need to ensure the replacement token is trimmed to the length of the value it is replacing.

## Issues to be closed

Closes #1661

## Checklist


data_validation/config_manager.py
```diff
@@ -1318,6 +1352,19 @@ def build_dependent_aliases(self, calc_type: str, col_list=None) -> List[Dict]:
 
         column_aliases = {}
         col_names = []
+        # Get any ifnull token limits from either the source or target system.
+        # Most engines have no requirement for limits and will return an empty dict.
+        string_column_ifnull_limits = self._string_column_ifnull_limits(
```
cofin: This may be acceptable, but if the source and target are both DB2 and the target has different constraints here, then only the source limits are applied.

cofin (approved):
LGTM - I've made a note about one piece of the logic, but it probably is not an issue.  Feel free to resolve the comment.

data_validation/config_manager.py
nj1973: Good shout. I think it's safe to assume Db2 will only be used as one of the targets but it's better to the job right so I've made the suggested change. Thanks for pointing it out.

PR #1674, opened 2026-02-13
fix: Fix exceptions when validating Db2 XML columns

### LINKED ISSUE #1664: Db2 exception for column validation  of XML column
**Describe the bug**
Column validation exception for `--count` validation of XML column:

```
ibm_db_dbi.ProgrammingError: ibm_db_dbi:
:ProgrammingError: Statement Execute Failed: 
[IBM][CLI Driver][DB2/LINUXX8664] SQL0245N  The invocation of routine "LENGTH" is ambiguous. The argument in position "1" does not have a best fit.  SQLSTATE=428F5 SQLCODE=-245
```

Example command:
```
data-validation -v validate column -sc mydb2 -tc bq -tbls pso_data_validator.dvt_db2_types --count="col_xml"
```

8.4.0

Db2 LUW

nj1973:
## Description of changes
Before this PR all column count validations included a length function, for example:

```
count(length(col_string))
```

For Db2 XML this would throw an exception, we need to just issue count on the column.

On the flip side some engine data types, e.g. Oracle CLOB, require a length function to be included because count cannot be applied directly to the column.

For most data types the length function is unnecessary and adds extra CPU consumption for no benefit. For this reason I decided to change the logic from:

Always use `COUNT(LENGTH(COLUMN))` except for Db2 XML which will use `COUNT(COLUMN)`

to:

Always use `COUNT(COLUMN)` except for Oracle LOBs and SQL Server TEXT which will use `COUNT(LENGTH(COLUMN))`

This feels like a better default to me, with the most CPU expensive expression used by exception. Ideally we would do the same for MIN/MAX but that's a change for a different issue.

## Issues to be closed

Closes #1664

## Checklist


nehanene15 (approved):
LGTM!

PR #1677, opened 2026-02-20
docs: Documentation and samples updates

### LINKED ISSUE #1672: Fix mistakes in sample SQL Server Dockerfile
The curl command in https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/samples/docker/Dockerfile_sql_server_alpine is incorrect.

The correct commands are:
```
RUN curl -o ${PKGDIR}/msodbcsql.apk https://download.microsoft.com/download/9dcab408-e0d4-4571-a81a-5a0951e3445f/msodbcsql18_18.6.1.1-1_$architecture.apk
RUN curl -o ${PKGDIR}/mssql-tools.apk https://download.microsoft.com/download/b60bb8b6-d398-4819-9950-2e30cf725fb0/mssql-tools18_18.6.1.1-1_$architecture.apk
```

### LINKED ISSUE #1676: BigQuery connection option --client-project-id is missing from docs
We need to add `--client-project-id` to connections.md for BigQuery connections

nj1973:
## Description of changes
Just fixing a missing documentation entry and a few sample file issues.

## Issues to be closed

* Closes #1676
* Closes #1672

## Checklist


data_validation/cli_tools.py
```diff
@@ -75,7 +75,7 @@
         ["project_id", "GCP Project to use for BigQuery"],
         [
             "client_project_id",
-            "(Optional) BigQuery job/billing project (can differ from data project)",
+            "(Optional) BigQuery compute/billing project (can differ from data project)",
```
nehanene15: We should call this the 'quota project id'. The compute/slots will still be billed to the project specified by --project-id.  Could reference this in the docs - https://docs.cloud.google.com/docs/quotas/quota-project

nj1973: But this project_id is also used to define the client here:

https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/48f9ad0ad8f1459ba8f32cda54a5f56a87b1222a/data_validation/clients.py#L122

Which I believe ensures compute is assigned to that project too.

This allows users to have their billing/compute in a different project to the one that houses the data.

Or did I misunderstand how this works?

data_validation/cli_tools.py
```diff
@@ -75,7 +75,7 @@
         ["project_id", "GCP Project to use for BigQuery"],
```
sundar-mudupalli-work, line 75: I recommend changing this explanation to "GCP project holding the BigQuery dataset"

data_validation/cli_tools.py
sundar-mudupalli-work: After reviewing with Gemini - setting "client_project_id" effectively sets the billing (quota and slots) to "client_project_id". The only place we use this is get_bigquery_client. Typical behavior is that the billing - quota & slots is paid by the caller (client). When we provide this option, we are telling the user that they can use another project for billing. Therefore, I would prefer that we call this option billing_project_id and potentially accept client_project_id with a warning that customer should use billing_project_id and the option client_project_id may be deprecated in the future.

What do both of you think ?

data_validation/cli_tools.py
sundar-mudupalli-work: or BigQuery table - since we deal with tables mostly or dataset/table to be inclusive of how find-tables works.

sundar-mudupalli-work (changes requested):

data_validation/cli_tools.py
nj1973: This a used for everything if you don't override the compute/billing side of things with the optional parameter. If we state it is for the dataset/table then it would be misleading for cases where only a single project is in play.

How about:

```
Default project for BigQuery data and client (client project can be overridden with --client-project-id)
```

data_validation/cli_tools.py
nj1973: Client in `client_project_id` was picked because it felt like the most accurate term, it is the project used when creating the client used to submit jobs. The customer who asked for this extra control was not just concerned about getting slot usage in the correct project, they didn't even have BigQuery job submit privileges in the "data" project.

The option is also used to define quota because that appeared to make sense. I suppose it is possible for someone to want that extra control but I imagine it is not common.

It feels like this can be resolved by setting the option description to something unambiguous without changing the option name.

Maybe this is slightly more accurate:

```
"(Optional) Overrides the default project id when creating BigQuery clients. Provides extra control on which project BigQuery jobs are submitted to (including quota).",
```

data_validation/cli_tools.py
sundar-mudupalli-work: Maybe I am not understanding this - the "client project" (call it A) in BigQuery billing terms is the project to which the service account making the submit job request belongs. By default A pays for query while "project_id" pays for storage. So "project_id" is **not default** for billing. Cross project service accounts are not common, so the service account often belong to project_id.

nj1973: Yep, I agree with that description.

My choice of "default" in the wording is a DVT default. If the user does not provide "A" (which is an optional argument) then "project_id" will be used by DVT for both the data project and the one used when creating the client. This is the behaviour we've always had, one "project_id" to rule them all.

We recently added "A" to support customers who have the split project model. If the user does provide the optional "A" then "A" is used for the client and the original "project_id" option continues to be used for data project.

sundar-mudupalli-work: Hi,

Here is what I set up
1. data-project - has the dataset and the table
2. query-project - project that holds the service account that is querying bigQuery
3. billing-project - project that is being billed for the query.
4. One service account with privileges to run jobs in the billing project and query project and view the data in the data-project.
5. Two connections to bq specifying the data project, each with a different client project id that is query-project and billing project respectively.
This was done as follows:
```
gcloud iam service-accounts create bq-query-sa --display-name "BigQuery Query Service Account" --description "Service Account to test BigQuery billing" --project=query-project
gcloud projects add-iam-policy-binding query-project --member="serviceAccount:bq-query-sa@query-project.iam.gserviceaccount.com" --role="roles/bigquery.jobUser
gcloud projects add-iam-policy-binding billing-project --member="serviceAccount:bq-query-sa@query-project.iam.gserviceaccount.com" --role="roles/bigquery.jobUser"
gcloud projects add-iam-policy-binding data-project --member="serviceAccount:bq-query-sa@query-project.iam.gserviceaccount.com" --role="roles/bigquery.dataViewer"
gcloud iam service-accounts add-iam-policy-binding "bq-query-sa@query-project.iam.gserviceaccount.com" --member="user:my_ldap@google.com" --role="roles/iam.serviceAccountTokenCreator"
gcloud auth application-default login --impersonate-service-account=bq-query-sa@query-project.iam.gserviceaccount.com
data-validation connections add --connection-name bq1 BigQuery --project-id data-project --client-project-id query-project
data-validation connections add --connection-name bq2 BigQuery --project-id data-project --client-project-id billing-project
```
The following two `bq` commands were run, both successful in billing the query project and billing project respectively.
```
bq query --use_legacy_sql=false 'SELECT * from `pso-kokoro-resources.pso_data_validator.dvt_core_types`'
bq query --use_legacy_sql=false --project_id=billing-project 'SELECT * from `pso-kokoro-resources.pso_data_validator.dvt_core_types`'
```
The following 2 data validation commands were run, both throwing an error saying permission denied:
```
data-validation query -c bq1 -q 'select * from pso-kokoro-resources.pso_data_validator.dvt_core_types'
data-validation query -c bq2 -q 'select * from pso-kokoro-resources.pso_data_validator.dvt_core_types'
```
However when lines 121-127 in [clients.py](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/63c16a80aaad467046c71cdce7fef8e96c7bac23/data_validation/clients.py#L121) were modified as follows (commenting out other parameters):
```
    return bigquery.Client(
        project=effective_project,
        # client_info=info,
        # credentials=credentials,
        # default_query_job_config=job_config,
        # client_options=options,
    )
```
Both of the above commands ran correctly.

Here are changes we could make for clarity:
1. Fix the issue with clients.py - why is it complaining about permissions - something strange with the other arguments.
2. Allow the service account to determine the default billing project - rather than choosing the data project. This may require us to write out bigquery table names as data-project.dataset.table when converting queries to SQL. This could be a breaking change we reserve for the next release.
3. Clarify that the project_id refers to the location of the data and the optional billing_project can be provided to bill to a specific project. If not provided, then the project that is billed is the project where the service account is created.

Sundar Mudupalli

sundar-mudupalli-work:
Hi,

I have raised Issue #1710 and provided a PR #1715. This PR should be updated after that PR is closed.

Sundar Mudupalli

nj1973:
@sundar-mudupalli-work , as you are planning to deprecate the option I'll scale this PR back to only fixing the sample files.

nj1973:
> Hi,
> 
> I have raised Issue #1710 and provided a PR #1715. This PR should be updated after that PR is closed.
> 
> Sundar Mudupalli

I've scaled back the changes on this PR, ready for re-review @sundar-mudupalli-work

sundar-mudupalli-work (approved):

PR #1687, opened 2026-02-26
feat: Reduce Redshift queries

nehanene15:
Sets isolation_level to AUTOCOMMIT to avoid overhead and latency of transactional syntax i.e BEGIN, ROLLBACK, COMMIT. Standard and custom query validations use CREATE/DROP VIEW and validation SELECT statements which don't require multi-statement atomicity and would benefit from AUTOCOMMIT mode.

## Checklist


nj1973:
I'm not in favour of removing the pre ping for any other engines. I don't know enough about Redshift to understand why the customer sees the `SELECT 1` queries as a threat.

It is interesting the issue reported is "customer issue about too many sessions" because this was introduced as part of reducing the number of connections. We used to take a fresh connection for every validation but changed the code so that, if a DVT call needed to process 20 validations it would do them all in a single session. This had a knock on effect that sometimes sessions would "drop" (e.g. a firewall idle timeout) and then the whole DVT command would fail. By adding pre ping that dead session would be reconnected and the overall DVT would be able to continue.

Here is the PR for the change that added pre ping: https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1320

nehanene15:
Thanks Neil, that makes sense. After some research, the better option is to set transaction level to AUTOCOMMIT to reduce transactional overhead and because we don't need multi-statement atomicity. Updated the code to reflect that

nj1973:
I think we need to update the PR description to reflect the actual change, also needs `black` formatter running.

nehanene15:
> I think we need to update the PR description to reflect the actual change, also needs `black` formatter running.

Done. Thanks

sundar-mudupalli-work:
Neha,

Looks good to me. We don't have a redshift instance in our test suite to test. Have you tested this on a redshift instance ? The AUTOCOMMIT option is somewhat DB specific, so not sure if you are going to get any mileage out of it. However the change is limited to the redshift branch, so it is unlikely to affect any other engine.

I have approved it.

Thanks.

Sundar Mudupalli

sundar-mudupalli-work (approved):

nj1973 (approved):

PR #1694, opened 2026-03-04
chore: Bump black version

### LINKED ISSUE #1693: chore: Bump black version
Bump black version used to format code.

nj1973:
## Description of changes
The pinned version of black we use for formatting does not match the Black extension in VS Code which results in unnecessary manual executions of black to format using an older spec.

As we only test on Python 3.10+ I've picked the latest version of black which drops 3.9 support. We can cycle back if we think development on 3.9 is a requirement.

## Issues to be closed

Closes #1693

## Checklist


renzokuken (approved):
LGTM

sundar-mudupalli-work:
LGTM.

We are modeled on BigQuery Python Client. They use a slightly older version of BLACK - perhaps because they "support" 3.8? 

They have moved to pyproject.toml - which is probably a good change for us to make. I have opened a [separate issue](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1701).

Sundar Mudupalli

sundar-mudupalli-work:
Neil,

I agree. Thank you.

Black guarantees that its style (if it does) will only change at the beginning of the year. My suggestion would be to revisit the black version whenever we upgrade the python version - as we will when we upgrade to ibis 7.1.

Thanks.

Sundar Mudupalli

PR #1707, opened 2026-03-19
fix: Add raw datatype support for Db2 custom queries

### LINKED ISSUE #1691: Db2 raw data types are not detected for custom-query validations
Db2 raw data types are not detected for custom-query validations. See:

https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/04ad3a8551e8d55226c0b1de567dad9cd16d8e0b/third_party/ibis/ibis_db2/__init__.py#L153-L160

This means that custom-query validations as missing:

- Padded primary key detection for row validations
- BLOB and XML detection for row and agg validations

The problem stems from the fact that the SQL Alchemy cursor returned by the IBM driver does not specify real Db2 data types. This differs from most other SQL Alchemy drivers.

Sundar noted:

"it appears that you will need to call the stored procedure SQLGetTypeInfo to get type information on a query. This can be done via the ibm_db_sa and is likely not trivial - you need to open a cursor for the query and another cursor for calling the stored procedure. These cursors will need to be "closed" once the operation is completed."

We should investigate this potential solution or any other options. If it cannot be resolved then we should document the limitation.

nj1973:
## Description of changes
This PR solves an issue in our Db2 implementation where we couldn't get the raw data types for custom queries.

Describing the cursor did not reveal the actual data type, only the superset, e.g. `string`.

We solve the problem in the same way do on Redshift, by creating a view. If a view cannot be created then a warning is output and we fall back to the previous behaviour.

## Issues to be closed

Closes #1691

## Checklist


sundar-mudupalli-work (approved):
Thank you Neil

PR #1711, opened 2026-03-30
feat: Add caching mechanism for Backend functions

nehanene15:
As written by @nj1973 in docs/internal/design_docs/backend_caching_implementation.md: The `raw_column_metadata` method in `third_party/ibis/ibis_db2/__init__.py` (and the Redshift equivalent) executes a `CREATE VIEW` statement and queries the system catalog to retrieve column metadata. For custom queries, this operation is slow and often called multiple times during a single data validation run. The goal is to implement a simple in-memory cache at the `Backend` instance level to store the metadata results, avoiding redundant view creation and system queries.

Added the decorator to Redshift `_metadata()` as it creates a view every time it's called (typically 3x for custom query validations). This also changes Redshift syntax `with self.begin() as con:` to `with self.con.connect() as con:` since we removed transactional syntax in #1687 and using self.begin() creates a new transaction.

## Checklist

### DIFF
--- a/third_party/ibis/ibis_redshift/__init__.py
+++ b/third_party/ibis/ibis_redshift/__init__.py
@@ -22,6 +22,7 @@ from ibis.backends.postgres.datatypes import _BRACKETS, _parse_numeric, _type_ma
 
 from third_party.ibis.ibis_addon.api import cache_generator_results
 
+
 class Backend(BaseAlchemyBackend):
     name = "redshift"
     compiler = RedshiftCompiler
@@ -73,7 +74,7 @@ class Backend(BaseAlchemyBackend):
         super().do_connect(engine)
 
     def list_databases(self, like=None):
-        with self.begin() as con:
+        with self.con.connect() as con:
             # http://dba.stackexchange.com/a/1304/58517
             databases = [
                 row.datname
@@ -97,7 +98,7 @@ class Backend(BaseAlchemyBackend):
         """
         if self.inspector.has_table(query):
             query = f"TABLE {query}"
-        with self.begin() as con:
+        with self.con.connect() as con:
             con.exec_driver_sql(f"CREATE VIEW {name} AS {query} WITH NO SCHEMA BINDING")
             type_info = con.execute(
                 sa.text(type_info_sql).bindparams(raw_name=raw_name)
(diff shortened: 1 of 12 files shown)


third_party/ibis/ibis_redshift/__init__.py
```diff
@@ -82,6 +84,7 @@ def list_databases(self, like=None):
             ]
         return self._filter_with_like(databases, like)
 
+    @cache_generator_results
     def _metadata(self, query: str) -> Iterable[Tuple[str, dt.DataType]]:
         raw_name = util.guid().lower()
```
nj1973, line 88: This might be out of scope but it might be better to use `dvt_temp_object_name` for the name so it is obvious to DVT users why the view was created? Feel free to ignore this if you don't want to touch that part of the code.

third_party/ibis/ibis_redshift/__init__.py
```diff
@@ -95,7 +98,7 @@ def _metadata(self, query: str) -> Iterable[Tuple[str, dt.DataType]]:
         """
         if self.inspector.has_table(query):
             query = f"TABLE {query}"
-        with self.begin() as con:
+        with self.con.connect() as con:
             con.exec_driver_sql(f"CREATE VIEW {name} AS {query} WITH NO SCHEMA BINDING")
```
nj1973: This is again potentially out of scope but there is possibility here to leave views behind if we encounter an exception. Would it be worth including a try/catch block?

For example:

```
        with self.con.connect() as con:
            try:
                con.exec_driver_sql(f"CREATE VIEW {name} AS {query} WITH NO SCHEMA BINDING")
                type_info = con.execute(
                    sa.text(type_info_sql).bindparams(raw_name=raw_name)
                )
                yield from ((col, _get_type(typestr)) for col, typestr in type_info)
            finally:
                con.exec_driver_sql(f"DROP VIEW IF EXISTS {name}")
```

nj1973 (commented):
Just a couple of comments, both related to lines of code you haven't actually changed so you might choose to push back on these.

third_party/ibis/ibis_redshift/__init__.py
nehanene15: Done! Good idea

nj1973 (approved):
This LGTM but I think we should add Sundar as a reviewer too.

nj1973 (changes requested):
Just noticed tests are failing so switching status back to Request Changes

nj1973:
The failing tests are caused by a circular import triggered by my suggestion to use `util.dvt_temp_object_name` for the view name.

The problem is that `util.ibis_table_to_sql` imports some heavy packages (`clients` which imports pretty much everything in `thirdparty`) and therefore is not suited to be in a lightweight util module like this.

Ideally we would move `ibis_table_to_sql` somewhere else, `clients` perhaps. But for the sake of your PR you might want to just revert the name code back to what you had before. Up to you.

nj1973 (approved):
LGTM

sundar-mudupalli-work (approved):
Wow!!

Excellent work Neil Johnson and Neha for the quick follow

I learned a lot of python by reading your code

PR #1715, opened 2026-03-31
fix: Modify BQ connection to use the correct project id for billing.

### LINKED ISSUE #1710: DVT 's BQ Connection does not use the correct billing project.
**Describe the bug**
GCP BigQuery is a client based API for the purpose of determining the [billing project](https://docs.cloud.google.com/docs/quotas/quota-project#project-client-based). When DVT creates a BQ connection, DVT explicitly sets the billing project, resulting in incorrect behavior.
Specifically, the following behavior is observed BigQuery Connections
1. When a user account is used for ADC (`gcloud auth application-default login`), the command outputs `Quota project "xxx" was added to ADC which can be used by Google client libraries for billing and quota`. DVT's BQ connection does not use this project and forces the billing project to be the project where the data is stored. The user can specify a different billing project with the `--client-project-id` parameter. This is not standard.
2. When a service account is used for ADC (`gcloud auth application default login --impersonate-service-account <svc-account>`), DVT's BQ connection without a `--client-project-id` works. A connection which uses `--client-project-id` fails.
The use of `--client-project-id` for billing project is confusing. The term`--billing_project` is clearer.
Ex: 8.5.2

BigQuery for Source and Target

The most recent BQ queries in a project can be listed by using the command `bq ls -j -a -n 4 --project_id=xxx`. This command can be used to verify the project where the BQ query is being run.
1. Set up two projects, one data-project containing the dataset and an exec-project where BQ Queries can be run. Create two service accounts, one in each project. Give the service accounts read access to the BQ dataset (roles/bigquery.dataViewer) Give the service accounts BQ job roles (roles/bigquery.jobUser and roles/bigquery.readSessionUser) role on both projects. The user has these permissions and the roles roles/iam.serviceAccountTokenCreator on the service account (to impersonate service account).
2. Create 2 BQ Connections bq_data, just the name of data_project and bq_data_exec specifying the data_project and the exec_project as client_project_id. 
3. *User Credentials* 
    * a - run `gcloud config set project exec-project` # sets billing project to exec-project
    * b - run `gcloud auth application-default login` # executes further commands as the user. Command returns with the message `Quota project "exec-project" was added to ADC which can be used by Google client libraries`
    * c - execute `data-validation validate column -sc bq_data -tc bq_data -tbls=pso_data_validator.dvt_core_types`
    * d - Check which project is used for this validation using the bq command mentioned earlier.
    * e Confirm that DVT queries are run in the data project. DVT ignores the billing project set in the ADC. This is not normal GCP client behavior. DVT queries can be run in a different project by using `--client-project-id` argument while setting up the connection
4. *Service Accounts*
    * a run `gcloud auth application-default --impersonate-service-account <exec-project-service-account> login` # executes further commands as the service account in the exec project
    *  b run `data-validation validate column -sc bq_data -tc bq_data -tbls=pso_data_validator.dvt_core_types`. DVT fails with a `Access Denied: Project data-project: User does not have bigquery.jobs.create permission in project data-project.`
    *  c . run `data-validation validate column -sc bq_data_exec-project -tc bq_data_exec-project
-tbls=pso_data_validator.dvt_core_types` DVT fails with `Caller does not have required permission to use project dataproc-templates. Grant the caller the roles/serviceusage.serviceUsageConsumer role`
    * d run `gcloud auth application-default --impersonate-service-account <data-project-service-account> login` # executes further commands as the service account in the data project 
    * e run `data-validation validate column -sc bq_data -tc bq_data -tbls=pso_data_validator.dvt_core_types`. DVT succeeds.
    * f run `data-validation validate column -sc bq_data_exec -tc bq_data_exec -tbls=pso_data_validator.dvt_core_types`, where bq2 has the client_project_id set. DVT fails with `Caller does not have required permission to use project exec-project. Grant the caller the roles/serviceusage.serviceUsageConsumer role`

**Expected behavior**
For user accounts
* DVT's BQ connection should use the project output by the ADC credential for billing. DVT should discourage the use of `billing-project` for user accounts because setting it via `gcloud` is a standard approach.
For service accounts
* When the connection does not specify a `--billing-project` - DVT should use the project in which the service account is located. When a connection specifies a `--billing-project` that project should be used for billing.

**Additional context**
For user accounts invoking BigQuery client using `client = bigquery.Client()` uses the billing project specified in ADC credentials . For service accounts, the project which the service account belongs to is used. Invoking the BigQuery client using `client = bigquery.Client(project_id="exec-project")`, ensures that the project is billed to the exec-project.

sundar-mudupalli-work:
## Description of changes
As outlined in issue 1710, DVT forces the billing project resulting in incorrect (or non standard) behavior. This fix modifies the way DVT creates the BQ client, so that it behaves like other client libraries/programs. Additionally
* Change the name of the parameter being used to `--billing-project-id` for clarity
* Deprecated `--client-project-id` - users who use it will receive a warning and it will be accepted.
* Updated docs to clarify how to specify the billing project for user and service account calls.
* Deprecated the use of Service Account Key - users will receive a warning and it will be accepted.

## Issues to be closed

Closes #1710

## Checklist

- [X ] I have followed the [`CONTRIBUTING` Guide](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/CONTRIBUTING.md).

### DIFF
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -75,7 +75,11 @@ CONNECTION_SOURCE_FIELDS = {
         ["project_id", "GCP Project to use for BigQuery"],
         [
             "client_project_id",
-            "(Optional) BigQuery job/billing project (can differ from data project)",
+            "(Deprecated) BigQuery job/billing project (can differ from data project)",
+        ],
+        [
+            "billing_project_id",
+            "(Optional) BigQuery billing project to override default billing project",
         ],
         ["google_service_account_key_path", "(Optional) GCP SA Key Path"],
         [
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -81,7 +81,7 @@ CONNECTION_SOURCE_FIELDS = {
             "billing_project_id",
             "(Optional) BigQuery billing project to override default billing project",
         ],
-        ["google_service_account_key_path", "(Optional) GCP SA Key Path"],
+        ["google_service_account_key_path", "(Deprecated) GCP SA Key Path"],
         [
             "api_endpoint",
             '(Optional) GCP BigQuery API endpoint (e.g. "https://bigquery-mypsc.p.googleapis.com")',
@@ -153,7 +153,7 @@ CONNECTION_SOURCE_FIELDS = {
         ["project_id", "GCP Project to use for Spanner"],
         ["instance_id", "ID of Spanner instance to connect to"],
         ["database_id", "ID of Spanner database (schema) to connect to"],
-        ["google_service_account_key_path", "(Optional) GCP SA Key Path"],
+        ["google_service_account_key_path", "(Deprecated) GCP SA Key Path"],
         [
             "api_endpoint",
             '(Optional) GCP Spanner API endpoint (e.g. "https://spanner-mypsc.p.googleapis.com")',
--- a/data_validation/clients.py
+++ b/data_validation/clients.py
@@ -111,15 +111,13 @@ def get_google_bigquery_client(
     job_config = bigquery.QueryJobConfig(
         connection_properties=[bigquery.ConnectionProperty("time_zone", "UTC")]
     )
-    effective_project = quota_project_id or project_id
     options = None
     if api_endpoint or quota_project_id:
         options = client_options.ClientOptions(
             api_endpoint=api_endpoint,
-            quota_project_id=quota_project_id if quota_project_id else None,
         )
     return bigquery.Client(
-        project=effective_project,
+        project=quota_project_id,
         client_info=info,
         credentials=credentials,
         default_query_job_config=job_config,
@@ -152,20 +150,24 @@ def get_bigquery_client(
     credentials=None,
     api_endpoint: Optional[str] = None,
     storage_api_endpoint: Optional[str] = None,
-    client_project_id: Optional[str] = None,
+    client_project_id: Optional[str] = None, # to be deprecated in the future
+    billing_project_id: Optional[str] = None,
 ):
+    if client_project_id:
+        logging.warning("client_project_id is deprecated and will be removed in the future, use --billing-project-id instead")
+        billing_project_id = client_project_id
     google_client = get_google_bigquery_client(
         project_id,
         credentials=credentials,
         api_endpoint=api_endpoint,
-        quota_project_id=client_project_id,
+        quota_project_id=billing_project_id,
     )
     bqstorage_client = None
     if storage_api_endpoint:
         bqstorage_client = _get_google_bqstorage_client(
             credentials=credentials,
             api_endpoint=storage_api_endpoint,
-            quota_project_id=client_project_id,
+            quota_project_id=billing_project_id,
         )
 
     return bigquery_connect(
@@ -342,6 +344,10 @@ def get_data_client(connection_config):
             consts.GOOGLE_SERVICE_ACCOUNT_KEY_PATH
         )
         if key_path:
+            logging.warning(
+                """"Use of Service Account Keys is strongly discouraged. This option is deprecated and may be removed.
+                Run DVT with the service account identity by using application-default-credentials with --impersonate-service-account"""
+            )
             decrypted_connection_config["credentials"] = (
                 google.oauth2.service_account.Credentials.from_service_account_file(
                     key_path
--- a/data_validation/clients.py
+++ b/data_validation/clients.py
@@ -134,7 +134,6 @@ def _get_google_bqstorage_client(
     if api_endpoint or quota_project_id:
         options = client_options.ClientOptions(
             api_endpoint=api_endpoint,
-            quota_project_id=quota_project_id if quota_project_id else None,
         )
     from google.cloud import bigquery_storage_v1 as bigquery_storage
 
--- a/data_validation/clients.py
+++ b/data_validation/clients.py
@@ -150,11 +150,13 @@ def get_bigquery_client(
     credentials=None,
     api_endpoint: Optional[str] = None,
     storage_api_endpoint: Optional[str] = None,
-    client_project_id: Optional[str] = None, # to be deprecated in the future
+    client_project_id: Optional[str] = None,  # to be deprecated in the future
     billing_project_id: Optional[str] = None,
 ):
     if client_project_id:
-        logging.warning("client_project_id is deprecated and will be removed in the future, use --billing-project-id instead")
+        logging.warning(
+            "client_project_id is deprecated and will be removed in the future, use --billing-project-id instead"
+        )
         billing_project_id = client_project_id
     google_client = get_google_bigquery_client(
         project_id,
--- a/docs/connections.md
+++ b/docs/connections.md
@@ -123,7 +123,9 @@ data-validation connections add
     [--secret-manager-project-id SECRET_PROJECT_ID]     Secret Manager project ID
     --connection-name CONN_NAME BigQuery                Connection name
     --project-id MY_PROJECT                             Project ID where BQ data resides
-    [--google-service-account-key-path PATH_TO_SA_KEY]  (Deprecated) Path to SA key, use `gcloud auth application-default --impersonate-service-account` instead.
+    [--google-service-account-key-path PATH_TO_SA_KEY]  (Deprecated) Path to SA key, use
+                                                         gcloud auth application-default --impersonate-service-account <service-account> login
+                                                         instead.
     [--billing-project-id BILLING_PROJECT_ID]           (Optional) BigQuery billing project to override default billing project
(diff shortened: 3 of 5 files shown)


data_validation/cli_tools.py
```diff
@@ -75,7 +75,11 @@
...
         [
             "client_project_id",
-            "(Optional) BigQuery job/billing project (can differ from data project)",
+            "(Deprecated) BigQuery job/billing project (can differ from data project)",
+        ],
+        [
+            "billing_project_id",
+            "(Optional) BigQuery billing project to override default billing project",
         ],
         ["google_service_account_key_path", "(Optional) GCP SA Key Path"],
```
nj1973: Should this also say "Deprecated"?

data_validation/clients.py
```diff
@@ -111,15 +111,13 @@ def get_google_bigquery_client(
     job_config = bigquery.QueryJobConfig(
         connection_properties=[bigquery.ConnectionProperty("time_zone", "UTC")]
     )
-    effective_project = quota_project_id or project_id
     options = None
     if api_endpoint or quota_project_id:
         options = client_options.ClientOptions(
             api_endpoint=api_endpoint,
-            quota_project_id=quota_project_id if quota_project_id else None,
```
nj1973, line 119: ClientOptions has a quota_project_id arg:

https://googleapis.dev/python/google-api-core/latest/client_options.html

"quota_project_id (Optional[str]) – A project name that a client’s quota belongs to."

I recall you saying the arg didn't exist so just pointing out that it does. If we don't need to include it then I'm okay with this line being removed.

data_validation/clients.py
```diff
@@ -111,15 +111,13 @@ def get_google_bigquery_client(
...
         )
     return bigquery.Client(
-        project=effective_project,
+        project=quota_project_id,
```
nj1973, line 120: The previous code used `project_id` when the override was not included. The new code does not.

So if a user has access to projects "abc" and "xyz" and explicitly puts project "abc" in their connection file it will not be used. Is that the behaviour we want?

If it is then the `project_id` arg is unused and can be removed.

data_validation/clients.py
nj1973: For info this same line is included for the storage api client.

docs/connections.md
```diff
@@ -123,21 +123,22 @@ data-validation connections add
...
-    [--google-service-account-key-path PATH_TO_SA_KEY]  Path to SA key
+    [--google-service-account-key-path PATH_TO_SA_KEY]  (Deprecated) Path to SA key, use `gcloud auth application-default --impersonate-service-account` instead.
+    [--billing-project-id BILLING_PROJECT_ID]           (Optional) BigQuery billing project to override default billing project
     [--api-endpoint API_ENDPOINT]                       BigQuery API endpoint (e.g.
                                                         "https://bigquery-mypsc.p.googleapis.com)
     [--storage-api-endpoint STORAGE_API_ENDPOINT]       BigQuery Storage API endpoint (e.g.
                                                         "bigquerystorage-mypsc.p.googleapis.com)
                                                         Note this is a GRPC endpoint and does not
-                                                        include a URI scheme.
+                                                        includ  e a URI scheme.
```
nj1973: Typo?

data_validation/clients.py
sundar-mudupalli-work: There is a disconnect between BigQuery and DVT's (general database?) on how to reference a table. In DVT's view, the only reference is \<schema\>.\<table\>. In BigQuery it is \<project-id\>.\<dataset\>.\<table\>. DVT bridges this disconnect by using the \<project-id\> provided in the connection parameter to fully qualify the table name - I checked the generated SQL. DVT needs the `project_id` to fully qualified table name. The BQ client infers the project id from the environment if the table name is not fully qualified, is my understanding. DVT does not use this feature.

data_validation/clients.py
sundar-mudupalli-work: Sorry, I was not clear. I discovered that setting the quota_project_id in client_options resulted in some permission issues. Removing that line, resulted in expected Client API behavior. I am not sure if you might have the same problems when you test storage_api. If you have permission issues, you could consider removing that line and observing what happens.

Sundar Mudupalli

nj1973: I tested on the develop branch and also saw the same `serviceusage.services.use` permission exception. So I agree we need to remove this line.

I tested with and without it for the storage API client and could not observe any difference. So I think we should remove it from there too to be consistent.

data_validation/clients.py
nj1973: When I tested with a service account all was good in this respect. The jobs were submitted to the service account project.

When I tested with a user account with no billing project set in the DVT connection config things didn't behave as I would expect. And this is nothing really to do with "how to reference a table".

### User based scenario

I have two connections with no billing project specified.

```
$ cat bq_std1.connection.json
{"source_type": "BigQuery", "project_id": "my-project-01"}

$ cat bq_std2.connection.json
{"source_type": "BigQuery", "project_id": "my-project-02"}
```

And I run a validation:
```
data-validation validate column -sc=bq_std1 -tc=bq_std2 \
-tbls=dataset1.tab1=dataset1.tab1
```

My expectation was:

- Table 1: `my-project-01.dataset1.tab1`
- Table 2: `my-project-02.dataset1.tab1`
- Query job 1 submitted to `my-project-01`
- Query job 2 submitted to `my-project-02`

But what I experienced was

- Table 1: `my-project-01.dataset1.tab1`
- Table 2: `my-project-02.dataset1.tab1`
- Query job 1 submitted to `users-default-project`
- Query job 2 submitted to `users-default-project`

I think this matches your intention. But it is a significant step away from current behaviour, and unrelated to deprecating the client project parameter.

Is your concern that the current behaviour of setting `project=project_id` prevents the defaulting to a service account project id?

sundar-mudupalli-work: In #1710, I identified two behaviors which are below:

1. When a user account is used for ADC (gcloud auth application-default login), the command outputs `Quota project "xxx" was added to ADC which can be used by Google client libraries for billing and quota`. DVT's BQ connection does not use this project and forces the billing project to be the project where the data is stored. The user can specify a different billing project with the `--client-project-id` parameter. This is not standard.
2. When a service account is used for ADC (`gcloud auth application default login --impersonate-service-account <svc-account>`), DVT's BQ connection without a `--client-project-id` works. A connection which uses `--client-project-id` fails.
The use of `--client-project-id` for billing project is confusing. The term `--billing_project` is clearer.

As you observed, there is no way to set cross project billing for service accounts. Project-A contains the quota and slots and Project-B contains the service account. This is not uncommon especially with customers who purchase slots. DVT does not support this, but the BQ and other (Spanner) client libraries support this functionality.

The functionality for user accounts is an artifact of how application default credentials are set up. DVT uses a non-standard approach. User accounts tend to have excess permissions most of the time, so this has not bubbled up as an issue.

Whichever project you are billing you need to have permissions to run the job. Permission issues are very difficult to troubleshoot and resolve. So it would be good to follow the standard approach.

Sundar Mudupalli

data_validation/clients.py
nj1973: As mentioned in this comment https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1715/changes#r3051553658 I think we should remove this line.

data_validation/clients.py
nj1973: My example was not really talking about billing, it was about the project where the BigQuery job is submitted, where it is visible in the job history.

In my example I have 2 BigQuery projects containing the same table. I want to validate the table across the two projects.

I believe, with your changes, both query jobs will be submitted to a single project, whichever I have configured in my gcloud config. Nothing to do with billing, just where the job runs.

I just want to make sure that is a conscious decision. If a user (not service account) wants to control which project they consume slots in, and which project they can see their query in job history they'll need to use the new billing option in each connection file.

nj1973 (changes requested):

data_validation/clients.py
sundar-mudupalli-work: Neil,

Billing project = where the BQ query job runs.

1. The `connections.md` says that `project_id` is the project where the dataset resides. Until now, `project_id` is also where the BQ query job runs. If this change is accepted, the project where BQ query job runs is used from the environment unless explicitly over-riden by the `billing_project_id` parameter.
2. Users would have configured their billing project in their ADC for using gcloud commands. DVT (and other client libraries) reuse that setting. This should make things easier (I hope), not harder. We are not tying the connection string to a billing account unless the user explicitly wants to override default behavior. 

Sundar Mudupalli

data_validation/clients.py
nj1973: I've removed this line to close this thread.

nj1973 (approved):
LGTM. I did remove the line I requested you to remove from the storage API client so please review that before clicking merge.

PR #1721, opened 2026-04-02
feat: Add Db2 z/OS support

### LINKED ISSUE #1630: Db2 z/OS support
When testing DVT on Db2 z/OS many features work but some do not.

Creating issue/branch to track this workstream.

8.3.0

Db2 11.5 on z/OS

Functional support as close as possible to Db2 LUW

nj1973:
## Description of changes
Adds support for Db2 z/OS and makes the distinction that our existing support was for Db2 LUW.

In the most part Db2 z/OS shares code with LUW. We've added a new Backend because there are some differences, but not a huge amount. The underlying ibm_db/ibm_db_sa drivers are used for both.

Notable differences:

- Sorting by rand() is not supported, we needed a specific random row sample query that selects rand and sorts by the column alias.
- LOB & XML data types are not as flexible on z/OS and needed extra restrictions.
- Catalog SQL for data type detection was slightly different.
- On Db2 z/OS we saw two hidden/auto generated columns being returned by the IBM driver, `db2_generated_docid_for_xml` and `db2_generated_rowid_for_lob`. We added code to exclude these from the Ibis tables because we could not influence the SQL used by the driver.
- Db2 z/OS ROWID data type is unsupported and causes an exception. The customer who needs this support did not use ROWIDs explicitly and therefore this was not a blocker.
- Generated SQL statements do not support bind variables. A few functions had to be changed to explicitly include tokens that would otherwise be changed to variables, for example the replacement text used with `coalesce`.

As a side task I moved some Impala/Hive code out of `operations.py` into dedicated files.

Unfortunately we have no access to a Db2 z/OS test system. For the most part I've been able to rely on testing code using LUW. I have also benefited from Db2 z/OS testing by the customer who requested this work.

## Issues to be closed

Closes #1630

## Checklist


third_party/ibis/ibis_db2/__init__.py
sundar-mudupalli-work: I think it is a bad idea for us to set the default value for this port - this is true for all dbs. The underlying driver sets the default port, so we don't have to. If there is every a change in the default in the future, you need to update this.

sundar-mudupalli-work:
LGTM. Unfortunately, we can't test these change nightly.

third_party/ibis/ibis_db2/__init__.py
nj1973: That's not actually a line I changed. Although I did copy that line into the z/OS version of the same method.

I've removed the port default from both implementations.

sundar-mudupalli-work (approved):

PR #1741, opened 2026-05-22
chore: Update black to most recent version

### LINKED ISSUE #1740: Update black package to newer version
There is a vulnerability in `black==26.1.0` which is the version we include in `setup.py`.

It is only part of our developer/test dependencies so low risk, but an easy fix nonetheless. We should update.

nj1973:
## Description of changes
Update black formatter due to vulnerability in current version.

## Issues to be closed

Closes #1740

## Checklist

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -40,7 +40,7 @@ BLACK_PATHS = (
     "noxfile.py",
     "setup.py",
 )
-LINT_PACKAGES = ["flake8", "black==26.1.0"]
+LINT_PACKAGES = ["flake8", "black==26.5.1"]
 UNIT_PACKAGES = ["pyfakefs", "freezegun", "teradatasql"]
 
 
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,7 @@ extras_require = {
     "apache-airflow": "1.10.11",
     "pyspark": "3.0.0",
     "develop": [
-        "black==26.1.0",
+        "black==26.5.1",
         "flake8",
         "freezegun",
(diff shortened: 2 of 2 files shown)


sundar-mudupalli-work (changes requested):
Neil,

I am not sure why we are pinning this to a specific version of black. I would suggest we let the version float. 
Looking at the history, this was updated in [PR 1694](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1694). The original black version was introduced at the beginning of time (Commit date Mar 2020, first PR April 2020).

So I recommend we let the version of black float unless there is a sepcific reason to not do so.

If we want to check for multiple specific versions of python, we can tell black to do so in noxfile.py. Not sure if you want to take that up as a change in this PR.

Thank you.

Sundar Mudupalli

nj1973:
> I am not sure why we are pinning this to a specific version of black. I would suggest we let the version float. Looking at the history, this was updated in [PR 1694](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1694). The original black version was introduced at the beginning of time (Commit date Mar 2020, first PR April 2020).
> 
> So I recommend we let the version of black float unless there is a sepcific reason to not do so.

We pin it to give us a stable test environment. If it floated then code I wrote/submitted/tested on `develop` one day might fail to pass the lint check the day after (if `black` released a new version).

We need `black` to be stable and only bumped after our code is reformatted using the new version.

sundar-mudupalli-work (approved):
Thank you.

PR #1742, opened 2026-05-22
feat: Add --config-dir/--config-dir-json to table validation commands

### LINKED ISSUE #1205: validation commands could support config-dir option
Currently we can pass multiple tables to a validation command, these will be executed sequentially. We can also take a config file path via `--config-file` which will generate a single YAML file containing the multiple validations. This single YAML file can then be processed and will execute the validations sequentially.

It has been suggested by an end user that validation commands could also support the `--config-dir` option and then create one YAML file per validation rather than adding them all to a single file.

This has a couple of advantages:

- Validations can be executed independently, e.g. if one validation is problematic it can be excluded or edited more easily
- Validations can be executed in parallel
- We support `--config-dir` for `data-validation configs run` commands therefore it is consistent to also support it for file generation too

Disadvantages:

- Do we need to worry about overwriting files that already exist?

In this case the end user achieved what they wanted with a shell script (provided by us), snippet below:
```
SIMPLE_LIST=$(echo ${TABLES}|python3 -c 'import sys, json; data = json.load(sys.stdin); print(" ".join(["{}.{}".format(_["schema_name"], _["table_name"]) for _ in data]))')
for t in $SIMPLE_LIST;do
    echo "Processing ${i}"
    data-validation validate column --source-conn=${SC} --target-conn=${TC} \
      --tables-list="${t}" --config-file=${CONFIG_DIR}/${t}.yaml \
      --sum="*" --min="*" --max="*" --count="*"
done
```

If we implement this feature request we can simplify this process for them.

nj1973:
## Description of changes
This PR adds `--config-dir`/`--config-dir-json` to table validation commands. We already have:

- `--config-dir` for `generate-table-partitions` commands to write multiple YAML config files to a single directory
- `--config-file`/`--config-file-json` for table validation commands to write all validations to a single file

This PR takes the approach from `generate-table-partitions` and applies it to normal table validations.

This PR adds the functional equivalent of this shell snippet to the Python code:
```shell
SIMPLE_LIST=$(echo ${TABLES}|python3 -c 'import sys, json; data = json.load(sys.stdin); print(" ".join(["{}.{}".format(_["schema_name"], _["table_name"]) for _ in data]))')
for t in $SIMPLE_LIST;do
    echo "Processing ${i}"
    data-validation validate column --source-conn=${SC} --target-conn=${TC} \
      --tables-list="${t}" --config-file=${CONFIG_DIR}/${t}.yaml \
      --sum="*" --min="*" --max="*" --count="*"
done
```

The shell code above is fine when using DVT locally but not ideal when DVT is hosted for Cloud Run. Having the functionality enables customers to feed `find-tables` output into a Cloud Run hosted DVT without the bootstrap/round trip penalties.

We also add a design doc describing current behaviour for all config-file/dir options (including the PR). This includes potential future enhancements. It is a good basis for an agent to understand current behaviour before making further changes.

Out of scope:
- We already support `--config-dir` for `configs run` commands therefore no changes were required there.
- I decided against adding an equivalent to `--parts-per-file` for table validations for now. If this PR is approved I will add an issue to deprecate `--parts-per-file` and replace it with `--configs-per-file`.

## Issues to be closed

Closes #1205

## Checklist


docs/internal/design_docs/config_management.md
```diff
@@ -0,0 +1,111 @@
...
+### C. Nested Directory-Based Configuration (Partitioned Validations)
+The `generate-table-partitions` command splits a single large table validation (specifically `row` validations) into multiple chunked validations based on primary keys. It **requires** directory-based storage.
+
+*   **Directory Structure**: Nested. DVT automatically creates a subdirectory for each table.
+    *   Structure: `<config_dir>/<schema>.<table_name>/`
+    *   Example: `my_partitions/dbo.customers/`
+*   **Naming Convention**: Files inside the subdirectory are sequentially numbered: `0000.yaml`, `0001.yaml`, `0002.yaml` up to `9999.yaml`.
+*   **Parts Per File**: The `--parts-per-file` (or `-ppf`) option (default: 1) controls how many partition validation blocks are packed into a single YAML file.
+*   **Overwrite Protection**: Unlike standard validations, `generate-table-partitions` **does not** enforce an empty directory. It will write or overwrite files in the target directory directly.
+*   **File Limit**: DVT limits the number of generated files to 10,000 per run. If `partition-num / parts-per-file` exceeds 10,000, it raises a parser error.
```
sundar-mudupalli-work, line 59: One reason is DVT expects that users will use Cloud Run Jobs to run each validation as a Cloud Run task. Cloud Run has a limit of 10,000 tasks in a job. So having more than 10,000 YAML files would not work well with Cloud Run. With multiple parts per file, almost any size table can fit into 10,000 YAML files.

sundar-mudupalli-work (commented):
Neil,

I am not sure about this feature. Do you think it is valuable or is it a quick win that may cause confusion later ?

Today, I can do `data-validation validate column -tbls=pso_data_validator.dvt_core_types,pso_data_validator.dvt_ora2pg_types -sc postgres -tc postgres -c ~/test123.yaml` and this command will put the validation for both tables in the same file. When passed to Cloud Run the validations would run sequentially. This is usually not an issue since column validations are pretty fast. If these are row validations, the tables are small enough to fit in memory, so it should be fast as well.

I had a customer (no kidding!!) write a command one time which was something like this
`data-validation validate column -sc ora -tc pg --tables-list=``"data-validation find-tables -sc ora -tc pg  --allowed-schemas ${SCHEMA} |jq ..."`` ...`.  I am pointing this out since comparing multiple tables in one command is something people use.

Thanks.

Sundar Mudupalli

nj1973:
> Neil,
> 
> I am not sure about this feature. Do you think it is valuable or is it a quick win that may cause confusion later ?
> 
> Today, I can do `data-validation validate column -tbls=pso_data_validator.dvt_core_types,pso_data_validator.dvt_ora2pg_types -sc postgres -tc postgres -c ~/test123.yaml` and this command will put the validation for both tables in the same file. When passed to Cloud Run the validations would run sequentially. This is usually not an issue since column validations are pretty fast. If these are row validations, the tables are small enough to fit in memory, so it should be fast as well.
> 
> I had a customer (no kidding!!) write a command one time which was something like this `` data-validation validate column -sc ora -tc pg --tables-list=``"data-validation find-tables -sc ora -tc pg --allowed-schemas ${SCHEMA} |jq ..."`` ... ``. I am pointing this out since comparing multiple tables in one command is something people use.
> 
> Thanks.
> 
> Sundar Mudupalli

We've had two customers ask for this in the past and in both cases we had to resort to a workaround like the shell snippet in the PR description. One of the customers was not using Cloud Run Jobs, they had their own Kubernetes system and wanted to plan the validation ahead of time, for example:

1) Create YAMLs for all tables
2) Test the individual YAMLs and refine/edit/recreate any files with issues
3) Repeat 2 until happy
4) Schedule DVT `configs run` commands on the tested YAMLs in Kubernetes

We don't currently support step 1, users have to craft it themselves. But we do have it for generate table partitions which is also confusing.

"valuable" is debatable. Where I think that has value is in planning validations as described above, especially in an agentic flow. It is much more deterministic to have `--config-dir` than ask an agent to create a shell script and hope it gets it right first time.

I do think a `--configs-per-file` option will be needed for Could Run Jobs (as mentioned in the PR description), I was just saving that for a second issue. Happy to pull it into this one if we feel they need to arrive together.

sundar-mudupalli-work:
Neil,

If more than one customer has asked for this and it looks like the workaround is not very helpful. Thank you for doing this.


Sundar Mudupalli

sundar-mudupalli-work (approved):
LGTM. Thank you.

PR #1747, opened 2026-06-02
fix: expand hash/concat wildcards in YAML validation path

### LINKED ISSUE #1743: hash: '*' in YAML config crashes with TypeError: reduce() of empty iterable (works on CLI, not YAML)
### Summary

`--hash '*'` works on the CLI but the equivalent `hash: '*'` in a YAML config crashes at execution. The wildcard is expanded into the table's columns only on the CLI path; the YAML loader leaves the literal `'*'` in place, so `calculated_fields` is never populated and downstream code fails with:

```
TypeError: reduce() of empty iterable with no initial value
```

### Reproduction

```yaml
result_handler: {}
source: my_source
target: my_target
validations:
- type: Row
  schema_name: mydataset
  table_name: mytable
  target_schema_name: mydataset
  target_table_name: mytable
  primary_keys:
  - {cast: null, field_alias: id, source_column: id, target_column: id}
  hash: '*'
  format: table
  use_random_rows: false
```

```
data-validation configs run -c repro.yaml
# -> TypeError: reduce() of empty iterable with no initial value
```

The same validation built directly with `data-validation validate row --hash '*' ...` works fine, and the YAML it emits (with `calculated_fields:` and `comparison_fields:` materialized) also works when re-run. The bug only hits hand-authored YAML.

### Root cause

Two code paths build a `ConfigManager`, and only the CLI path expands wildcards.

**CLI path** — `data_validation/__main__.py:337-359` `build_config_managers_from_args` → `build_config_from_args` (line 250) → `_get_calculated_config` (line 179):

```python
# __main__.py:183-189
if config_manager.hash:
    col_list = (
        None
        if config_manager.hash == "*"
        else cli_tools.get_arg_list(config_manager.hash)
    )
    fields = config_manager.build_dependent_aliases("hash", col_list)
```

`build_dependent_aliases(..., col_list=None)` (`config_manager.py:1432`) falls through `_filter_columns_by_column_list` to every column on the ibis table, producing the `calculated_fields` entries and appending `hash__all` to `comparison_fields`.

**YAML path** — `data_validation/__main__.py:431-456` `build_config_managers_from_yaml`:

```python
for config in yaml_configs[consts.YAML_VALIDATIONS]:
    ...
    config_manager = ConfigManager(
        config, source_client, target_client, verbose=args.verbose
    )
    config_manager.config[consts.CONFIG_FILE] = config_file_path
    config_managers.append(config_manager)
```

The dict is wrapped as-is — `_get_calculated_config` is never invoked, so the literal `'*'` stays in `config_manager.hash`, `calculated_fields` is `[]`, and the subsequent `functools.reduce(...)` over that empty list raises the `TypeError`.

### Suggested fix

Run the same expansion in the YAML loader, gated on `calculated_fields` being empty so that regenerated YAML files (which already contain materialized fields) don't get re-expanded:
```python
      # data_validation/__main__.py — build_config_managers_from_yaml
      for config in yaml_configs[consts.YAML_VALIDATIONS]:
          config[consts.CONFIG_SOURCE_CONN] = source_conn
          config[consts.CONFIG_TARGET_CONN] = target_conn
          config[consts.CONFIG_RESULT_HANDLER] = yaml_configs[consts.YAML_RESULT_HANDLER]
          config_manager = ConfigManager(
              config, source_client, target_client, verbose=args.verbose
          )
          config_manager.config[consts.CONFIG_FILE] = config_file_path
       #Expand `hash: '*'` / `concat: '*'` for hand-authored YAML.
          if (
              config_manager.validation_type == consts.ROW_VALIDATION
              or (
                  config_manager.validation_type == consts.CUSTOM_QUERY
                  and config_manager.custom_query_type == consts.ROW_VALIDATION.lower()
              )
          ) and not config_manager.calculated_fields:
              config_manager.append_calculated_fields(
                  _get_calculated_config(args, config_manager)
              )

          config_managers.append(config_manager)
  ```

Why it differs from the CLI

  -  args.custom_query_type → config_manager.custom_query_type: same value, but on the YAML path it lives on the ConfigManager, not on argparse.
  -  Extra "and not config_manager.calculated_fields" guard: YAML files may already contain materialized calculated_fields from a prior CLI run; without the guard we'd double-expand. (I.e. _get_calculated_config appends, never replaces, and validate ... -c file.yaml writes a YAML containing both the wildcard and the materialized fields. The guard
  tells the YAML loader: "if the fields are already there, don't expand again."). CLI input can never carry pre-existing fields, so it doesn't need this check. 

I'm willing to send a PR with the patch + a regression test in `tests/unit/test__main.py` if that's helpful.

Aaditya-git:
## Summary

- `build_config_managers_from_yaml()` never called `_get_calculated_config()`, so a hand-authored YAML config with `hash: '*'` (or `concat: '*'`) left the literal `'*'` unexpanded
- Downstream code then tried to `reduce()` over an empty list and crashed
- Fix calls `_get_calculated_config()` for row and custom-query-row validations, gated on `calculated_fields` being empty to avoid double-expanding regenerated YAML that already has materialized fields

Fixes #1743

## Test plan

- [x] New unit test `test_build_config_managers_from_yaml_expands_hash_wildcard` added to `tests/unit/test__main.py`
- [x] Verifies `_get_calculated_config` is called exactly once and its result is appended to `calculated_fields`
- [x] Full `tests/unit/test__main.py` suite: 21/21 passing

### DIFF
--- a/data_validation/__main__.py
+++ b/data_validation/__main__.py
@@ -451,6 +451,23 @@ def build_config_managers_from_yaml(args, config_file_path):
             config, source_client, target_client, verbose=args.verbose
         )
         config_manager.config[consts.CONFIG_FILE] = config_file_path
+
+        # Expand hash/concat wildcards for hand-authored YAML configs.
+        # The CLI path calls _get_calculated_config() which expands `hash: '*'` into
+        # per-column calculated_fields entries. The YAML path skips that step, leaving
+        # the literal '*' in place and causing a downstream reduce() on an empty list.
+        # Guard on calculated_fields being empty so regenerated YAML (which already
+        # contains materialized fields) is not double-expanded.
+        is_row_validation = config_manager.validation_type == consts.ROW_VALIDATION
+        is_custom_row = (
+            config_manager.validation_type == consts.CUSTOM_QUERY
+            and config_manager.custom_query_type == consts.ROW_VALIDATION.lower()
+        )
+        if (is_row_validation or is_custom_row) and not config_manager.calculated_fields:
+            config_manager.append_calculated_fields(
+                _get_calculated_config(args, config_manager)
+            )
+
         config_managers.append(config_manager)
 
     return config_managers
--- a/tests/unit/test__main.py
+++ b/tests/unit/test__main.py
@@ -605,3 +605,59 @@ def test_successful_generate_partitions_with_mocked_partition_builder(
 )
 def test_successful_deploy_with_mocked_app_run(mock_args, mock_run):
     main.main()
+
+
+# Regression test for https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1743
+# YAML configs with `hash: '*'` crashed because build_config_managers_from_yaml() never
+# called _get_calculated_config(), leaving the wildcard unexpanded.
+ROW_YAML_HASH_WILDCARD_CONFIG = {
+    consts.CONFIG_TYPE: consts.ROW_VALIDATION,
+    consts.CONFIG_SCHEMA_NAME: "my_schema",
+    consts.CONFIG_TABLE_NAME: "my_table",
+    consts.CONFIG_PRIMARY_KEYS: [],
+    consts.CONFIG_ROW_HASH: "*",
+}
+
+ROW_YAML_VALIDATIONS = {
+    consts.YAML_SOURCE: "my_source",
+    consts.YAML_TARGET: "my_target",
+    consts.YAML_RESULT_HANDLER: None,
+    consts.YAML_VALIDATIONS: [ROW_YAML_HASH_WILDCARD_CONFIG],
+}
+
+
+DUMMY_CALCULATED_FIELD = {"name": "hash__all", "calc_type": "hash"}
+
+
+@mock.patch("data_validation.clients.get_data_client", return_value=MockIbisClient())
+@mock.patch(
+    "data_validation.state_manager.StateManager.get_connection_config",
+    return_value=TEST_CONN,
+)
+@mock.patch(
+    "data_validation.cli_tools.get_validation",
+    return_value=ROW_YAML_VALIDATIONS,
+)
+@mock.patch(
+    "data_validation.__main__._get_calculated_config",
+    return_value=[DUMMY_CALCULATED_FIELD],
+)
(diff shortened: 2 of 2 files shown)


sundar-mudupalli-work:
Aditya,

Thank you for submitting this PR. I have provided my comments in the issue [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1743#issuecomment-4617296970). Let us continue our discussion there.

I see a lot of challenges with this approach - some covered in the issue discussion.

I am going to pause this PR until we can have a full understanding of the use case you are experiencing.

Sundar Mudupalli

Aaditya-git:
Thank you for the detailed explanation, Sundar. I have been thinking about this since the past 12 hours and want to share my understanding of what I found and why I submitted the fix.
  
To address your question first: I had not used the CLI to generate the YAML before hitting this. My use case was writing YAML configs directly for a validation pipeline where I wanted the flexibility of hash: '*' without enumerating every column at config-write time, so the config would automatically pick up new columns as the schema evolved. I understand now that the CLI workflow already materializes wildcards into explicit column entries before writing the YAML file.
  
That said, here is the technical root cause so you have full context for your decision:
  
The YAML loading path (build_config_managers_from_yaml) builds a ConfigManager and appends it to the result list. The CLI path also builds a ConfigManager, but it additionally calls _get_calculated_config() which expands hash: '*' (or concat: '*') into per-column calculated_fields entries. The YAML path skips that call entirely. When a hand-authored YAML contains hash: '*', the wildcard is left as a literal string, calculated_fields remains empty, and downstream code that calls reduce() over that empty list crashes with an unhandled exception.

The fix adds a call to _get_calculated_config() inside the YAML path, gated on calculated_fields being empty. The guard ensures that CLI-generated YAMLs, which already contain materialized field entries, are not double-expanded.
  
I completely understand if the position is that hand-authored YAMLs with wildcards are outside the intended use case and this should not be fixed at the loader level. I just wanted to make sure the root cause is documented clearly in case other users hit the same crash.

sundar-mudupalli-work:
Thank you for your submission. Yes hand authored YAML is not an expected use case for this tool. 

Still your investigation may be valuable to understand our codebase - and you comments will remain documented in this issue. 

I am closing this comment as the status is Will Not Fix.

PR #1752, opened 2026-06-10
fix: YAML deserialization fix

nj1973:
## Description of changes
Update YAML deserialization to `safe_read`.

## Checklist

### DIFF
--- a/data_validation/cli_tools.py
+++ b/data_validation/cli_tools.py
@@ -54,14 +54,6 @@ import os
 import math
 from typing import Dict, List, Optional, TYPE_CHECKING
 import yaml
-from yaml import Dumper, dump, safe_load
-
-
-def _construct_yaml_tuple(loader, node):
-    return tuple(loader.construct_sequence(node))
-
-
-yaml.SafeLoader.add_constructor("tag:yaml.org,2002:python/tuple", _construct_yaml_tuple)
 
 from data_validation import (
     clients,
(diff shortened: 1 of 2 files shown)


sundar-mudupalli-work (changes requested):
Neil,

I now remember the problem I had. It turns out that SafeLoader cannot handle tuples and  python objects - basically anything that needs python tag including random code. I was trying to consume some tuples and SafeLoader will not work, but Loader does.

Gemini says there are ways to "teach" the safeloader to handle tuples, I have not tested them.

We use tuples in a few places, specifically with labels - which should make it into the yaml file. Can you review if our tests have labels and see if this change works with that use case ?

Thank you.

Sundar Mudupalli

nj1973:
> I now remember the problem I had. It turns out that SafeLoader cannot handle tuples and python objects - basically anything that needs python tag including random code. I was trying to consume some tuples and SafeLoader will not work, but Loader does.
> 
> We use tuples in a few places, specifically with labels - which should make it into the yaml file. Can you review if our tests have labels and see if this change works with that use case ?

You are correct, great info, I had no idea.

Example:

```
data-validation validate column -sc=ora -tc=pg -tbls=a.b --count='*' --format=csv --config-file=/tmp/c.yaml --labels=name=test,bill=ted
```

```
...
  labels:
  - !!python/tuple
    - name
    - test
  - !!python/tuple
    - bill
    - ted
...
```

sundar-mudupalli-work (approved):
Thank you for the update and the unit test.

We don't need to support python objects in general - don't think we have any of those.

Sundar Mudupalli

PR #1756, opened 2026-06-15
fix: Workaround for ibm_db_sa ForeignKeyConstraint exception

### LINKED ISSUE #1753: Db2 z/OS exception ForeignKeyConstraint
For all of schema, column and row validations. Example below:

```
$ data-validation validate schema -sc src -tc tgt -tbls OWNR1.TEST_TABLE
...
  File "/.../python3.11/dist-packages/sqlalchemy/engine/reflection.py", line 1799, in _reflect_fk
    sa_schema.ForeignKeyConstraint(
  File "/.../python3.11/dist-packages/sqlalchemy/sql/schema.py", line 4715, in __init__
    raise exc.ArgumentError(
sqlalchemy.exc.ArgumentError: ForeignKeyConstraint with duplicate source column references are not supported.
```

Tested with latest versions of `ibm_db` and `ibm_db_sa` packages.

nj1973:
## Description of changes
This PR disables foreign key reflection for Db2 backends. DVT does not use foreign key information so this should not have any downsides.

Ideally we would have figured out the root cause for an upstream fix but we could not identify the problem SQL and have therefore gone with this approach instead.

Strictly speaking we don't need this workaround for Db2 LUW but having it there is better for our automated testing.

## Issues to be closed

Closes #1753

## Checklist


renzokuken (approved):
LGTM

PR #1757, opened 2026-06-15
fix: Handle Snowflake table locations for Ibis 7

hashkanna:
## Description of changes
Fix Snowflake table and schema lookup on the Ibis 7.1 modernization branch.

This change splits Snowflake `DATABASE.SCHEMA` table locations before calling Ibis so that the database and schema are passed as separate arguments. It also updates the Snowflake datatype patch for Ibis 7.1 by patching `SnowflakeType.to_ibis` directly instead of using the removed `dt.dtype.register` API.

The Snowflake metadata path now uses `SnowflakeType.from_string(..., nullable=...)`, and unit coverage was added for Snowflake table location handling and NUMBER/BINARY datatype conversion.

## Issues to be closed
N/A

## Testing
- `venv/bin/python -m pytest tests/unit/test_clients.py tests/unit/ibis_snowflake/test_datatypes.py -q`
- `venv/bin/python -m black --check data_validation/clients.py tests/unit/test_clients.py third_party/ibis/ibis_snowflake/datatypes.py tests/unit/ibis_snowflake/test_datatypes.py`
- `git diff --check origin/ibis-7.1-modernization..HEAD`
- `PROJECT_ID=tpu-research-cloud-490704 venv/bin/python -m pytest --collect-only tests/system/data_sources/test_snowflake.py -q`
- `venv/bin/python -m pytest tests/unit -q -k 'not string_to_epoch and not execute_epoch_seconds_new'`

Full `venv/bin/python -m pytest tests/unit -q` still has the existing epoch conversion failures that reproduce on `origin/ibis-7.1-modernization`; those failures are unrelated to this Snowflake change.

## Checklist

### DIFF
--- a/third_party/ibis/ibis_snowflake/datatypes.py
+++ b/third_party/ibis/ibis_snowflake/datatypes.py
@@ -14,26 +14,31 @@
 from typing import Iterable, Tuple
 
 import ibis.expr.datatypes as dt
-import sqlalchemy as sa
 from ibis.backends.snowflake import Backend as SnowflakeBackend
-from ibis.backends.snowflake.datatypes import parse
+from ibis.backends.snowflake.datatypes import SnowflakeType
 from snowflake.connector.constants import FIELD_ID_TO_NAME
 from snowflake.sqlalchemy import NUMBER, BINARY
-from snowflake.sqlalchemy.snowdialect import SnowflakeDialect
+
+orig_snowflake_to_ibis = SnowflakeType.to_ibis
 
 
-@dt.dtype.register(SnowflakeDialect, NUMBER)
-def sa_sf_numeric(_, satype, nullable=True):
-    return dt.Decimal(
-        precision=satype.precision or 38,
-        scale=satype.scale or 0,
+@classmethod
+def dvt_snowflake_to_ibis(cls, typ, nullable=True):
+    if isinstance(typ, NUMBER):
+        return dt.Decimal(
+            precision=typ.precision or 38,
+            scale=typ.scale or 0,
+            nullable=nullable,
+        )
+    elif isinstance(typ, BINARY):
+        return dt.Binary(nullable=nullable)
+    return orig_snowflake_to_ibis(
+        typ,
         nullable=nullable,
     )
 
 
-@dt.dtype.register(SnowflakeDialect, BINARY)
-def sa_sf_binary(_, satype, nullable=True):
-    return dt.Binary(nullable=nullable)
+SnowflakeType.to_ibis = dvt_snowflake_to_ibis
 
 
 def _metadata(self, query: str) -> Iterable[Tuple[str, dt.DataType]]:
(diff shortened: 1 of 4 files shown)


third_party/ibis/ibis_snowflake/datatypes.py
```diff
@@ -14,26 +14,31 @@
...
-from snowflake.sqlalchemy.snowdialect import SnowflakeDialect
 
+orig_snowflake_to_ibis = SnowflakeType.to_ibis
 
-@dt.dtype.register(SnowflakeDialect, NUMBER)
-def sa_sf_numeric(_, satype, nullable=True):
-    return dt.Decimal(
-        precision=satype.precision or 38,
-        scale=satype.scale or 0,
+
```
nj1973, line 24: I put a breakpoint in this file and ran integration tests and manual DVT commands. The breakpoint was **not** fired.

So it seems this file is no longer in DVTs code path and any changes you made to it are redundant. This also means the new `test_datatypes.py` file is testing unused code.

I'm sure these changes were introduced for a reason though so I guess we need to revisit the original problem.

nj1973: I ran this test:

```
data-validation generate-table-partitions  -sc snowflake -tc ora_local -tbls=PSO_DATA_VALIDATOR.PUBLIC.DVT_CORE_TYPES=pso_data_validator.dvt_core_types  --comparison-fields="*"  --config-dir /tmp/dtpart  --partition-num 2
...
AttributeError: 'Backend' object has no attribute 'list_primary_key_columns'
```

Which is caused by this file not being used so we do not want to remove this file, we need to get it back into the DVT code path.

nj1973 (changes requested):

third_party/ibis/ibis_snowflake/datatypes.py
sundar-mudupalli-work: Neil,

Thank you. This problem has been resolved in the latest commit to this branch by JetSki.

I believe I have tracked down the source of the issue and it is [here](https://github.com/hashkanna/professional-services-data-validator/blob/cda4b5eabd4c3ced4416a89d744fc66c46c5b5fb/data_validation/clients.py#L90). You will notice that the call is directly to `ibis.snowflake.connect`, bypassing our `datatypes.py`. The latest version and develop ([here](https://github.com/hashkanna/professional-services-data-validator/blob/326770c4219b40a6a68d474ade11f756ef1f6694/data_validation/clients.py#L86)) have a reference to `snowflake_connect` in `third_party.ibis.ibis_snowflake.api`.

Thank you for your attention to this level of detail.

Tests for all backends pass in the latest commit - you can see the build results [here](https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/8e68c2d0-011f-43b7-bd2f-c331a01e3847?project=pso-kokoro-resources)

Hope that clarifies.

Sundar Mudupalli

PR #1761, opened 2026-06-16
fix: Use UTC for naive timestamp epoch conversion

hashkanna:
## Summary
- Treat timezone-naive parsed timestamp strings as UTC before epoch conversion.
- Add explicit timezone-aware string cases, including a pre-1970 UTC timestamp.
- Move the `data_validation.clients` import inside `ibis_table_to_sql()` to avoid import-cycle failures.

## Rationale
`datetime.astimezone()` interprets naive datetimes as local machine time, which makes epoch conversion depend on the host timezone. Existing tests already expect naive timestamp strings to use UTC epoch semantics.

The utility import change keeps `data_validation.util` lightweight at import time and avoids recursively importing `data_validation.clients` through backend modules before initialization is complete.

## Tests
- `venv-develop/bin/python -m pytest tests/unit/ibis_addon/test_operations.py -q`
- `venv-develop/bin/python -m pytest tests/unit -q`
- `venv-develop/bin/python -m black --check tests/unit/ibis_addon/test_operations.py`
- `git diff --check`

### DIFF
--- a/data_validation/util.py
+++ b/data_validation/util.py
@@ -18,7 +18,7 @@ import re
 import time
 import uuid
 
-from data_validation import clients, exceptions
+from data_validation import exceptions
 
 from typing import TYPE_CHECKING
 
@@ -76,6 +76,10 @@ def ibis_table_to_sql(ibis_table: "IbisTable", alchemy_client: "BaseBackend") ->
 
     We need the client in order to find the dialect, otherwise we end up with generic literals.
     """
+    # Import clients here because it is a heavy weight import for a utility module
+    # and we want to avoid circular imports since clients also (eventually) imports utils.
+    from data_validation import clients
+
     # If the backend uses sqlalchemy, we will need to request sqla to bind variables
     # for a non sqlalchemy backend, the parameters are already bound
     if alchemy_client and clients.is_sqlalchemy_backend(alchemy_client):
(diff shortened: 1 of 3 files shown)


tests/unit/ibis_addon/test_operations.py
```diff
@@ -79,6 +79,18 @@ def test_format_raw_sql_expr(module_under_test):
             "1969-12-31 23:59:00",
             -60,
         ),
+        (
```
nj1973, line 86: For completeness should we add a tz aware datetime prior to 1970?

e.g.:

```
        (
            "1969-12-31 23:59:00Z",
            -60,
        ),
```

data_validation/util.py
```diff
@@ -76,6 +76,10 @@ def ibis_table_to_sql(ibis_table: "IbisTable", alchemy_client: "BaseBackend") ->
 
     We need the client in order to find the dialect, otherwise we end up with generic literals.
     """
+    # Import clients here because it is a heavy weight import for a utility module
+    # and we want to avoid circular imports since clients also (eventually) imports utils.
+    from data_validation import clients
```
nj1973, line 81: When I tried to test locally I was getting a recursive import error. I appreciate this is nothing to do with your change but I think we need to protect ourselves from it. It must be differences in the import sequence for different Python versions.

tests/unit/ibis_addon/test_operations.py
hashkanna: Done

sundar-mudupalli-work (approved):
LGTM

PR #1767, opened 2026-06-17
test: Add Ibis 7 compiler matrix coverage

hashkanna:
## Description of changes
Add regression coverage for DVT's local Ibis addon layer under Ibis 7.1.

DVT uses custom Ibis operations for validation-specific expressions such as row hashing, binary length, decimal-to-string formatting, and timestamp epoch conversion. These operations are registered with backend-specific SQL compilers, so the same DVT expression compiles differently for BigQuery, PostgreSQL, MySQL, SQL Server, Impala, Redshift, and Oracle.

The new compiler matrix verifies those custom operations still compile to the expected backend SQL fragments after the Ibis 7.1 migration. Oracle cases run when the Oracle compiler dependency is available.

This also fixes the Impala `BinaryLength` registration to compile as `length(<expr>)` instead of using the SQLAlchemy generic function path that produced a bound function name.

## Issues to be closed
N/A

## Testing
- `venv/bin/python -m pytest tests/unit/ibis_addon/test_compiler_matrix.py -q`
- `TZ=UTC venv/bin/python -m pytest tests/unit/ibis_addon -q`
- `TZ=UTC venv/bin/python -m pytest tests/unit -q`
- `venv/bin/python -m black --check tests/unit/ibis_addon/test_compiler_matrix.py third_party/ibis/ibis_addon/operations.py`
- `git diff --check upstream/ibis-7.1-modernization...HEAD`

## Checklist

### DIFF
--- a/tests/unit/ibis_addon/test_compiler_matrix.py
+++ b/tests/unit/ibis_addon/test_compiler_matrix.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import importlib.metadata
+
 import ibis
 import pytest
 from ibis.backends.bigquery.compiler import BigQueryCompiler
@@ -22,7 +24,24 @@ from ibis.backends.postgres.compiler import PostgreSQLCompiler
 
 # Import required in order to register DVT operations.
 import third_party.ibis.ibis_addon.operations  # noqa: F401
+from third_party.ibis.ibis_cloud_spanner.compiler import SpannerCompiler
 from third_party.ibis.ibis_redshift.compiler import RedshiftCompiler
+
+try:
+    from ibis.backends.snowflake import SnowflakeCompiler
+except (ImportError, importlib.metadata.PackageNotFoundError) as exc:
+    SnowflakeCompiler = None
+    SNOWFLAKE_SKIP_REASON = f"Snowflake compiler unavailable: {exc}"
+else:
+    SNOWFLAKE_SKIP_REASON = ""
+
+try:
+    from third_party.ibis.ibis_teradata.compiler import TeradataCompiler
+except ImportError as exc:
+    TeradataCompiler = None
+    TERADATA_SKIP_REASON = f"Teradata compiler unavailable: {exc}"
+else:
+    TERADATA_SKIP_REASON = ""
 
 try:
     from third_party.ibis.ibis_oracle.compiler import OracleCompiler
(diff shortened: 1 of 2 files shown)


tests/unit/ibis_addon/test_compiler_matrix.py
```diff
@@ -0,0 +1,179 @@
...
+    },
+    name="t",
+)
+
+
+def _compile(compiler_class, expr):
+    return str(compiler_class().to_sql(expr))
+
+
+COMPILER_CASES = [
```
nj1973: I find this a little confusing. We don't have all engines in here incomplete coverage of expressions.

For example:
- we have some PostgreSQL tests but not binary_length, which we have for some other engines
- we don't have any Spanner tests

Maybe it would be better to have a set of binary length parameters that covers all engines and a specific test. And then have a second dedicated test for hash bytes for all engines.

Then it is a true matrix and not partial in both axis.

nj1973 (changes requested):
I love new tests so am on board with this new file. But I think it could be clearer/more complete.

tests/unit/ibis_addon/test_compiler_matrix.py
hashkanna: sure, thanks for the feedback! Let me have a look.

tests/unit/ibis_addon/test_compiler_matrix.py
```diff
@@ -0,0 +1,179 @@
...
+        PostgreSQLCompiler,
+        lambda: TABLE.s.hashbytes().name("h"),
+        ["encode(sha256(convert_to(t0.s, 'UTF8')), 'hex')"],
+        [],
+        id="postgres-hashbytes",
+    ),
+    pytest.param(
+        PostgreSQLCompiler,
+        lambda: TABLE.d.cast("string").name("dstr"),
+        ["rtrim(to_char(t0.d"],
```
nj1973: Also, this expression looks incorrect. It appears to be missing at least two trailing parentheses.

tests/unit/ibis_addon/test_compiler_matrix.py
hashkanna: I have created dedicated `hashbytes` and `binary_length` compiler matrices, and added the missing PostgreSQL binary-length and Spanner coverage.

tests/unit/ibis_addon/test_compiler_matrix.py
hashkanna: fixed it now. Postgres decimal cast assertion now checks the complete generated expression:
`rtrim(to_char(t0.d, :to_char_1), :rtrim_1)`

nj1973 (approved):
LGTM

PR #1768, opened 2026-06-17
fix: Handle Spanner schema lookup for Ibis 7

hashkanna:
## Summary

- route Cloud Spanner table and schema lookup through Spanner backend methods without passing DVT `schema_name`
- avoid treating DVT schema values as Spanner database names; Spanner should use the database configured on the connection
- restore the Spanner backend `re` import and remove an unused schema extraction in the execute path
- add unit coverage for Spanner schema handling and timestamp precision equivalence parsing

## Tests

- `venv/bin/python -m pytest tests/unit/test_clients.py tests/unit/test_schema_validation.py -q`
- `TZ=UTC venv/bin/python -m pytest tests/unit -q`

### DIFF
--- a/data_validation/clients.py
+++ b/data_validation/clients.py
@@ -254,6 +254,8 @@ def get_ibis_table(client, schema_name, table_name, database_name=None):
             schema_name, database_name
         )
         return client.table(table_name, database=database_name, schema=schema_name)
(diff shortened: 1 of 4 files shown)


nj1973 (approved):
LGTM

PR #1779, opened 2026-07-15
fix: SQL Server padded char columns are detected correctly

### LINKED ISSUE #1778: Padded char primary keys not matching for SQL Server/PostgreSQL validation
**Describe the bug**
Padded char primary keys not matching for SQL Server/PostgreSQL validation

8.8.0

SQL Server vs PostgreSQL

Add standard fixed char tests to test suite.

- test_fixed_char_pk_row_validation_to_bigquery
- test_fixed_char_pk_query_row_validation_to_bigquery

**Expected behavior**
PK columns should align

nj1973:
## Description of changes
This PR fixed fixed padded char detection for SQL Server and adds missing tests.

Also temporarily disables Snowflake testing due to account issues.

## Issues to be closed

Closes #1778

## Checklist

### DIFF
--- a/tests/system/data_sources/test_sql_server.py
+++ b/tests/system/data_sources/test_sql_server.py
@@ -820,7 +820,7 @@ def test_row_validation_datetime_pk_to_bigquery():
 )
 def test_fixed_char_pk_row_validation_to_bigquery():
     """Test fixed char primary keys"""
-    id_column_row_validation_test("pso_data_validator.dvt_fixed_char_id")
+    id_column_row_validation_test("pso_data_validator.dvt_fixed_char_id", hash="id")
 
 
 @mock.patch(
@@ -1049,7 +1049,9 @@ def test_custom_query_row_hash_validation_core_types_to_bigquery():
 )
 def test_fixed_char_pk_query_row_validation_to_bigquery():
     """Test fixed char primary keys on custom query"""
-    id_column_query_row_validation_test("pso_data_validator.dvt_fixed_char_id")
+    id_column_query_row_validation_test(
+        "pso_data_validator.dvt_fixed_char_id", hash="id"
+    )
 
 
 @mock.patch(
--- a/tests/system/data_sources/test_sql_server.py
+++ b/tests/system/data_sources/test_sql_server.py
@@ -820,7 +820,9 @@ def test_row_validation_datetime_pk_to_bigquery():
 )
 def test_fixed_char_pk_row_validation_to_bigquery():
     """Test fixed char primary keys"""
-    id_column_row_validation_test("pso_data_validator.dvt_fixed_char_id", hash="id")
+    id_column_row_validation_test(
+        "pso_data_validator.dvt_fixed_char_id", hash="id", use_random_row=True
+    )
 
 
 @mock.patch(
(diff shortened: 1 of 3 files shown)


tests/system/data_sources/test_sql_server.py
```diff
@@ -52,6 +53,7 @@
 SQL_SERVER_USER = os.getenv("SQL_SERVER_USER", "sqlserver")
 SQL_SERVER_PASSWORD = os.getenv("SQL_SERVER_PASSWORD")
 SQL_SERVER_DATABASE = os.getenv("SQL_SERVER_DATABASE", "guestbook")
+SQL_SERVER_CONFIG_JSON = os.getenv("SQL_SERVER_CONFIG_JSON")
```
sundar-mudupalli-work, line 56: What is SQL_SERVER_CONFIG_JSON ? It is not set in `noxfile.py` where it should be.

This is a new feature? It should be documented.

sundar-mudupalli-work:
Hi Neil,

I tried this branch - there is likely something wrong with the test. When I tried it in the "wild" the test did not pass for me. Specifically, I tried this command `data-validation validate row -sc bq -tc sqlserver -hash=id,other_data -pk=id -tbls=pso_data_validator.dvt_fixed_char_id`. I got 3 failures and two successes. Looking at the data in BQ, I see:
```
data-validation query -c bq -q "SELECT id,other_data from pso_data_validator.dvt_fixed_char_id"
[('DVT2', 'Row 2  \t'), ('DVT4', 'Row 4  \t  '), ('DVT3', 'Row 3  '), ('DVT1', 'Row 1\t  '), ('DVT5', 'Row 5')]
```
Basically the BQ data is never trimmed and we only trim spaces, not tabs, ids 'DVT4', 'DVT1' and 'DVT3' should always fail. I don't know why this table is set up this way. I probably set it up, not sure what the thinking was.

When I look at your test, it expects 5 success results - it is possible that our random row function does really poorly and picks the successful rows only - not sure. I don't know why this was set up as a random row test.

Thanks.

Sundar Mudupalli

tests/system/data_sources/test_sql_server.py
nj1973: This is in the tests only, it is not a new feature. The hardcoded default for ODBC driver works for our kokoro testing but I have v18 of the ODBC driver installed in my project which means I can't run tests. With this env var I can pass in my specific override, just like for SQL_SERVER_PASSWORD, etc.

nj1973:
My test passes because it only hashes the `id` column, not `other_data`. This is because of a SQL Server limitation we documented in the past: https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/docs/limitations.md#sql-server

"SQL Server does not have a function to "right trim" all whitespace, only spaces, therefore any validations relying on removal of trailing white space may encounter issues."

For this specific test we are testing padded chars so I believe it is acceptable to ignore the `other_data` column. The tabs in the data are confusing, when I first added the test it failed and I was confused, until I looked into it and then saw the limitation.

tests/system/data_sources/test_sql_server.py
sundar-mudupalli-work: LGTM - we are consistent with sqlalchemy, however pyodbc uses a simpler approach. We may have to revisit the "query" parameter when we let go of sqlalchemy and use the underlying connection library.

Sundar Mudupalli

sundar-mudupalli-work (approved):
Thank you.

PR #1786, opened 2026-07-17
chore: Table name indicator in DataValidation exceptions

### LINKED ISSUE #1776: Exceptions should indicate the table causing them
A customer has many tables to validate and plan to redirect all stderr to an error log file so the operator only has to review a single file the following morning, rather than open 100s of error logs to review issues.

When DVT throws an exception then there is no easy way to know which table failed. We only have the error stack. This makes is laborious to know which command to revisit.

It would be nice if the exception could include an indicator to either the table or command that threw the exception.

nj1973:
## Description of changes
Adds a small usability change to include the source table name (where possible) in `DataValidation()` exceptions.

## Issues to be closed

Closes #1776

## Checklist

### DIFF
--- a/data_validation/__main__.py
+++ b/data_validation/__main__.py
@@ -529,33 +529,40 @@ def run_validation(config_manager: ConfigManager, dry_run=False, verbose=False):
         logging.warning(
             "Trim String Primary Keys has been deprecated, validation results may vary"
         )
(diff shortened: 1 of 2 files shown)


sundar-mudupalli-work (approved):
LGTM

Thank you.

Sundar Mudupalli

PR #1806, opened 2026-08-03
fix: Fix SQL Server decimal format bug

### LINKED ISSUE #1805: SQL Server row validation bug formatting decimals
**Describe the bug**
SQL Server row validation has a bug formatting decimals with a zero directly after the decimal place, e.g. `9.01` is formatted as `91`.

8.9.0

SQL Server

nj1973:
## Description of changes
PR fixes a decimal formatting bug for SQL Server row validations.

Also adds integration tests across common SQL engines.

## Issues to be closed

Closes #1805

## Checklist


third_party/ibis/ibis_mssql/registry.py
```diff
@@ -144,8 +144,7 @@ def sa_cast_mssql(t, op):
         # Considering any number of fractional digits
         format_string = f'0.{("#" * scale)}'
         formatted_value = sa.func.format(sa_arg, format_string)
-        # Replace trailing '.0' with ''
-        return sa.func.replace(formatted_value, ".0", "")
+        return formatted_value
```
sundar-mudupalli-work, line 147: This change appears to be correct. This code was introduced in [PR 1499](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1499/changes#diff-b652099958b5dddbcbc1ad70a5ab1930a7b773aaca29d53daf814df06347775eR429). The developer felt the need to replace ".0" with "".  There is no need for this replacement. The use of f'0.###' guarantees that there will be no trailing 0's.

`sa.func.replace` is a very dangerous function, replacing any occurrence of the given string with the new string. We use this function in line 97 of this file as well. Looking at [PR 1579](https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1579), that change was to workaround a quirk in the SQL Server Length function which ignores trailing spaces. That change is OK as the result of the replacement is just used to calculate length and not for any other purpose.

There are a couple of other places in the code base where the use of this function is a more concerning:
1. [Line 305] (https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/d6d16dab92a798db8aaedd067bd95dbd5d9078ee/third_party/ibis/ibis_oracle/registry.py#L305) and 
2. [Line 311 of db2](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/d6d16dab92a798db8aaedd067bd95dbd5d9078ee/third_party/ibis/ibis_db2/registry.py#L311)
These two appear to be copies of each other. The Oracle code was introduced during the upgrade to ibis 5.1.0. The developer wanted to replace occurrences of '05', '06' in month number as ' 5' and ' 6'. It is not clear if string needs to have a space. If not, the developer could use the format specified 'FMDD' which will not produce '05' and '06', instead '5' and '6'. As the code is currently written, it would appear that dates in October (month 10) will appear as '1 ' and are likely to show as not matching. This needs to be verified and if true an issue needs to be opened.

Sundar Mudupalli

tests/resources/bigquery_test_tables.sql
sundar-mudupalli-work, line 62: To catch a regression of this bug, I recommend adding a row with the id of 4 to this table for all engines with the numeric 123.02. I think it is worth it.

Sundar Mudupalli

sundar-mudupalli-work (approved):
Neil,

Overall good. I am approving this, so that you can go ahead and merge it after making the change I suggested to prevent regressions.

Thank you.

Sundar Mudupalli

third_party/ibis/ibis_mssql/registry.py
nj1973: I've raised issue 1808 to review this code. In practice we don't use strftime `%e` so I don't think it's an urgent risk. But still should be reviewed/updated.

tests/resources/bigquery_test_tables.sql
nj1973: I've added the requested value to `dvt_decimals` instead of `dvt_core_types`. Also added `dvt_decimals` to any engines I hadn't done previously.

PR #1811, opened 2026-08-10
fix: Cast binary values to hex in combiner before reporting them to screen

### LINKED ISSUE #1810: UnicodeDecodeError during comparison fields validation of tables containing binary data
**Describe the bug**
If a binary column contains only valid UTF-8/ASCII bytes then comparison fields validation succeeds.

If the binary column contains any arbitrary binary bytes (non-UTF-8) then the validation crashes with UnicodeDecodeError.

During the validation run, comparison columns are fetched as raw bytes into a Pandas DataFrame. When generating the report in the combiner stage, DVT executes an Ibis Pandas query that casts these aggregate values to "string" so they can be merged into a single output column (source_agg_value/target_agg_value).

The standard Pandas backend executes this cast via `.astype(str)`, which defaults to a UTF-8 decoding operation. If the binary column contains arbitrary non-UTF-8 bytes (such as the standard Java serialization magic header 0xac), this fails with a decode exception. e.g.:

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xac in position 0: invalid start byte
```

8.9.1

SQL Server (but applies to all engines).

nj1973:
## Description of changes
This PR changes the DVT combiner to convert binary columns to hex before reporting them.

This is because and binary column containing non-UTF8 characters will trigger an exception:

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xac in position 0: invalid start byte
```

This only applies to comparison fields validations, concat/hash already cast to hex in the source/target SQL statements.

The down side to this is that any binary columns containing all UTF8 tokens will be output as hex, even though they "could" be output as a pure binary-to-string cast. But it brings comparison fields into line with other validations and prevent exceptions. I think the pros outweigh the cons.

## Issues to be closed

Closes #1810

## Checklist

### DIFF
--- a/tests/unit/test_combiner.py
+++ b/tests/unit/test_combiner.py
@@ -1264,3 +1264,51 @@ def test_convert_large_ints_to_decimals(module_under_test):
 
     assert isinstance(res["normal_int"][0], (int, numpy.integer))
     assert isinstance(res["string_col"][0], str)
+
+
+def test_generate_report_with_binary_data(module_under_test):
+    """Test report generation with binary data containing non-UTF-8 characters.
+
+    This ensures that DVT does not crash with a UnicodeDecodeError when
+    casting binary columns to string in the Pandas combiner phase. Instead,
+    the binary data should be formatted as a hex string in the report.
+    """
+    # Binary data with non-UTF-8 bytes (e.g. 0xac)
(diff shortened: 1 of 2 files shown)


renzokuken (approved):
LGTM

PR #1820, opened 2026-08-22
fix: clean up Teradata connection parameters

### LINKED ISSUE #1016: Teradata Connection Parameters may need cleanup and/or better documentation.
Hi,

DVT uses [teradatasql](https://pypi.org/project/teradatasql/). The way we connect to Teradata may need to be cleaned up and/or documented better.

1. The [current documentation](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/docs/connections.md#teradata) says that port and logmech default to 1025 and TD2 respectively. The teradatasql library uses the same default values. However, the [backend](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/develop/third_party/ibis/ibis_teradata/__init__.py) forces these two values to 1025 and TD2 rather than letting teradatasql set the defaults. Don't know why this is - look like this can be removed and from ``__init__.py``. 
2. The documentation that if any of the `host`, `port`, `user-name`, `password` and `logmech` are mentioned in `json-params`, the values in `json-params` will take priority. I am not sure if this is the right behavior - that is what the code will do as per my review.

Thanks.

Sundar Mudupalli

miachillgood:
## Summary

- let the Teradata driver supply its own port and authentication defaults when those options are omitted
- apply explicit connection options after `json-params`, so command-line values take precedence
- document the precedence and driver parameter names, with regression tests for both behaviors

## Validation

- `python -m pytest tests/unit/ibis_teradata/test_init.py` (6 passed)
- `python -m black --check third_party/ibis/ibis_teradata/api.py third_party/ibis/ibis_teradata/__init__.py tests/unit/ibis_teradata/test_init.py`
- `python -m flake8 tests/unit/ibis_teradata/test_init.py`
- `git diff --check`

Fixes #1016

### DIFF
--- a/docs/connections.md
+++ b/docs/connections.md
@@ -180,13 +180,17 @@ data-validation connections add
     [--secret-manager-project-id SECRET_PROJECT_ID]     Secret Manager project ID
     --connection-name CONN_NAME Teradata                Connection name
     --host HOST                                         Teradata host
-    --port PORT                                         Teradata port, defaults to 1025
+    [--port PORT]                                       Teradata port; uses the driver default when omitted
     --user-name USER                                    Teradata user
     --password PASSWORD                                 Teradata password
-    [--logmech LOGMECH]                                 Teradata logmech, defaults to "TD2"
+    [--logmech LOGMECH]                                 Teradata logmech; uses the driver default when omitted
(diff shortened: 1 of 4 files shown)


nj1973 (approved):
These changes look good, thanks for contributing to DVT!

nj1973:
@miachillgood , I created release 8.9.3 today which contains your recent Teradata improvement.

Issue #1822, opened 2026-08-31
Random Row fails for DECIMAL primary keys with 31 digits


sundar-mudupalli-work:
**Describe the bug**
Random Row fails for DECIMAL primary keys with 30 digits. The error is likely because these are being cast into floats, and the no rows match due the loss of precision

**Steps to reproduce the behavior**
See that the table has proper data with a large decimal number as primary key.
```
data-validation query -c postgres -q 'select id, col_data from pso_data_validator.dvt_large_decimals'
[(Decimal('123456789012345678901234567890'), 'Row 1'), (Decimal('223456789012345678901234567890'), 'Row 2'), (Decimal('323456789012345678901234567890'), 'Row 3'), (Decimal('423456789012345678901234567890'), 'Row 4'), (Decimal('523456789012345678901234567890'), 'Row 5')]
```
Run validation on this data
```
data-validation validate row -tc postgres -sc postgres -tbls=pso_data_validator.dvt_large_decimals  -hash=col_data -rr -rbs=3
╒═══════════════════╤═══════════════════╤═════════════════════╤══════════════════════╤════════════════════╤════════════════════╤══════════════════╤═════════════════════╤══════════╕
│ validation_name   │ validation_type   │ source_table_name   │ source_column_name   │ source_agg_value   │ target_agg_value   │ pct_difference   │ validation_status   │ run_id   │
╞═══════════════════╪═══════════════════╪═════════════════════╪══════════════════════╪════════════════════╪════════════════════╪══════════════════╪═════════════════════╪══════════╡
╘═══════════════════╧═══════════════════╧═════════════════════╧══════════════════════╧════════════════════╧════════════════════╧══════════════════╧═════════════════════╧══════════╛
```

This should show 3 rows and is showing 0 rows, likely because of the loss of precision in converting decimal to float. Also see [random_row_validation.md](docs/random_row_validation.md) for an alternative approach to random row validation.

Sundar Mudupalli

nj1973:
@sundar-mudupalli-work I'm glad you raised this. Because tests were passing I assumed this worked but the assertion in `test_row_validation_large_decimals_to_bigquery` was flawed. Fixing it has thrown up a few issues.

I am working on this. Thanks.

PR #1823, opened 2026-09-02
feat: Update Combiner in DVT to use duckdb to connect to Pandas dataframe

### LINKED ISSUE #1702: Use ibis DuckDB backend to access pandas dataframe instead of ibis pandas backend
A feature request - DVT currently uses a pandas backend that has been removed in Ibis 10.0 (we are currently on Ibis 5.1). You can use DuckDB to access data stored in Pandas dataframes. There are some performance benefits and some feature changes. Ibis explains why in this [article](https://ibis-project.org/posts/farewell-pandas). In version 9.0, ibis introduced the concept of a default backend and they chose [DuckDB](https://ibis-project.org/posts/why-duckdb/).
8.3

### LINKED ISSUE #1760: Lack of support for BIGNUMERIC when using duckdb backend.
**Describe the bug**
The Duckdb backend does not support Google BigNumeric. Any data with more than 38 digits of precision will fail with the Duckdb backend.

Probably DVT version 9 and later.

BigQuery for Source and Target.

See test `test_numeric_types` in [test_bigquery.py](tests/system/data_sources/test_bigquery.py)

**Expected behavior**

**Screenshots**

Ex: CLI commands on local computer or VM, within a DAG, Cloud Run job, etc

**Additional context**

### LINKED ISSUE #1763: Oracle Test fails after using duckdb backend to process pandas dataframe
**Describe the bug**
Tests `test_column_validation_all_null_oracle_to_postgres` and `test_row_validation_comp_fields_core_types` fail due to overflow of DECIMAL(18,3). There appear to be more than 18 digits in the value - which means the overflow is legitimate.
```
FAILED tests/system/data_sources/test_oracle.py::test_column_validation_all_null_oracle_to_postgres - ibis.common.exceptions.IbisTypeError: Arguments differences_target_value:float64 and differences_source_value:string are not...
FAILED tests/system/data_sources/test_oracle.py::test_row_validation_comp_fields_core_types - sqlalchemy.exc.ProgrammingError: (duckdb.duckdb.InvalidInputException) Invalid Input Error: Attempting to execute an unsucce...
```
Latest version based on Ibis 7.1

Oracle

Run the tests - after removing the statement to skip this test.

**Expected behavior**
Complete successfully.

**Screenshots**

Ex: CLI commands on local computer or VM, within a DAG, Cloud Run job, etc

**Additional context**

### LINKED ISSUE #1764: SQL Server Test fail after migration to Duckdb Backend for Pandas dataframes
**Describe the bug**
Tests `test_column_validation_ss_types_to_bigquery` and `test_row_validation_comp_fields_ss_types_to_bigquery` fail due to overflow of DECIMAL(18,3). There appear to be more than 18 digits in the value - which means the overflow is legitimate. Also see issue 1763
```
FAILED tests/system/data_sources/test_sql_server.py::test_column_validation_ss_types_to_bigquery - sqlalchemy.exc.ProgrammingError: (duckdb.duckdb.InvalidInputException) Invalid Input Error: Attempting to execute an unsucce...
FAILED tests/system/data_sources/test_sql_server.py::test_row_validation_comp_fields_ss_types_to_bigquery - sqlalchemy.exc.ProgrammingError: (duckdb.duckdb.InvalidInputException) Invalid Input Error: Attempting to execute an unsucce...
```

Ex: 7.1.0

SQL Server and BigQuery

Ex: 1. Add '...'
2. Execute '....'

**Expected behavior**

**Screenshots**

Ex: CLI commands on local computer or VM, within a DAG, Cloud Run job, etc

**Additional context**

sundar-mudupalli-work:
## Description of changes
Ibis has deprecated the pandas backend and has removed it in 10.0. The way to work with pandas dataframes is with the DuckDB backend. The reasoning for this is explained in this [article](https://ibis-project.org/posts/farewell-pandas/).

With this change we can work on upgrading DVT from ibis 7.1 to something newer, getting away for SQL Alchemy.

## Issues to be closed
* Closes #1702
* Closes #1763 
* Closes #1764 
* Closes #1760 (DVT converts BigNumeric into float - which creates other issues).

## Checklist


data_validation/combiner.py
```diff
@@ -278,18 +293,29 @@ def _calculate_difference(
...
     ):
         # String data types i.e "None" can be returned for NULL timestamp/datetime aggs
-        if is_value_comparison:
-            difference = pct_difference = ibis.null()
-        else:
-            difference = pct_difference = ibis.null().cast("float64")
+        difference = pct_difference = ibis.null().cast("float64")
+        is_source_null = (
+            source_value.isnull()
+            | (source_value.cast("string") == "<NA>")
```
nj1973: This section looks odd to me. I think this will give us status="success" if a string column contains a string "None" in one system and a real null/None in the other?

If you haven't already I think you should check this is not the case.

data_validation/combiner.py
```diff
@@ -88,6 +88,9 @@ def generate_report(
     is_value_comparison=False,
     verbose=False,
 ) -> "DataFrame":
+    import sys
+
+    sys.setrecursionlimit(max(sys.getrecursionlimit(), 20000))
```
nj1973: Increasing the recursionlimit while also halving the columns we can process (`COMBINER_COLUMN_SLICE_WIDTH`) is a red flag that something else is going wrong. It doesn't seem right that an Ibis query with DuckDB should be so less efficient than with Pandas.

I asked Gemini about risks of changing the recursion limit and it came up with the list below:

• Why it is broken:
1. sys.setrecursionlimit() mutates the global CPython interpreter state. When DVT is deployed as a multi-threaded service (e.g., via data-validation deploy Flask service or multi-threaded worker pools), worker threads are allocated fixed OS C-stack limits (typically 512KB to 2MB).
2. Setting the Python recursion limit to 20,000 prevents Python from safely raising a RecursionError. Instead, deep AST traversals exhaust the thread's native OS C-stack, resulting in an uncatchable Hard Segmentation Fault (SIGSEGV) that terminates the entire host process.
3. This is an anti-pattern used to mask the unoptimized, exponential AST tree generated by DuckDB expressions in combiner.py:519.

In the supporting "design docs" I don't see information about why these were added, only that they were tuned. I think we need to take a step back and understand whether these changes are a workaround for a problem in our code or genuinely necessary because Ibis/DuckDB is less efficient than Ibis/Pandas.

data_validation/combiner.py
```diff
@@ -88,6 +88,9 @@ def generate_report(
...
     """Combine results into a report.
```
nj1973, line 49: I had a conversation with Gemini try and understand why whether these issues were down to how we are using DuckDB or fundamental Ibis/DuckDB issues. This is the outcome of the conversation:

# Gemini analysis
DuckDB itself is an exceptional analytical SQL engine, but forcing it into DVT's existing combiner.py under Ibis 7.1.0 exposes the worst aspects of both systems.

Here is the detailed breakdown of what is an implementation flaw versus what is a genuine DuckDB limitation for DVT, and how this should actually be solved.
──────
### 1. What is down to "How the Change Was Made"

#### A. The 157x Slowdown & Recursion Crashes Are Implementation Artifacts

In the original Pandas client, combiner.py:191 evaluated each pivot query into an intermediate DataFrame, then joined those three flat DataFrames.

In PR #1823, the author attempted to "streamline" the logic by piping table expressions directly into joins without intermediate execution. In Ibis, each column validation in `combiner.py:368` and `combiner.py:438` builds a subquery that gets chained via `functools.reduce(lambda a, b: a.union(b))`.

By joining `combiner.py:227`, `combiner.py:230`, and `combiner.py:224` as unmaterialized expressions, the resulting SQL query becomes a 3-way outer join of three 60-level-deep nested UNION ALL
subqueries:

• Generating and compiling this giant AST exceeds Python's default call stack of 1,000 (hence the hacky `sys.setrecursionlimit(20000))`.
• DuckDB's parser and optimizer must parse and optimize tens of kilobytes of nested subqueries on every single slice.
• Execution time explodes from 0.35s in Pandas to 54.94s in DuckDB.

#### B. Why the Author Did Not Materialize Intermediate Tables

Why didn't the author just materialize the pivots as the Pandas implementation did?
Because under Ibis 7.1.0, attempting to pass the pivoted DataFrame back to DuckDB via `con.create_table("diff", diff_df)` crashes immediately:

```
ibis.common.exceptions.IbisTypeError: DuckDB cannot yet reliably handle `null` typed columns; got null typed columns: ['primary_keys', 'num_random_rows']
```

Ibis 7.1.0's DuckDB backend is an obsolete SQLAlchemy-based dialect (BaseAlchemyBackend). Modern Ibis (9.0+) completely rewrote the DuckDB backend using direct relational APIs and SQLGlot
without SQLAlchemy. Adopting DuckDB within Ibis 7.1.0 forces DVT to deal with bugs and limitations that modern Ibis has long discarded.

nj1973: Gemini then went on to offer a totally different approach which sounds like a much better approach to me:

### 3. The Core Architectural Question: Why Use an SQL Database to Combine DataFrames?

The reason Ibis is deprecating ibis.pandas is because maintaining an execution engine that translates relational algebra into Pandas imperative methods (apply, groupby, merge) was a maintenance nightmare for the Ibis team. For Ibis users running queries against Parquet/CSV files or DataFrames, DuckDB is vastly superior.

However, DVT's combiner is not an arbitrary query engine.

Let's look at what combiner.py actually does:

  1. It receives source_df and target_df (which are already in local Python memory as Pandas DataFrames).
  2. It pivots/melts the aggregate columns into rows (validation_name, source_agg_value, target_agg_value).
  3. It computes differences, percentage differences, and assigns success or fail.
  4. It performs a full outer join on (validation_name, *join_on_fields).

Using Ibis to accomplish this was originally chosen simply because "DVT uses Ibis everywhere." But routing two in-memory Pandas DataFrames into an in-memory SQL database (DuckDB) through an SQL compiler (Ibis/SQLAlchemy) just to unpivot and join them is the root cause of the recursion issues, the slice thresholds, and the performance penalties:

```
    In-memory DataFrame
      -> Ingest into DuckDB Tables
      -> Construct 60-union AST in Ibis
      -> Compile to 60KB SQL via SQLAlchemy
      -> Optimize and execute in DuckDB
      -> Export back to in-memory DataFrame
```

### Recommendation

1. Do not use DuckDB for the Combiner:
Replacing ibis.pandas with ibis.duckdb in combiner.py creates massive complexity, slows down execution by two orders of magnitude, and corrupts >38 digit decimals.
2. Rewrite combiner.py Using Native Pandas / Python:
      • Implementing the pivoting, difference calculation, and outer join directly in Pandas (using pandas.melt and pandas.merge(how='outer')) takes roughly 150 lines of clean Python.
      • It runs in 5 milliseconds (rather than 55 seconds).
      • It completely eliminates the need for column slicing (COMBINER_COLUMN_SLICE_WIDTH) and RecursionError workarounds.
      • It natively preserves decimal.Decimal precision for BIGNUMERIC.
      • Most importantly: It completely insulates DVT's combiner from Ibis backend deprecations. When DVT upgrades to Ibis 8, 9, or 10, the combiner will not be affected because it will operate purely on the resulting DataFrames.

third_party/ibis/ibis_oracle/__init__.py
```diff
@@ -318,6 +318,7 @@ def list_databases(self, like: Optional[str] = None) -> list[str]:
             schema
             for schema in all_schemas
             if schema.upper() not in ORACLE_SYSTEM_SCHEMAS
+            and not schema.upper().startswith("C##")
```
nj1973, line 321: I don't understand what this has to do with this PR?

tests/system/data_sources/test_sybase.py
```diff
@@ -61,7 +61,7 @@
     "password": SYBASE_PASSWORD,
     "database": SYBASE_DATABASE,
     "odbc_driver": SYBASE_ODBC_DRIVER,
-    "query": '{"autocommit": "True"}',
+    "query": '{"autocommit": "True", "tds_version":"5.0"}',
```
nj1973, line 64: Is this necessary for our automated tests? They are passing fine without on `develop`?

tests/system/data_sources/test_db2.py
```diff
@@ -544,8 +546,6 @@ def test_row_validation_large_decimals_to_bigquery():
         tables="pso_data_validator.dvt_large_decimals",
         tc="bq-conn",
         hash="id,col_dec_18",
-        use_random_row=True,
```
nj1973, line 547: But this test is passing on `develop` so does that mean the DuckDB upgrade is downgrading our Db2 support?

data_validation/combiner.py
```diff
@@ -40,7 +40,7 @@
 # At around 140 columns we hit RecursionError when unioning Ibis subqueries.
```
nj1973: I ran test_combiner.py unit tests on this branch and they took > 9 minutes:
```
$ pytest tests/unit/test_combiner.py
...
tests/unit/test_combiner.py ........................                                                                                                                                      [100%]

================================================================================ 24 passed in 562.37s (0:09:22)
```

I then ran them on `develop` and they took 1m19s:
```
$ pytest tests/unit/test_combiner.py
...
tests/unit/test_combiner.py ........................                                                                                                                                      [100%]

================================================================================= 24 passed in 79.24s (0:01:19)
```

We are taking a significant performance penalty for these changes.

nj1973 (changes requested):
Based on my brief investigation I don't think we should do this work now.

Instead we should either wait for an Ibis v9 upgrade or look into whether we can refactor the combiner to not use Ibis at all (I accept this might not be feasible - it is only based on Gemini saying it is a cleaner way forwards).

tests/system/data_sources/test_sybase.py
sundar-mudupalli-work: This works on Linux, but is required to get tests running on my Mac.

third_party/ibis/ibis_oracle/__init__.py
sundar-mudupalli-work: For some reason Oracle now returns some [system schemas with C## prefix](https://github.com/GoogleCloudPlatform/professional-services-data-validator/blob/issue-1702-using-duckdb/docs/internal/design_docs/issue-1702-duckdb/walkthrough_first.md#4-oracle-database-listing)

tests/system/data_sources/test_db2.py
sundar-mudupalli-work: This is an [issue](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1822) I reported yesterday

sundar-mudupalli-work:
Neil,

I saw your comment in chat (not your detailed notes above). I thought about the issues with the combiner and realized that a lot of the complexity was coming from SQL results (strongly typed table) to Pandas dataframe (flexible typed dataframe) to DuckDB (strongly typed table). This is unnecessary.

Ibis has a way to return results of SQL Queries as a PyArrow table, which is strongly typed. Further PyArrow natively offers joins and other functions - so we don't need to use Ibis or Pandas. We can just convert the final PyArrow results to Pandas dataframe for downstream applications.  I worked with Generative AI (for some reason the model on VSCode appears to ChatGPT, I need to change that) to refactor combiner to use PyArrow and the native functions rather than Pandas/Ibis. The combiner is about 150 lines shorter. I think that is a good thing. I have checked in the changes into this branch - not all combiner unit tests are passing - I am a bit unsure if all the changes are correct - here is what Generative AI said

-- Begin Generative AI --
Implementation is partially complete, but not ready to call finished.
*Completed*:
* Removed Ibis/DuckDB from the combiner execution path.
* Added Arrow-native table alignment, full joins, filtering-style compute operations, comparison calculations, report construction, metadata, and Arrow-to-Pandas conversion only at the end.
* Preserved binary hex output and Pandas timestamp/random-row dtype compatibility.
* The native Arrow decimal128 and binary report tests pass.

*Remaining report-contract failures*:
* The legacy no-key multi-row test expects 16 rows from the old Ibis pivot cross-product; the Arrow-native implementation produces the direct Cartesian result of 4 rows.
* Two ungrouped cases differ in numeric string formatting such as Arrow’s "1" versus DuckDB’s "1.0".
* One grouped case has row-order differences after Arrow full joins.
* Private _get_summary tests still pass Pandas DataFrames and need conversion to Arrow tables under the new internal contract.

I stopped here rather than changing those behavioral expectations without resolving whether the old 16-row behavior is intentional or an artifact of the removed pivot pipeline.
-- End Generative AI -- 

I don't understand everything that Generative AI is saying - probably worthwhile for you to take a look at it as well. One thing I noticed is the previous `generate_report` code has a for loop which processed things in slices and the current one does not. However there is still a `next(iterator)` in `generate_report` - this will only get executed once, that seems like a bug.

I ran the integration tests on Postgres, Oracle and SQL Server. Each had about 5-7 failures each, completely explainable because we were playing fast and loose since a Pandas dataframe is not strongly typed. We should be more careful and cast the results (in the SQL) into the types we need. One failure was somewhat amusing on Postgres `select avg(t0.col_dec_18) AS avg__col_dec_18 FROM pso_data_validator.dvt_large_decimals t0;` returns decimal `987654321012345678`, while in BigQuery `SELECT avg(t0.``col_dec_18``) AS ``avg__col_dec_18`` FROM ``pso-kokoro-resources``.pso_data_validator.dvt_large_decimals AS t0;` returns float `9.876543210123457E17`. 

In terms of performance, the direct PyArrow code integration tests on Postgres (with 5 failures) took nearly 15:59 on my mac, while the same integration tests on the develop branch took 16:03. So nearly the same performance.

The failures are below:
```
FAILED tests/system/data_sources/test_postgres.py::test_column_validation_pg_types - pyarrow.lib.ArrowInvalid: Could not convert Decimal('1666.5000000000000000') with type ...
FAILED tests/system/data_sources/test_postgres.py::test_column_validation_core_types_to_bigquery - pyarrow.lib.ArrowInvalid: Could not convert Decimal('2.0000000000000000') with type dec...
FAILED tests/system/data_sources/test_postgres.py::test_column_validation_large_decimals_to_bigquery - pyarrow.lib.ArrowInvalid: Integer value 987654321012345678 not in range: -16777216 to 1...
FAILED tests/system/data_sources/test_postgres.py::test_row_validation_comp_fields_pg_types - pyarrow.lib.ArrowInvalid: Casting from timestamp[us] to timestamp[s] would lose data: 1...
FAILED tests/system/data_sources/test_postgres.py::test_column_validation_decimals_no_precision - pyarrow.lib.ArrowInvalid: Could not convert Decimal('2.5000000000000000') with type dec...
FAILED tests/system/data_sources/test_oracle.py::test_column_validation_core_types - pyarrow.lib.ArrowInvalid: Could not convert Decimal('123456.2109375') with type decimal...
FAILED tests/system/data_sources/test_oracle.py::test_column_validation_core_types_to_bigquery - pyarrow.lib.ArrowInvalid: Rescaling Decimal128 value would cause data loss
FAILED tests/system/data_sources/test_oracle.py::test_column_validation_oracle_to_postgres - pyarrow.lib.ArrowInvalid: Rescaling Decimal128 value would cause data loss
FAILED tests/system/data_sources/test_oracle.py::test_column_validation_all_null_oracle_to_postgres - pyarrow.lib.ArrowInvalid: Could not convert Decimal('4.0000000000000000') with type dec...
FAILED tests/system/data_sources/test_oracle.py::test_column_validation_large_decimals_to_bigquery - pyarrow.lib.ArrowInvalid: Integer value 987654321012345678 not in range: -16777216 to 1...
FAILED tests/system/data_sources/test_oracle.py::test_column_validation_tricky_dates_to_bigquery - pyarrow.lib.ArrowTypeError: int or Decimal object expected, got float
FAILED tests/system/data_sources/test_oracle.py::test_column_validation_decimals_no_precision - pyarrow.lib.ArrowInvalid: Rescaling Decimal128 value would cause data loss
FAILED tests/system/data_sources/test_oracle.py::test_row_validation_comp_fields_oracle_to_postgres - pyarrow.lib.ArrowTypeError: int or Decimal object expected, got float
FAILED tests/system/data_sources/test_oracle.py::test_row_validation_datetime_pk_to_bigquery - AssertionError: We expect 5 rows with status success from this validation
FAILED tests/system/data_sources/test_oracle.py::test_row_validation_comp_fields_bool_to_postgres - pyarrow.lib.ArrowInvalid: Could not convert 1 with type int: tried to convert to boolean
FAILED tests/system/data_sources/test_sql_server.py::test_column_validation_core_types - pyarrow.lib.ArrowTypeError: int or Decimal object expected, got float
FAILED tests/system/data_sources/test_sql_server.py::test_column_validation_core_types_to_bigquery - pyarrow.lib.ArrowTypeError: int or Decimal object expected, got float
FAILED tests/system/data_sources/test_sql_server.py::test_column_validation_ss_types_to_bigquery - pyarrow.lib.ArrowInvalid: Integer value 123456789012345678 not in range: -9...
FAILED tests/system/data_sources/test_sql_server.py::test_column_validation_large_decimals_to_bigquery - pyarrow.lib.ArrowInvalid: Integer value 987654321012345678 not in range: -1...
FAILED tests/system/data_sources/test_sql_server.py::test_row_validation_comp_fields_ss_types_to_bigquery - AssertionError
```

Further, Ibis can also return PyArrow tables in batches - so we can guarantee that DVT never OOM, but that is a feature for the future. 

Thank you.

Sundar Mudupalli

third_party/ibis/ibis_oracle/__init__.py
nj1973: I suspect that is because your test Oracle database is using the CDB instead of the PDB. We could add additional entries to ORACLE_SYSTEM_SCHEMAS. But it's not impossible for users to create their own schemas prefixed with "C##" so we should not use `startswith`.

Probably an issue in its own right rather than hid inside the changes on this branch.

tests/system/data_sources/test_sybase.py
nj1973: Probably better to do something similar to what we do for SQL Server then:

```
SQL_SERVER_CONFIG_JSON = os.getenv("SQL_SERVER_CONFIG_JSON")
```

Enabling the tester to set it using something like:
```
export SQL_SERVER_CONFIG_JSON='{"driver": "ODBC Driver 18 for SQL Server", "TrustServerCertificate": "yes"}'
```

I think that would be better than hardcoding your Mac requirements into the tests.

data_validation/combiner.py
nj1973: This branch has switched to implementing pyarrow therefore closing this thread.

data_validation/combiner.py
```diff
@@ -185,424 +122,341 @@ def _generate_report_slice(
...
+            pc.cast(source_value, pyarrow.timestamp("s")), pyarrow.int64()
+        )
+        target_value = pc.cast(
+            pc.cast(target_value, pyarrow.timestamp("s")), pyarrow.int64()
+        )
+    elif pyarrow.types.is_boolean(source_type) or pyarrow.types.is_boolean(target_type):
+        source_value = pc.cast(source_value, pyarrow.bool_())
+        target_value = pc.cast(target_value, pyarrow.bool_())
+    elif pyarrow.types.is_decimal(source_type) or pyarrow.types.is_floating(source_type):
+        # Cast to float32 and round to 4 decimal places to prevent false-positive failures
```
nj1973: We need to rethink this, we can't cast all decimals to float32. That will corrupt all decimals with a precision > 9. float64 is no good either because that will still corrupt many decimal values.

If we did this because of the imprecision of floating point columns then we could remove `is_decimal` from the `if` condition. But if we did this because we sometimes get decimal on one side and float on the other then we need a different approach.

nj1973:
A comment related to the `docs/internal/design_docs/combiner_review.md` document you added to the branch:

https://github.com/GoogleCloudPlatform/professional-services-data-validator/pull/1823/changes/d5a815b05d3261b457a6811186168e18831ae1af

If we go back to the base reason for this PR, I believe it was Ibis deprecating Pandas support in a future Ibis release and we wanted to be proactive and deal with it now. So we:

1. explored switching the combiner from using Pandas to DuckDB but that was not a good fit.
2. explored switching the combiner from using Pandas to PyArrow the changes on this branch.
3. as part of item 2 we realised we didn't actually need Ibis in the combiner and removed that too.

Does the realisation in item 3 above give us an alternative end state? If we start over and just refactor the combiner to not use Ibis (pure Pandas processing) then the deprecation becomes a moot point. The combiner continues to use Pandas which derisks moving from flexible to strict typing. Plus the combiner becomes simpler to understand by removing Ibis.

And maybe in the future we still want to review other processing engines but that's a nice-to-have task unrelated to the deprecation issue.

PR #1828, opened 2026-09-17
feat: Support dynamic chunking in configs run -kc

### LINKED ISSUE #1827: Support dynamic round-robin chunking in configs run --kube-completions
### 1. Summary & Context
This PR adds support for **dynamic round-robin chunking** to the `configs run --kube-completions` (`-kc`) command. 

Currently, the `-kc` flag is designed for a strict **1-to-1 sequential numeric mapping**:
* It expects files in the `--config-dir` to be named sequentially (e.g., `0000.yaml`, `0001.yaml`, etc.).
* It maps the `CLOUD_RUN_TASK_INDEX` or `JOB_COMPLETION_INDEX` directly to a single file: `f"{job_index:04d}.yaml"`.

While this works perfectly for validations generated via `generate-table-partitions` (which splits a single massive table into numbered chunks), it **does not scale** for database migrations involving hundreds or thousands of different tables where configuration files are named after their schemas/tables (e.g., `hr.employees.yaml`, `sales.orders.yaml`) rather than sequential numbers.

### 2. Proposed enhancement: Dynamic Round-Robin Chunking
Propose extending `-kc` to automatically detect the **total task count** via standard environment variables (`CLOUD_RUN_TASK_COUNT` or `JOB_COMPLETION_COUNT`).

* **If a total task count is detected**:
  1. DVT lists all validation files in the `--config-dir` (supporting GCS and local paths) and sorts them alphabetically.
  2. It distributes the files round-robin to the current task using: `my_files = all_files[task_index::task_count]`.
  3. It executes the assigned files sequentially within the warm container.
* **If no task count is detected (Legacy Fallback)**:
  * DVT falls back to the legacy 1-to-1 sequential numeric naming (e.g., `0000.yaml`). This ensures equivalent behaviour when task count equals file count.

nj1973:
## Description of changes
This PR enhances our `configs run -kc` code. Prior to this change only files formatted as `0000n.yaml` were processed because the code was targeted only at the output of `generate-table-partitions`, although it was possible to manually subvert this to work for regular table validations by naming yaml files to match the expected format.

After these changes the `-kc` code shares yaml files out amongst Kubernetes/Cloud Run workers using similar `mod` slicing but it is not file name based, so it will work for a set of files with any name. The advantage of this is that a user could store YAML files for 100s of tables in a GCS folder and process them in parallel without having to rename them in the `0000n.yaml` format.

In addition to the enhancement above the following changes are also included:

1. Safeguards for empty or misspelt `--config-dir`
2. Task index & task count bounds validation
3. Fixed pollution of os.environ by unit tests using monkeypatch functionality

## Issues to be closed

Closes #1827

## Checklist


data_validation/__main__.py
```diff
@@ -379,65 +379,175 @@ def _build_configs():
...
+                )
+                if not my_files:
+                    logging.warning(
+                        "Task %d has no config files to run, consider reducing the number of tasks.",
+                        job_index,
+                    )
+
+                _run_config_files(args, my_files)
+            else:
+                # --- Legacy 1-to-1 Fallback ---
```
sundar-mudupalli-work: Replace this with 
```
# --- Fallback if using running in Kubernetes and the user did not set JOB_COMPLETION_COUNT in the Job manifest
```

data_validation/__main__.py
```diff
@@ -379,65 +379,175 @@ def _build_configs():
...
+                config_file_path = (
+                    f"{args.config_dir}{job_index:04d}.yaml"
+                    if args.config_dir.endswith("/")
+                    else f"{args.config_dir}/{job_index:04d}.yaml"
+                )
+                setattr(args, "config_dir", None)
```
sundar-mudupalli-work:
```
# -- Detect if the user inadvertently forgot to set JOB_COMPLETION_COUNT
all_files = _list_validation_files(args.config_dir)
if config_file_path not in all_files:
    raise ValueError(
            f"Validation YAML files exist: {config_dir}. Please set JOB_COMPLETION_COUNT in the Job manifest to run validations in parallel"
        )
```

sundar-mudupalli-work (changes requested):
Great feat.

With this change, the config runner code is over 100 lines long. `__main.py__` is already 700+ lines long, only `config_manager.py` and `cli_tools.py` have more lines of code. I suggest moving the config runner code out of `__main__.py` into its own module along with its unit tests. It is fairly separate module.

Sundar Mudupalli

data_validation/__main__.py
nj1973: Changed

data_validation/__main__.py
nj1973: Changed.

nj1973:
> With this change, the config runner code is over 100 lines long. `__main.py__` is already 700+ lines long, only `config_manager.py` and `cli_tools.py` have more lines of code. I suggest moving the config runner code out of `__main__.py` into its own module along with its unit tests. It is fairly separate module.

This is a great suggestion. I have refactored to config and validation runner code into their own modules. I think it is much better. Thanks

sundar-mudupalli-work:
Neil,

Thank you for the work. After some thought, I no longer support chunking and [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1827#issuecomment-5769434922) is why. Sorry about it - I think it is better to not chunk files.

I very much appreciate the rest of the changes.

Cheers.

Sundar Mudupalli

nj1973:
> Thank you for the work. After some thought, I no longer support chunking and [here](https://github.com/GoogleCloudPlatform/professional-services-data-validator/issues/1827#issuecomment-5769434922) is why. Sorry about it - I think it is better to not chunk files.

I responded to the individual points from the comment on the issue.

In summary, we already "chunk files" for table partition validations using `--parts-per-file` so I don't understand that element of the critique. But also, we need to not lose sight that this PR is about multi table validations - not partition-wise validations, I don't see a better alternative in our current prod code and disagree with "better to not chunk files".

